Issue #2, opened 2016-12-02
Fix erroneous logic in OLS regression standard error


abshinn:
From @dirkjordan:
> 1. I think statsmodels may have these coefficients available, but if you want to calculate them I would multiply Rd by 100 for the percent. Not critical but reads easier.
> 2. The RMSE is divided by N-2, not N.
> 3. SE_b and SE_m were switched. I think I made the same mistake in my 2010 PVMRW presentation and I corrected it there too.
> 4. The new SE_b is divided by N-1, not just N
> 5. For the uncertainty of the regression after you square the SE, you need to take the square root and multiply by 100 (only if you do it on line 1 as well).

PR #3, opened 2016-12-02
Fix incorrect OLS standard error logic.

abshinn:
Fixed standard error logic. See #2.

Also, I added comments and rename some variables to improve code readability.

### DIFF
--- a/rdtools/degradation.py
+++ b/rdtools/degradation.py
@@ -34,35 +34,49 @@ def degradation_with_ols(normalized_energy):
     # remove NaN values
     y = y.dropna()
 
-    # integer-months
+    # number of examples
+    N = len(y)
+
+    # integer-months, the exogeneous variable
     months = np.arange(0, len(y))
+
+    # add intercept-constant to the exogeneous variable
     X = sm.add_constant(months)
     columns = ['constant', 'months']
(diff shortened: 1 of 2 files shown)


rdtools/degradation.py
```diff
@@ -34,35 +34,49 @@ def degradation_with_ols(normalized_energy):
...
-    SE_b = rmse * np.sqrt(1 / np.power(months - months.mean(), 2).sum())
-    SE_Rd = np.power(SE_m * 12/b, 2) + np.power((-12*m / b**2) * SE_b, 2)
+    # root mean square error
+    rmse = np.sqrt(np.power(y - results.predict(exog=exog), 2).sum() / (N - 2))
+
+    # total sum of squares of the time variable
+    tss_months = np.power(months - months.mean(), 2).sum()
+
+    # standard error of the slope and intercept
+    stderr_b = rmse * np.sqrt((1/N) + np.power(months.mean(), 2) / tss_months)
```
dirkjordan: stderr_b = rmse * np.sqrt((1/(N-1)) + np.power(months.mean(), 2) / tss_months)

The SE for the intercept should be divided by N-1. Often in statistical notation it is indicated by n, but n=N-1. Whoever came up with this confusing notation  ought to be shot, but hey that's the convention.

PR #7, opened 2017-01-06
Included new functions: degradation_year_on_year and degradation_clas…

cdeline:
…sical_decomposition.

I've tested the two methods on local data, but could use some help with testing on your end.  Also, any suggestions on format for input / output variables, structure of subroutines, etc. would be helpful.  thanks!

### DIFF
--- a/rdtools/degradation.py
+++ b/rdtools/degradation.py
@@ -136,6 +136,7 @@ def degradation_classical_decomposition(normalized_energy, interpolate_flag = Fa
     # OLS regression with constant
     x2 = sm.add_constant(x2)
     model2 = sm.OLS(y2,x2).fit()
+    b_cd, m_cd = model2.params
     Rd_cd, SE_Rd_cd = ols_rd_uncertainty(model2)    
 
     print '\nDegradation and Standard Error of Classical decomposition:'
--- a/rdtools/degradation.py
+++ b/rdtools/degradation.py
@@ -13,6 +13,8 @@ import statsmodels.api as sm
 def degradation_with_ols(normalized_energy):
     '''
     Description
+    -----------
+    OLS routine
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


abshinn:
@cdeline, this looks great! First I will go through and make some style suggestions and ask questions in-line. When I get a moment this week I will check out the branch and test it out.

rdtools/degradation.py
```diff
@@ -80,3 +80,293 @@ def degradation_with_ols(normalized_energy):
...
+        x2=sm.add_constant(x2)
+        model2 = sm.OLS(y2,x2).fit()
+    
+        b_cd = model2.params['const']
+        m_cd = model2.params['Month']
+        SE_b_cd = model2.bse['const']
+        SE_m_cd = model2.bse['Month']
+    
+        Rd_cd = (m_cd * 12) / b_cd * 100
+        SE_Rd_cd = np.sqrt(np.power(SE_m_cd * 12/b_cd, 2) + np.power((-12*m_cd/b_cd**2) * SE_b_cd, 2))*100
```
abshinn: This OLS uncertainty calculation is also in `degradation_with_ols` function. In order to not duplicate logic, this can be factored-out into another function, perhaps `ols_rd_uncertainty(ols_model)` -- which takes the ols model class as input.

rdtools/degradation.py
```diff
@@ -80,3 +80,293 @@ def degradation_with_ols(normalized_energy):
...
+    # (YOY_median,YOY_unc,YoY_filtered1) = degradation_year_on_year(normalized_energy)
+    #
+    # year-on-year  decomponsition method 
+    # from Dirk Jordan (NREL) 12/10/16
+    # input parameters:  
+    #       series['normalized_energy'] : corrected performance ratio series
+    #       index:  timeseries index in monthly format
+    #       
+    #output parameters:
+    #   (YOY_median,YOY_unc) : tuple of median degradation rate and standard error
```
abshinn: Comment block should be docstring.

rdtools/degradation.py
```diff
@@ -80,3 +80,293 @@ def degradation_with_ols(normalized_energy):
...
+    # TODO:  check timebase for other input time options (day, week)    
+
+    ################################################################################################################################
+    # YOY approach
+    
+    # check the index frequency
+    freq = pd.infer_freq(normalized_energy.index)
+    if freq == 'MS': #monthly
```
abshinn: Monthly can also be 'M', month-end.

rdtools/degradation.py
```diff
@@ -80,3 +80,293 @@ def degradation_with_ols(normalized_energy):
...
+        YearSampleSize = 12
+    elif freq == 'W':
+        YearSampleSize = 52
+    elif freq == 'D':
+        YearSampleSize = 365
+    else:  #sample to daily by default
```
abshinn: For debugging purposes, it might be best to explicitly name frequencies under a day and throw an exception if the frequency is not supported, something like:

```
elif freq in ['H', '30T', '15T', 'T']:
    # resample to a daily
else:
    raise Exception('Frequency {} not supported'.format(freq))
```

rdtools/degradation.py
abshinn: [Pandas offset aliases.](http://pandas.pydata.org/pandas-docs/stable/timeseries.html#offset-aliases)

rdtools/degradation.py
```diff
@@ -80,3 +80,293 @@ def degradation_with_ols(normalized_energy):
...
+        normalized_energy = normalized_energy.resample('D').mean()
+        YearSampleSize = 365
+    #  year-on-year approach.
+    YoYresult = []
+    for index in range(normalized_energy.size-YearSampleSize):
+        YoYresult.append( (normalized_energy[index+YearSampleSize]-normalized_energy[index])/+normalized_energy[index]*100 )
```
abshinn: I think the [Pandas diff method](http://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.diff.html) would be useful here. It would look something like:

```
YoYresult = normalized_energy.diff(YearSampleSize) / normalized_energy
```

rdtools/degradation.py
```diff
@@ -80,3 +80,293 @@ def degradation_with_ols(normalized_energy):
...
+        'Rd_pct': Rd_cd,
+        'slope': m_cd,
+        'intercept': b_cd,
+        #'rmse': rmse,
+        #'slope_stderr': stderr_m,
+        #'intercept_stderr': stderr_b,
+        'Rd_stderr_pct': SE_Rd_cd,
+        'Dataframe':dataframe
+        }
+    return(degradation_values)
```
abshinn: Return is a statement. This should be:

```
return degradation_values
```

rdtools/degradation.py
```diff
@@ -80,3 +80,293 @@ def degradation_with_ols(normalized_energy):
...
+        'Rd_pct': Rd_decomp,
+        'test_trend': test_trend,
+        'p': p,
+        #'rmse': rmse,
+        #'slope_stderr': stderr_m,
+        #'intercept_stderr': stderr_b,
+        'Rd_stderr_pct': SE_Rd_decomp,
+        'Dataframe':df4,
+        }
+    return(degradation_values)
```
abshinn: `degradation_values` is not defined in the `if len(df4) <=2` if-block, therefore this return statement will throw an exception in that case.

rdtools/degradation.py
```diff
@@ -80,3 +80,293 @@ def degradation_with_ols(normalized_energy):
...
+    x3=df4['Month']           
+    
+    
+    
+    if len(df4) <=2:
+        Rd_decomp = 11111
+        SE_Rd_decomp = 11111
+        test_trend = 11111
+        p = 11111
+        print '\nNot enough data for ARIMA decomposition:'
```
abshinn: If there is not enough data, perhaps return with an empty results dictionary:

```
if len(df4) <= 2:
    print '\nNot enough data for ARIMA decomposition:'
    return {}
```

rdtools/degradation.py
```diff
@@ -80,3 +80,293 @@ def degradation_with_ols(normalized_energy):
...
+
+#    Input:
+#        x:   a vector of data
+#        alpha: significance level (0.05 default)
+#    Output:
+#        trend: tells the trend (increasing, decreasing or no trend)
+#        h: True (if trend is present) or False (if trend is absence)
+#        p: p value of the significance test
+#        z: normalized test statistics 
+
```
abshinn: Perhaps copy the [original file](https://github.com/mps9506/Mann-Kendall-Trend/blob/master/mk_test.py) and put it in a `rdtools/utils/` directory?

PR #8, opened 2017-02-17
Issue 4 generalize freq

mdeceglie:
Updated degradation functions to accept various frequencies. Removed ARIMA function (it was the same as classical decomposition).

### DIFF
--- a/rdtools/degradation.py
+++ b/rdtools/degradation.py
@@ -157,7 +157,7 @@ def degradation_classical_decomposition(normalized_energy):
     test_trend, h, p, z = _mk_test(df.energy_ma.dropna(), alpha=0.05)
 
     #Monte Carlo for error in degradation rate
-    dist = [z[1]/z[0] for z in np.random.multivariate_normal(results.params, results.cov_params(), 10000)]
+    dist = [x_[1]/x_[0] for x_ in np.random.multivariate_normal(results.params, results.cov_params(), 10000)]
     Rd_CI = np.percentile(dist, [50-34.1, 50+34.1])*100.0
 
     degradation = {
--- a/rdtools/degradation.py
+++ b/rdtools/degradation.py
@@ -10,7 +10,7 @@ import numpy as np
 import statsmodels.api as sm
 
 
-def degradation_with_ols(normalized_energy):
+def degradation_ols(normalized_energy):
     '''
     Description
     -----------
--- a/rdtools/degradation.py
+++ b/rdtools/degradation.py
@@ -18,54 +18,44 @@ def degradation_with_ols(normalized_energy):
 
     Parameters
     ----------
-    normalized_energy: Pandas Series (numeric)
-        Monthly time series of normalized system ouput.
+    normalized_energy: Pandas Time Series (numeric)
+        Daily or lower frequency time series of normalized system ouput.
 
     Returns
     -------
     degradation: dictionary
-        Contains degradation rate and standard errors of regression
+        Contains degradation rate (%/year), standard errors of regression, and the 
+        ols RegressionResults object
     '''
 
-    y = normalized_energy
-
-    # remove NaN values
-    y = y.dropna()
-
-    # number of examples
-    N = len(y)
-
-    # integer-months, the exogeneous variable
-    months = np.arange(0, len(y))
-
-    # add intercept-constant to the exogeneous variable
-    X = sm.add_constant(months)
-    columns = ['constant', 'months']
-    exog = pd.DataFrame(X, index = y.index, columns = columns)
-
-    # fit linear model
-    ols_model = sm.OLS(endog = y, exog = exog, hasconst = True)
+    normalized_energy.name = 'normalized_energy'
+    df = normalized_energy.to_frame()
+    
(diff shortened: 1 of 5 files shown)


rdtools/degradation.py
```diff
@@ -74,82 +71,107 @@ def degradation_with_ols(normalized_energy):
...
-    # OLS regression with constant
-    x2 = sm.add_constant(x2)
-    model2 = sm.OLS(y2,x2).fit()
-    b_cd, m_cd = model2.params
-    Rd_cd, SE_Rd_cd = ols_rd_uncertainty(model2)    
-
-    print '\nDegradation and Standard Error of Classical decomposition:'
-    print 'Rd = {:.2f} +/- {:.2f}'.format(Rd_cd, SE_Rd_cd)  
+    #Monte Carlo for error in degradation rate
+    dist = [z[1]/z[0] for z in np.random.multivariate_normal(results.params, results.cov_params(), 10000)]
```
abshinn: Just to point out an alternate approach here, if you want to use numpy array arithmetic, you can use slice notation on the ndarray returned from `multivariate_normal()`:

```
sampled_normal = np.random.multivariate_normal(results.params, results.cov_params(), 10000)
dist = sampled_normal[:,1] / sampled_normal[:,0]
```

rdtools/degradation.py
```diff
@@ -222,7 +244,7 @@ def remove_outliers(x):
   
     med1 = np.median(YoY_filtered1)                       
     
-    # bootstrap to determine 95% CI for the 2 different outlier removal methods
+    # bootstrap to determine 68% CI for the 2 different outlier removal methods
```
abshinn: Why the change to 68% confidence interval?

rdtools/degradation.py
```diff
@@ -74,82 +71,107 @@ def degradation_with_ols(normalized_energy):
...
-    x2 = sm.add_constant(x2)
-    model2 = sm.OLS(y2,x2).fit()
-    b_cd, m_cd = model2.params
-    Rd_cd, SE_Rd_cd = ols_rd_uncertainty(model2)    
-
-    print '\nDegradation and Standard Error of Classical decomposition:'
-    print 'Rd = {:.2f} +/- {:.2f}'.format(Rd_cd, SE_Rd_cd)  
+    #Monte Carlo for error in degradation rate
+    dist = [z[1]/z[0] for z in np.random.multivariate_normal(results.params, results.cov_params(), 10000)]
+    Rd_CI = np.percentile(dist, [50-34.1, 50+34.1])*100.0
```
abshinn: Since the confidence interval logic is used in the `degradation_ols` function as well, perhaps it should be factored-out into a separate function. That way, if there are changes to this logic, it only needs to be changed in one place.

requirements.txt
```diff
@@ -7,6 +7,6 @@ python-dateutil==2.6.0
 pytz==2016.7
 scipy==0.18.1
 six==1.10.0
-statsmodels==0.6.1
+statsmodels==0.8.0
```
abshinn: Can you make this change to `setup.py` as well? When rdtools is installed via pip, it uses `setup.py` to download all of the corresponding dependencies.

rdtools/degradation.py
```diff
@@ -10,62 +10,59 @@
 import statsmodels.api as sm
 
 
-def degradation_with_ols(normalized_energy):
+def degradation_ols(normalized_energy):
```
abshinn: Oh! One more comment -- I am ok with the function name change, however `degradation_with_ols` is still referenced within `rdtools/__init__.py`. This is so that the function is available in rdtools' namespace. The function reference should be updated to reflect `degradation_ols`.

rdtools/degradation.py
```diff
@@ -74,82 +71,107 @@ def degradation_with_ols(normalized_energy):
...
-    x2 = dataframe['Month']
+    #calculate a years column as x value for regression, ignoreing leap years
+    day_diffs = (df.index - df.index[0])
+    df['days'] = day_diffs.astype('timedelta64[s]')/(60*60*24)
+    df['years'] = df.days/365.0
     
+    #Compute yearly rolling mean to isolate trend component using moving average
+    it = df.iterrows()
+    energy_ma = []
+    for i, row in it:
```
mdeceglie: Good question. We need a centered rolling mean for seasonal decomposition. However `center = True` is not implemented for  offset-based windows e.g. `.rolling('365D', center = True)` raises a `NotImplementedError`.

rdtools/degradation.py
mdeceglie: This is a questions for @dirkjordan and @cdeline , but my guess is that this aligns with the standard error one report from a regression. (I just updated the comment to reflect the code.)

rdtools/degradation.py
mdeceglie: Ah, should we then also add `degradation_year_on_year` and `degradation_classical_decomposition` as well?

mdeceglie:
Thanks for the review @abshinn, I have incorporated your suggestions.

rdtools/degradation.py
abshinn: Correct.

PR #14, opened 2017-04-04
update degradation_year_on_year

BMcDonie:
Simplify removal of outliers.
Fixed what I assume is an issue in the calculation where YoY change was = (T1 - T0) / T1 and is now (T1 - T0) / T0. 
Change format of returned data (YoY_filtered) from array to pandas series. Allowing easier analysis of degradation results (e.g., evaluating results by hour, month, year, etc...).

### DIFF
--- a/rdtools/degradation.py
+++ b/rdtools/degradation.py
@@ -192,7 +192,7 @@ def degradation_year_on_year(normalized_energy, freq = 'D'):
     degradation_values: dictionary
         Contains values for median degradation rate and standard error
         'Rd_med', 'Rd_stderr_pct', 'YoY_filtered'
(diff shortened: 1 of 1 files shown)


GregoryKimball:
If the metric is normalized, why are you dividing by T0?

mdeceglie:
@BMcDonie, Thank you for the input. You are right, there is a mistake in the current YoY implementation. However, it raises a bigger question as @GregoryKimball points out. 

Regarding the returned data, issue #16  (open) calls for the outputs to be standardized, it will be best to update all outputs to a consistent format at once.

We will work toward consensus on these two questions in the coming days. For now I will close this pull request.

Issue #10, opened 2017-03-02
Provide usage examples


abshinn:
- Standardize degradation function outputs and printed output.
- Commit notebooks with usage examples of the code.
- Supply links to notebooks ([notebooks would be rendered so that users do not need to launch Jupyter Notebook to view the code and plots](http://blog.jupyter.org/2015/05/07/rendering-notebooks-on-github/)).

mdeceglie:
This appears to be a duplicate of #13. I will close this and open a new issue on standardizing degradation function outputs.

PR #17, opened 2017-04-19
adding example jupyter notebook

GregoryKimball:
This request adds a jupyter notebook and a "docs" folder to the repository.  PVLIB uses a similar structure to provide tutorials to the functionality.


GregoryKimball:
This pull request may address issue #10
[Provide usage examples #10 ]

mdeceglie:
@GregoryKimball I made a few minor changes, please take a look and make sure they look good to you.

mdeceglie:
This is a great start to illustrate the workflow. Let's make sure to keep it updated as we build out the tools.

GregoryKimball:
Thank you for improving the notebook.

I added a centering step so that the data is initially centered around 1 before the analysis begins.

mdeceglie:
Thanks for illustrating the centering step. I suggest it might be more general and readable to use `DataFrame.rolling().median()`

GregoryKimball:
I did some experimenting with .rolling() functions, but for sites with seasonality in the performance metric (i.e. a lot of them), I think the 1-year resample max may be more appropriate.

cdeline:
I reviewed this branch, and really like it.  I think it's important to get some usage examples up quickly, and this is a great start.  The fact that the Kyocera system isn't a clean dataset may lead to issues - e.g. -8% per year degradation rates, but hopefully this will be resolved as we improve the filtering.

I did include a few additional plots to visualize what the output of the YoY and OLS regression looks like.
image

image

mdeceglie:
The data from the BP system (12) system may make for a cleaner example: http://dkasolarcentre.com.au/source/alice-springs/dka-m5-b-phase

cdeline:
Good suggestion Mike - I like system 12 very much!

dirkjordan:
We could also consider one of the NREL systems that nobody cares about anymore like the ASE.

GregoryKimball:
Thank you Chris and Mike for adding the new BP system.  I made some updates to the metadata and filter to make sure this site runs OK.

Ready to merge?

cdeline:
I vote for merge.

Chris Deline
National Renewable Energy Laboratory
303-384-6359

mdeceglie:
I will update the example to work with the newly merged versions of the degradation functions.

mdeceglie:
The previous version was using ambient as cell temperature leading to significant seasonality. I updated the normalization step to model cell temperature based on POA irradiance. But there are still remaining problems:

- OLS is not a good idea with 15 minute data, there are a lot of low data points that cause a highly suspect fit
-The new version of YOY won't accept anything more frequent than daily.

The solution to both these problems is to aggregate with an irradiance weighted sum daily. I'll keep working on this in the example.

PR #22, opened 2017-05-12
pull request for Yoy upgrades branch

amnag:


tests/degradation_test_v1.py
```diff
@@ -0,0 +1,182 @@
...
+        # - support for different time series frequencies
+        # - inputs
+
+
+if __name__ == '__main__':
+
+    # Allowed frequencies for degradation_ols
+    list_ols_input_freq = ['MS','M','W','D','H','T','S']
+    # Allowed frequencies for degradation_classical_decomposition and degradation_year_on_year
+    list_CD_YOY_input_freq = ['MS','M','W','D']
```
abshinn: I prefer v2 to v1. Either way gets the job done, but it is better practice to define variables for the tests inside the test classes themselves.

tests/degradation_test_v2.py
```diff
@@ -0,0 +1,175 @@
...
+
+        self.test_ols_corr_energy = test_ols_corr_energy
+        
+        for input_freq in self.list_CD_YOY_input_freq:
+            corr_energy = self.get_CD_YOY_corr_energy(self.rd, input_freq)
+            test_CD_YOY_corr_energy[input_freq] = corr_energy
+         
+        self.test_CD_YOY_corr_energy = test_CD_YOY_corr_energy
+
+    def tearDown(self):
```
abshinn: We can remove the tearDown since we are not using it. This function is useful for cleaning up variables after each test is run.

tests/degradation_test_v2.py
```diff
@@ -0,0 +1,175 @@
...
+            days = np.arange(N)
+            y = np.ones(N) * np.power(1 + rd/365, days)
+        else :
+            sys.exit("Unknown frequency type")
+        
+        corr_energy = pd.Series(data=y, index=x)
+        return corr_energy
+
+
+    def setUp(self):
```
abshinn: The setUp method runs before every test function (functions named test_). So this method ends up running three times. We only need to define the test data once, so what we need is [setUpClass](https://docs.python.org/2/library/unittest.html#unittest.TestCase.setUpClass).

All you need to do is rename `setUp(self)` to `setUpClass(cls)`, add the `@classmethod` decorator, and replace `self` with `cls`.

tests/degradation_test_v2.py
```diff
@@ -0,0 +1,175 @@
...
+        for input_freq in self.list_ols_input_freq:
+            print 'Frequency: ', input_freq
+            rd_result = degradation_ols(self.test_ols_corr_energy[input_freq])
+            self.assertAlmostEqual(rd_result[0], 100*self.rd, places=1)
+            print 'Actual: ', 100*self.rd
+            print 'Estimated: ', rd_result[0]
+
+        # TODO
+        # - support for different time series frequencies
+        # - inputs
```
abshinn: Can be removed now.

tests/degradation_test_v2.py
```diff
@@ -0,0 +1,175 @@
...
+        # test classical decomposition degradation calc
+        for input_freq in self.list_CD_YOY_input_freq:
+            print 'Frequency: ', input_freq
+            rd_result = degradation_classical_decomposition(self.test_CD_YOY_corr_energy[input_freq])
+            self.assertAlmostEqual(rd_result[0], 100*self.rd, places=1)
+            print 'Actual: ', 100*self.rd
+            print 'Estimated: ', rd_result[0]
+        # TODO
+        # - support for different time series frequencies
+        # - inputs
```
abshinn: Can be removed now.

tests/degradation_test_v2.py
```diff
@@ -0,0 +1,175 @@
...
+        for input_freq in self.list_CD_YOY_input_freq:
+            print 'Frequency: ', input_freq
+            rd_result = degradation_year_on_year(self.test_CD_YOY_corr_energy[input_freq])
+            self.assertAlmostEqual(rd_result[0], 100*self.rd, places=1)
+            print 'Actual: ', 100*self.rd
+            print 'Estimated: ', rd_result[0]
+
+        # TODO
+        # - support for different time series frequencies
+        # - inputs
```
abshinn: Can be removed now.

tests/run_tests
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
```
abshinn: To make this file executable, run `chmod +x tests/run_tests`. Then, all you need run the test in the command line is `tests/run_tests`.

tests/degradation_test_v2.py
```diff
@@ -0,0 +1,175 @@
...
+import unittest, sys
+
+import pandas as pd
+import numpy as np
+
+from rdtools import degradation_ols, degradation_classical_decomposition, degradation_year_on_year
+
+class DegradationTestCase(unittest.TestCase):
+    ''' Unit tests for degradation module. 
+        Works with both python and nosetests
```
abshinn: Can remove this comment if we are moving away from nosetests.

tests/degradation_test_v2.py
```diff
@@ -0,0 +1,175 @@
+""" Degradation Module Tests. """
```
abshinn: Since this is the file we are going with, you can remove the other two files and rename this to tests/degradation_test.py.

mdeceglie:
Thanks @amnag , Version 2 looks great. I'm modifying the tests slightly to capture some of my intent with the updated degradation functions. Stay tuned for an update.

amnag:
@mdeceglie, you are welcome.

tests/run_tests
amnag: Hi @abshinn, I will address your comments.
Thanks

mdeceglie:
@amnag, I made some updates in version 2, starting from there I suggest you go ahead and work through @abshinn's comments. Thanks!

amnag:
Hi @abshinn and @mdeceglie, I have pushed the changes suggested by @abshinn to the YOY_upgrades branch of the repository.
Thanks,
Ambarish.

mdeceglie:
Now that the tests are in good shape (thanks @ambarishnag). I am requesting @GregoryKimball  and @cdeline as reviewers to take a look at the updates to the degradation rate calculations.

cdeline:
I checked things out and they look good.  I did get an error on degradation_classical_decomposition because of NaN's - maybe that's a separate issue we should look at.

mdeceglie:
Thanks for the review @cdeline, When you tried classical decomposition with NaNs hopefully you got the value error ''Classical decomposition requires a regular time series with defined frequency and no missing data." That is the expected behavior, I believe those are mathematical requirements of the technique.

If you got some other error message, let me know, that would be an issue we should address.

GregoryKimball:
Thank you for the great work on this code.  I made some changes to the renormalizing part of the function:

- changed renormalizing approach to take biggest year median, rather than first year median.  this is because sites sometimes have poor year1 performance.
- separated renormalizing step from time delta computation.
- added renormalizing factor to calc_dict.
- btw, the "bootstrap" code yields almost the same number as `series.std()`.  Perhaps it would be better to just use `.std()`?

mdeceglie:
Thank you for the review and contributions, @GregoryKimball.

@cdeline and @dirkjordan, what do you guys think of Greg's proposal to use the biggest year median and the normalization factor, rather than the first year? It's a relatively minor change but different than we have discussed. Also, it brings the YOY approach a bit more out of alignment from the OLS and seasonal decomposition approaches. (But they could be updated as well...)

GregoryKimball:
@mdeceglie Thank you for your comment.  We might consider separating the renormalization function from the YOY, OLS and DECOMP approaches, rather than exposing a boolean toggle.

cdeline:
Thanks Mike,

I think we discussed this earlier, but I still do believe that if you’re thinking of setting a baseline for your annual degradation, it really should be the first year, considered the time zero point.  If you do have initial data quality issues as Greg has suggested, the erroneous data should just be removed from the regression, rather than re-basing the assessment on another following year.  To be consistent with the OLS regresssion approach which considers the y-axis intercept as the ‘initial’ t0 value, I feel that if we’re including a default behavior for YoY recentering, it makes the most sense to do it based on year 1 performance.  If a more advanced workflow wants to consider a different re-centering behavior, they could do something other than this default behavior.

Thanks!

Chris Deline
National Renewable Energy Laboratory
303-384-6359

mdeceglie:
Thanks to everyone for their work and input on this branch. I have changed the default behavior back to recentering on first year by default (sophisticated users can turn off this default behavior), and will go ahead and merge.

Issue #21, opened 2017-04-26
Mixing of power and energy in normalization.py


mdeceglie:
If irradiance and PV energy data are provided to the normalization functions with the same frequency, or with PV energy data being higher frequency than the irradiance data, the results returned by `normalize_with_pvwatts` and `normalize_with_sapm` will be in units of power/energy.

mdeceglie:
Closed with merge of pull request #28

Issue #19, opened 2017-04-20
Include aggregation function for irradiance-weighted averages


mdeceglie:

mdeceglie:
Closed with merge of pull request #28

PR #28, opened 2017-06-20
Aggregation and energy/power fix

mdeceglie:
This addresses issues #21 and #19. Includes updates to example and tests. It would be great to have this merged by PVSC!


rdtools/normalization.py
```diff
@@ -95,11 +96,39 @@ def normalize_with_pvwatts(energy, pvwatts_kws):
         freq = energy.index.freq
 
     dc_power = pvwatts_dc_power(**pvwatts_kws)
+    irrad = pvwatts_kws['poa_global']
+
+    # Length of each right labeled interval
+    model_tds = (dc_power.index - dc_power.index.shift(-1)).astype('int64') / (10.0**9 * 3600.0)
+    irrad_tds = (irrad.index - irrad.index.shift(-1)).astype('int64') / (10.0**9 * 3600.0)
+    measure_tds = (energy.index - energy.index.shift(-1)).astype('int64') / (10.0**9 * 3600.0)
```
abshinn: If I am reading this right, when you use `.astype('int64')` on a timedelta object, the result is in nanoseconds?

rdtools/normalization.py
```diff
@@ -118,11 +147,14 @@ def sapm_dc_power(pvlib_pvsystem, met_data):
         Measured irradiance components, ambient temperature, and wind speed.
         Expected met_data DataFrame column names:
             ['DNI', 'GHI', 'DHI', 'Temperature', 'Wind Speed']
+    Note: All series are assumed to be right-labeled
 
     Returns
```
abshinn: Same comment as above. This now returns a tuple containing Pandas Series.

rdtools/normalization.py
```diff
@@ -188,21 +217,50 @@ def normalize_with_sapm(energy, sapm_kws):
...
     if energy.index.freq is None:
         freq = pd.infer_freq(energy.index)
     else:
         freq = energy.index.freq
 
-    dc_power = sapm_dc_power(**sapm_kws)
+    dc_power, irrad = sapm_dc_power(**sapm_kws)
+
+    # Length of each right labeled interval
+    model_tds = (dc_power.index - dc_power.index.shift(-1)).astype('int64') / (10.0**9 * 3600.0)
```
abshinn: Is there a convenient way to refactor this logic so that it is not duplicated?

tests/aggregation_test.py
```diff
@@ -0,0 +1,31 @@
+import unittest
```
abshinn: Thanks for adding a unit test!

tests/aggregation_test.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def setUp(self):
+        ind = pd.DatetimeIndex(freq='12h', start='2015-01-01', end='2015-01-02 23:59')
+
+        self.insol = pd.Series(data=[500, 1000, 500, 1000], index=ind)
+        self.energy = pd.Series(data=[1.0, 4, 1.0, 4], index=ind)
+
+        self.aggregated = aggregation_insol(self.energy, self.insol, frequency='D')
+
+    def tearDown(self):
```
abshinn: No need to include a tearDown method if it is not being used. This method is useful if the variables defined in setUp are changed by the individual tests.

tests/normalization_pvwatts_test.py
```diff
@@ -40,7 +40,9 @@ def setUp(self):
         energy_index = pd.date_range(start='2012-01-01',
                                      periods=12,
                                      freq='MS')
-        dummy_energy = np.repeat(a=19.75, repeats=12)
+        power_meas = 19.75  # power in dummy conditions
+        hours = (energy_index - energy_index.shift(-1)).astype('int64') / (10.0**9 * 3600.0)
```
abshinn: I think pandas datetimeIndexes are phasing out use of the minus operator in newer versions, and instead include a `.difference()` method. But, this unit test will fail when that happens!

rdtools/aggregation.py
```diff
@@ -0,0 +1,24 @@
+import pandas as pd
+
+
+def aggregation_insol(normalized_energy, insolation, frequency='D'):
+    '''
+    Insolation weighted aggregation
```
mdeceglie: The flow we had mapped out at the outset was to aggregate after filtering (which is after normalization). We should all be open to re-evaluating that flow (I am!), but I suggest making that a separate (larger) discussion. I have added some explanation in the example notebook

rdtools/aggregation.py
```diff
@@ -0,0 +1,24 @@
...
+
+    Parameters
+    ----------
+    normalize_energy: Pandas series (numeric)
```
mdeceglie: Corrected, thanks.

rdtools/normalization.py
```diff
@@ -24,7 +24,7 @@ def pvwatts_dc_power(poa_global, P_ref, T_cell=None,
         Total effective plane of array irradiance.
     P_ref: numeric
         Rated DC power of array.
-    T_cell: Pandas Series (numeric)
+        T_cell: Pandas Series (numeric)
```
mdeceglie: nope, fixed now, thanks.

rdtools/normalization.py
```diff
@@ -34,6 +34,8 @@ def pvwatts_dc_power(poa_global, P_ref, T_cell=None,
     gamma_pdc: numeric, default is None
         Linear array efficiency temperature coefficient [1 / degree celsius].
 
+    Note: All series are assumed to be right-labeled
```
mdeceglie: That is correct. I have clarified this.

rdtools/normalization.py
```diff
@@ -82,11 +80,14 @@ def normalize_with_pvwatts(energy, pvwatts_kws):
             Reference temperature at standard test condition [degrees celsius].
         gamma_pdc: numeric, default is None
             Linear array efficiency temperature coefficient [1 / degree celsius].
+    Note: All series are assumed to be right-labeled
 
     Returns
```
mdeceglie: ✔︎

rdtools/normalization.py
```diff
@@ -82,11 +80,14 @@ def normalize_with_pvwatts(energy, pvwatts_kws):
...
     -------
     normalized_energy: Pandas Series (numeric)
-        Energy divided by Sandia Model DC energy.
+        Energy divided by PVWatts DC energy.
+    insolation:: Pandas Series (numeric)
```
mdeceglie: ✔︎

rdtools/normalization.py
mdeceglie: That's correct. Expanded the comments.

rdtools/normalization.py
mdeceglie: ✔︎

rdtools/normalization.py
mdeceglie: Good idea, I've made this a function.

tests/aggregation_test.py
mdeceglie: ✔︎

tests/normalization_pvwatts_test.py
mdeceglie: OK, we'll keep it in mind but I guess leave it for now.

mdeceglie:
Thanks for the review @abshinn! I've made updated based on your suggestions.

rdtools/normalization.py
cdeline: Mike - code looks good and I was able to run the example .ipny document.  My only question here is on units  - pvwatts_dc_power uses kW as default power input, and normalize_with_pvwatts uses Wh as the unit for energy.  I guess I would expect that if you're passing Wh in as energy, you should specify system power in W.  Or pass in energy in kWh and system size in kW.  If you keep things as-is, I would specify units explicitly in the docstring.

mdeceglie:
Thanks @cdeline. I think that if you stick to exclusively W or kW etc., you will always be safe. But to avoid any confusion I specified that everything should be in watts and updated the example notebook accordingly.

mdeceglie:
Merged. Thanks for the reviews @abshinn and @cdeline!

PR #38, opened 2017-08-23
update pvlib requirement

mdeceglie:
pvlib 0.4.1 had an error associated with `pvlib.location.Location.get_clearsky()` returning an ordered dict rather than a dataframe.


mdeceglie:
Let me know if anyone see's a problem with this minor update. Otherwise I will proceed in 24 hours.

PR #43, opened 2017-09-25
Combined update

mdeceglie:
Please provide a review by Friday, Sept. 29.

This update includes:

- clear sky workflow (including automatic rescaling of the clearsky model)
- exceedance values for YoY calculations
- filtering
- new consolidated example notebook.

**In reviewing please remember the distinction between the rdtools library and the example notebook. The example notebook is just intended to show people how to use the library. It is not intended to include specialized code for working with datasets other than the single example system considered.**


cdeline:
I know the review is focused on the module, but I'm getting an error in the .ipnyb example plots when it tries to plot color 'C1' in Clear Sky 4: Degradation Calculation.    I recommend specifying the color explicitly, either 'r' or 'orangered' if you want it something close to the original Broncos colors:

```
# Visualize the results

.....

ax2.hist(cs_yoy_values, label='YOY', bins=len(cs_yoy_values)/40, color =  'orangered')

.....

ax1.plot(clearsky_daily.index, clearsky_daily/cs_yoy_info['renormalizing_factor'], 'o', alpha = 0.5, color = 'orangered')

```

cdeline:
I've updated setup.py with the new NREL github repo URL.  
We haven't updated readme.md in a while - it's looking pretty bare.  Anyone want to take a stab at improving it?

GregoryKimball:
Thank you for the excellent work.  I'm reviewing a group of sites and generally the behavior is good.

Few suggestions:

- The example notebook should use the clear sky index filter
- Sites with low initial performance ratio show poor normalizing behavior, but I think we've decided that issue is out of scope.
image
- We do not have any functions for helping users correct clock errors such as drifts, shifts or inconsistent usage of daylight savings time.  I think this function is in scope of rdtools, but perhaps implemented later.
- Site with outages where performance is still > 0 don't filter well.  This is a problem with the original 30% rolling outage filter as well as the non-filtered rdtools case.
image

mdeceglie:
Thanks for you're comments @GregoryKimball.
1) The example does include the clear sky filter under 'Clear Sky 2: Filter' in the clear sky portion of the notebook toward the bottom. Unless I misunderstand your comment?

2) Yes, we've decided that advanced useres can use `degradation_year_on_year(recenter = False)` and recenter in a different way if needed.

3) This is a good suggestion I will open an issue on it.

4) Also an important challenge. Let's keep robust outlier filter development on the table. For now I will add extra warnings to the example to communicate that not all outages are caught with the current filters.

Thanks again for your feedback and contributions!

dirkjordan:
Great comments Greg. I have been using a wider csi window of 0.15 in my analysis and have found pretty consistent results. I didn't see much change by going to 0.2, but then we don't have a lot of badly drifting sensors. 
For the timezones, I think Mike will open a new issue on this. We'll also add some notes to guide the users. I had some issues with daylight savings time as well and how you need to specify it.
We will probably add a note too that the outage filter is still under development.   I have experimented with outages. I took a good data set and introduced non-zero outages. I generally found if there are less than 10% of the data missing or having outages very little effect but above 10% one does get some bias. That's maybe another note we can add to guide the users that if you have too many outages, missing data then the analysis might be biased.

PR #46, opened 2017-10-04
Issue 45 update dependencies in setup.py

abshinn:


setup.py
```diff
@@ -21,20 +21,20 @@
...
 
 INSTALL_REQUIRES = [
     'numpy >= 1.11.0',
     'pandas >= 0.19.0',
     'pvlib >= 0.5.0',
     'statsmodels >= 0.8.0',
     'patsy >= 0.4.1',
-    'nose >= 1.3.7',
+    'h5py >= 2.7.1',
     'pytz',
```
mdeceglie: Let's add `'scipy >= 0.19.1'`. I think it will generally be installed with `statsmodels`, but we use it explicitly in `normalization.py` so probably best to specify it.

abshinn:
Thanks, Mike. Suggested changes have been made. When running the tests, I noticed that tests/aggregation_test.py was not included in the tests/run_tests script. I went ahead and made that fix as well.

mdeceglie:
Thanks @abshinn. Everything looks good to me and works fine in testing on my end. I think we're good to merge.

Issue #50, opened 2017-11-02
Failed building wheel on install


mdeceglie:
The issue appears to be related to the `tests_require` argument in `setup()` being a list of an empty string. Removing that argument should fix the problem.

mdeceglie:
Fixed with pull request #51

PR #51, opened 2017-11-02
Fix issue #50 with building wheels on install

mdeceglie:


mdeceglie:
Great, thanks for taking a look @abshinn. What are your thoughts on rolling this out. Should we issue a v.1.1.2 patch? Or wait until we've collected more substantial changes? (or something else...?)

abshinn:
The only reason to hold-off on a v.1.1.2 patch is if there are other `pip install rdtools` problems that should to be addressed. 

I ran `pip install rdtools` in a clean environment. Even with the wheel error, the tests seem to pass. The only thing I noticed is that compared to our requirements in setup.py, newer versions for pandas (0.21.0), numpy (1.13.3), and scipy (1.0.0) were all installed. At the moment I do not see this as an issue. However, if at any time these packages make an API change that causes an error in rdtools, we can then make the fix and update rdtools requirements.

Since it looks like the tests_require was the only issue, I am in favor of a v.1.1.2 patch.

mdeceglie:
Thank you for pointing out the newer versions issue. After some experimenting, we need to do some more cleaning up on this front. The minimum versions listed in `setup.py` are actually not sufficient (I think they got updated in `requirements.txt` but not `setup.py`. Upgrading to pandas 0.21.0 also breaks some plots in the example. For now, I think we can make install require pandas 0.20.3 (and some appropriate version of numpy tdb), then open an issue to address 0.21.0. I will make the changes and open the issue.

mdeceglie:
@abshinn would you mind taking another quick look before we release v.1.1.2? I updated the pandas and numpy requirements, specifying `pandas == 0.20.3`. Version 0.21.0 causes problems with the example notebook. I have tested both the notebook and the tests with the minimum requirements specified in `setup.py` and with every specified package upgraded to the latest release (with the exception of pandas). I will capture this testing procedure for releases in the wiki.

mdeceglie:
@abshinn If you haven't already started reviewing the most recent changes, please hold off, I found the issue with pandas 0.21.0 and will modify the notebook accordingly. I will push the changes shortly.

mdeceglie:
Thanks for you patience @abshinn, I think we are good to move forward with releasing v.1.1.2 at this point. Please run the current version through your testing and confirm that everything works. In case you are interested, I have updated the wiki with the test procedure that I have developed.

abshinn:
Using this branch, I installed rdtools via `pip install .` in a fresh virtual environment, with package versions:

```
numpy==1.13.3
pandas==0.21.0
pvlib==0.5.1
statsmodels==0.8.0
scipy==1.0.0
```

Tests pass, and the example notebook runs without a problem. This looks good to me.

PR #63, opened 2018-01-26
Issue 59 confidence interval

mdeceglie:
This update gives users the ability to specify the size of confidence intervals for degradation calculations. I also changed the example to use a 95% confidence interval.


dirkjordan:
I'm not sure that changing the CI for the example notebook is a good idea. Determining degradation rates at 95% is extremely difficult. Essentially you are doubling all the uncertainties. I have system that have been in the field for almost 10 years, the doubling of the uncertainties leads to the conclusion that after all this field exposure you are still not sure whether it is degrading or not. This may be technically correct but useless to analysts, investors, independent engineers etc. If you are uncomfortable with 68% (1 standard deviation) we talk about other CI like 80% but 95% is too large in my opinion.

mdeceglie:
Thanks @dirkjordan, I have changed the confidence interval back to 68.2 in the example. If we get a strong message from users that a different value would be more appropriate we can change it down the road.

GregoryKimball:
I just added some basic tests for confidence interval results.  They are almost all passing except for the test that `cls.rd` fall within the 95% confidence interval.  Do you know why it doesn't?

mdeceglie:
Thanks for adding the tests @GregoryKimball.

I think it is failing for 2 reasons:
1) `cls.rd` is in fractional units and should be scaled up to %/year for comparison. (This is fixed in my last commit)
2) I think the compounding degradation in `get_corr_energy()` doesn't square with a linear interpretation of degradation rate  that the degradation calculations use.

Have to think about the best way to fix 2

mdeceglie:
I addressed the failing test by redefining the `get_corr_energy()` method. I think there were a few problems. One was the compounding degradation in the old version didn't have a clear interpretation in terms of a linear rate. Another is that there were some approximations, such a 52 weeks = 1 year that made the rate inexact. Finally, I added some noise so the confidence intervals are meaningful.

It still doesn't work for degradation_classical_decomposition, we should think about deprecating classical decomposition anyway, I will open an issue. 

@GregoryKimball, what do you think?

mdeceglie:
Good idea to add the random seed. 

Regarding the error, if we remove the noise, both OLS and YOY give exactly -0.5% which is reassuring. But if you uncover an underlying problem, it would be interesting to hear about (and fix!)

Thanks for your help with this pull request. @GregoryKimball, are you comfortable with merging?

GregoryKimball:
Wow!  Exactly -0.5% works for me!  Let's merge.

mdeceglie:
Great, thanks everyone!

PR #64, opened 2018-02-05
Irradiance rescale stability

mdeceglie:
I have updated irradiance_rescale with a more stable iterative calculation method. This may affect results that people get, so I have it issue a warning if someone doesn't specify what method to use.


mdeceglie:
Currently I issue a `UserWarning`, should we change this to `SyntaxWarning` or `FutureWarning`?

rdtools/normalization.py
```diff
@@ -282,20 +287,74 @@ def irradiance_rescale(irrad, modeled_irrad):
...
-    min_result = minimize(_rmse, guess, method='Nelder-Mead')
-    factor = min_result['x'][0]
-
-    out_irrad = factor * modeled_irrad
-    return out_irrad
+
+    if method is None:
+        warnings.warn("The underlying calculations for irradiance_rescale have changed "
+                      "which may affect results. To revert to the version of irradiance_rescale "
+                      "from rdtools v1.1.3 or earlier, use method = 'single_opt'. ")
```
abshinn: This is a nice way to keep the code backwards compatible while warning that the default behavior has changed. In this situation, I think either UserWarning or SyntaxWarning are appropriate.

FutureWarning is most applicable in the case where a code feature is slated to be removed in a future version -- if we decide that we no longer wish to support `method='single-opt'`, for example.

rdtools/normalization.py
```diff
@@ -282,20 +287,74 @@ def irradiance_rescale(irrad, modeled_irrad):
...
+            filt = (csi >= 0.8) & (csi <= 1.2) & (irrad > 200)
+            min_result = minimize(_rmse, guess, method='Nelder-Mead')
+
+            factor = min_result['x'][0]
+            return factor
+
+        # Calculate an initial guess for the rescal factor
+        factor = np.percentile(irrad.dropna(), 90) / np.percentile(modeled_irrad.dropna(), 90)
+
+        # Itteretively run the optimization, recalculating the clear sky filter each time
```
abshinn: Spelling error: itteretively

rdtools/normalization.py
```diff
@@ -282,20 +287,74 @@ def irradiance_rescale(irrad, modeled_irrad):
...
+            "Optimizes rescale factor once"
+            global filt
+            csi = irrad / (guess * modeled_irrad)  # clear sky index
+            filt = (csi >= 0.8) & (csi <= 1.2) & (irrad > 200)
+            min_result = minimize(_rmse, guess, method='Nelder-Mead')
+
+            factor = min_result['x'][0]
+            return factor
+
+        # Calculate an initial guess for the rescal factor
```
abshinn: Spelling error: rescal

mdeceglie:
Thanks for the reviews @cdeline and @abshinn

Issue #67, opened 2018-02-22
degradation notebook in docs doesn't work


mikofski:
* trace back message was "no module named normalization" raised on the 1st line of `rdtools/__init__.py`
* looking in `rdtools/` the file `normalization.py` does exist, but looking in `rdtools/__init__.py` shows that it was imported using a relative path, it should use an absolute path or a dot.

mdeceglie:
Fixed in development branch with merge of PR #72

Issue #61, opened 2017-12-19
AmbiguousTimeError in get_clearsky_tamb


mdeceglie:
minimum code example:

```
import pandas as pd
import rdtools
times = pd.date_range('2012-11-12', '2014-01-01', freq='H', tz='America/New_York')
rdtools.get_clearsky_tamb(times, 40.8, 74.0)
```

traceback:
```
  File "<stdin>", line 1, in <module>
  File "/anaconda2/lib/python2.7/site-packages/rdtools/clearsky_temperature.py", line 37, in get_clearsky_tamb
    dt_daily = pd.date_range(times[0] - buffer, times[-1] + buffer, freq='D')
  File "/anaconda2/lib/python2.7/site-packages/pandas/core/indexes/datetimes.py", line 2042, in date_range
    closed=closed, **kwargs)
  File "/anaconda2/lib/python2.7/site-packages/pandas/util/_decorators.py", line 91, in wrapper
    return func(*args, **kwargs)
  File "/anaconda2/lib/python2.7/site-packages/pandas/core/indexes/datetimes.py", line 317, in __new__
    ambiguous=ambiguous)
  File "/anaconda2/lib/python2.7/site-packages/pandas/core/indexes/datetimes.py", line 526, in _generate
    ambiguous=ambiguous)
  File "pandas/_libs/tslib.pyx", line 4580, in pandas._libs.tslib.tz_localize_to_utc (pandas/_libs/tslib.c:77701)
pytz.exceptions.AmbiguousTimeError: Cannot infer dst time from Timestamp('2012-11-04 01:00:00'), try using the 'ambiguous' argument
```

proposed solution in line 37:
`  dt_daily = pd.date_range(times.date[0] - buffer, times.date[-1] + buffer, freq='D')`

mdeceglie:
Thanks @amnag . Please go ahead and create a pull request to the **development** branch.

amnag:
Hi @mdeceglie and @cdeline, I added notebooks to rdtools/docs that read in csv/pickle files to calculate degradation rates. These notebooks are up-to-date with the python code that @mdeceglie  used on the entire set on merged files. There are two versions, one is for NREL users and the other is public-facing. I added a couple of pickle files to the repo to enable the public notebooks to run, and also wanted to add a couple of csv files. But .gitignore prohibits addition of csv files. Is it OK to modify .gitignore to add a couple of csv files to the repo ?

mdeceglie:
[...]

Issue #59, opened 2017-12-15
Allow control of confidence interval size in degradation calculations


mdeceglie:
Users should have the ability to set the size of the confidence interval for degradation calculations.

Issue #57, opened 2017-12-14
Problem with get_clearsky_tamb function in rdtools/clearsky_temperature.py


amnag:
The get_clearsky_tamb function in rdtools/rdtools/clearsky_temperature.py still fails to work unless the line:

freq_actual = pd.infer_freq(times) 

is changed to to

freq_actual = pd.infer_freq(times) if pd.infer_freq(times) != None else pd.infer_freq(times[:10])

or to 

freq_actual = times.freqstr

Please see new rdtools branch fix_freq_inference

mdeceglie:
@amnag please take a look at my proposed solution in the branch `issue_57` and let me know what you think. (The branch `fix_freq_inference` appeared to have unrelated changes to the notebooks, so I started a new branch.)

mdeceglie:
This has been addressed in pull request #61 to `development`. We can close this once it it is merged to master in the next release.

PR #73, opened 2018-03-23
Update master for release

mdeceglie:
This incorporates changes that address issues: #67, #61, #59, #57, #56 as well as:
- improvements/typo fixes to docstrings
- Fixes error in check for two years of data in `degradation_year_on_year`
- Improves the calculations underlying `irradiance_rescale`

All changes have ben through review when incorporated into the `development` branch. The purpose of this PR is to check that this version of rdtools installs and runs correctly.

For me, it passes everything in the [release procedure](https://github.com/NREL/rdtools/wiki/Release-procedure), with the exception of the minimum requirement check on windows, but failure was due to problems building old versions of some requirements.


cdeline:
I get installation errors in new environments of Python 2.7 and Python 3.6 with my Win10 machine.  When installed into my existing environment, there are no errors.

Python 2.7: 
from numpy.distutils.misc_util import get_info
   ModuleNotFoundError: No module named 'numpy'

Python 3.6:
from numpy.distutils.misc_util import get_info
   ImportError: No module named numpy.distutils.misc_util

This appears to be a known bug with statsmodels for Windows 10 machines:

https://github.com/statsmodels/statsmodels/issues/3207

https://stackoverflow.com/questions/45492810/statsmodels-installation-no-module-named-numpy-distutils-msvccompiler-in-num

image

image

cdeline:
Update:  if I use 'conda install statsmodels' followed by 'pip install .' to install the development branch, then everything works in py2.7 and py3.6.  Is there any way to install statsmodels through conda rather than pip in our setup.py file?

mdeceglie:
@cdeline I was able to recreate this in a clean environment with only `pip` and `python`. However, I suggest we can proceed without directly addressing this. If a user has installed the anaconda python distribution (that's the most likely way they would get the `conda` package manager), they will already have `statsmodels` installed. If they are setting up environments, hopefully they have enough experience to navigate the requirements gymnastics.

We could add a note in the readme encouraging people to install requirements by hand if they run into trouble. Do you think that would help?

cdeline:
OK, I'm good with this approach.  If you could include an extra line in the readme about manually installing components using conda if they get errors during install, I'm fine with this.
I suspect that we'd still have installation issues if someone has a fresh non-Anaconda python distribution and just tries 'pip install rdtools',  We don't specifically require Anaconda in the readme anywhere.

mdeceglie:
Thanks for the input everyone!

PR #81, opened 2018-07-03
Issue 49 test runner

abshinn:
The goal of this Pull Request is to switch to using Python unittest module's TestSuite framework for running unit tests.

The motivation for doing this is to turn the test runner script to a python script rather than a bash script, and make it easier to profile and measure coverage of the unit tests (making use of cProfile and coverage python modules).


rdtools/clearsky_temperature.py
```diff
@@ -70,9 +70,9 @@ def get_clearsky_tamb(times, latitude, longitude, window_size=40, gauss_std=20):
         day = _get_pixel_value(a, lon_index, lat_index, k, radius)
         night = _get_pixel_value(b, lon_index, lat_index, k, radius)
 
-        if day == float("NaN"):
+        if pd.isnull(day):
             day = a[:, lat_index, k]
-        if night == float("NaN"):
+        if pd.isnull(night):
```
GregoryKimball: Thank you for finally fixing this!!!

abshinn: No problem.

tests/filtering_test.py
```diff
@@ -0,0 +1,93 @@
...
+        filtered = tcell_filter(self.tcell,
+                                low_tcell_cutoff=-50,
+                                high_tcell_cutoff=110)
+
+        # Expected high and low tcell cutoffs to be non-inclusive.
+        expected_result = np.array([False, True, True, True, False])
+        self.assertListEqual(filtered.tolist(), expected_result.tolist())
+
+
+class ClipFilterTestCase(unittest.TestCase):
```
mdeceglie: I'd like to try and define two tests here, one to test the high limit and one for the low, but totally indepedently. Perhaps something like this:
```
class ClipFilterTestCase(unittest.TestCase):
    ''' Unit tests for inverter clipping filter.'''

    def setUp(self):
        self.power = pd.Series(np.arange(1, 101))
        # Note: Power is expected to be Series object because clip_filter makes
        #       use of the Series.quantile() method.

    def test_clip_filter_upper(self):
        filtered = clip_filter(self.power, quant=0.98,
                               low_power_cutoff=0)

        # Expect 99% of the 98th quantile to be filtered
        expected_result = self.power < (98 * 0.99)
        self.assertTrue((expected_result == filtered).all())

    def test_tcell_filter_low_cutoff(self):
        filtered = clip_filter(self.power, quant=0.98,
                               low_power_cutoff=2)

        # Expect power <=2 to be filtered
        expected_result = (self.power > 2)
        self.assertTrue((expected_result.iloc[0:5] == filtered.iloc[0:5]).all())
```
Note if we use this verbatim in negates my test name comments below.

abshinn:
Thanks, @mdeceglie. I plan to implement your suggestions. I will also replace the runner script by adding the pytest module. Unfortunately we will lose the `--coverage` kwarg, but I figure that is okay since we want to move toward automated coverage testing anyway.

tests/filtering_test.py
abshinn: Done.

tests/filtering_test.py
```diff
@@ -0,0 +1,93 @@
+""" Filtering Module Tests. """
+
+import unittest
+import sys
```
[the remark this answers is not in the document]
abshinn: Ah, good catch.

abshinn:
@mdeceglie Your suggested changes have been made.

If all looks good to you: between this PR and PR #82, I do not have a strong preference on merge order. Either way, the second branch will need to change the way TravisCI runs the tests (it is now set up so pytest is installed and run through the setup.py script, i.e., `python setup.py test`).

Finally, the release procedure wiki will need to be updated. Since tests are run automatically on TravisCI (for min/max versions, for Python 2/3), then my thought is that we can remove steps related to the unit tests.

rdtools/test/filtering_test.py
```diff
@@ -66,26 +65,20 @@ def setUp(self):
...
-        self.assertFalse(filtered.all())
+                               low_power_cutoff=0)
 
-        # Expect 99% of the 98th quantile to be cut off.
+        # Expect 99% of the 98th quantile to be filtered
         expected_result = self.power < (98 * 0.99)
         self.assertTrue((expected_result == filtered).all())
 
-    def test_tcell_filter_cutoff(self):
+    def test_tcell_filter_low_cutoff(self):
```
mdeceglie: This test simultaneously checks both the upper and lower filtering. If something breaks the upper side filtering both the upper and lower tests will fail, and this may confuse the developer trying to diagnose the problem. I suggest either implementing my suggestion to only check the first ~5 elements of the expected result, or add detailed comments to help developers interpret the test results.

mdeceglie: should be `test_clip_filter_low_cutoff()`

mdeceglie:
Thanks everyone!

Issue #76, opened 2018-04-11
set up CI server  (like Travis) for testing


mikofski:
# Current State

* there is no continuous integration to test, build, and deploy rdtools
* there is an undue burden on maintainers, contributors, and users to manually perform tests on all platforms and Python versions that is error prone
* merging pull requests requires manual testing on different systems that may have unknown environments, package versions, `PATH` clashes, and other detritus that make testing flawed, producing false results
* there is an undue burden on maintainers to manually release new versions to PyPI, Anaconda, and GitHub releases

# Proposal

* set up a free [TravisCI](https://travis-ci.org/) account to test both Python 2.7 and Python 3.6
* since rdtools is pure python this server is sufficient for all platforms
* add a section to `setup.cfg` that sets wheels to be universal (py2 & py3)
    ```
    [bdist_wheel]
    universal = 1
    ```
* add a section to the `.travis.yml` file to build wheels and archives and then deploy them to GitHub and PyPI

# Future work

* add a `requirements.txt` file at the top level so that CI servers and IDE apps can infer the virtual 
environment or start using [`pipenv`](https://docs.pipenv.org/)
* set up conda forge to build and release rdtools to Anaconda automatically from GitHub

# examples of `.travis.yml`

* [pvmismatch](https://github.com/SunPower/PVMismatch/blob/master/.travis.yml)
* [pecos](https://github.com/sandialabs/pecos/blob/master/.travis.yml)
* [pvlib-python](https://github.com/pvlib/pvlib-python/blob/master/.travis.yml)
* [SolarUtils](https://github.com/SunPower/SolarUtils/blob/master/.travis.yml)
* [Carousel](https://github.com/SunPower/Carousel/blob/master/.travis.yml)
* [UncertaintyWrapper](https://github.com/SunPower/UncertaintyWrapper/blob/master/.travis.yml)

# Travis CI docs and CLI

* [docs](https://docs.travis-ci.com/)
* [CLI](https://github.com/travis-ci/travis.rb#installation) only necessary to create secure password hashes, alternately use the web UI to create hidden secure credentials

abshinn:
CI is a great idea, @mikofski. Thanks for suggesting. I have experience with [Concourse](https://concourse-ci.org/) for private repositories. I agree that TravisCI is the best option here.

Do you have the bandwidth to set this up? If not, I am happy to take it on and loop you in for review.

[...]

PR #82, opened 2018-07-12
Setup Travis for Continuous Testing

abshinn:
This PR sets up Travis for continuous testing on branches and pull-requests. This is also a first step to automated deployment (if we choose to do so).


mdeceglie:
Thanks for advancing this @abshinn. Is there a way we can include automated testing of both the minimum and maximum requirements specified in `setup.py`?

abshinn:
@mdeceglie Reading up on best practices on dependencies, I came across [this blog post](https://blog.miguelgrinberg.com/post/the-package-dependency-blues). It turns out setup.py and requirements.txt have different purposes. The purpose of `install_requires` within setup.py is to help pip to sort out dependencies if RdTools is used by another package, which is why it is necessary to specify a range. It is also recommended to only list direct dependencies (for example if numpy is required by pandas but not called anywhere in the code, it is not a direct dependency). The purpose of requirements.txt, however, is to provide an exact `==` dependency list, explicitly stating the list of dependency versions that have been tested.

It seems like we do want to maintain a requirements.txt, which represents the minimum requirements known to be stable. This is also supposed to be the set environment that we want to test in TravisCI.

For the requirement range in `install_requires`, we should not only use a lower bound, but also use an upper bound whenever possible. This would allow us to indicate that we have tested pandas 0.20.3 to 0.23.3, and not allow RdTools to work for anything newer (this would be a little restrictive, but not unreasonable since pandas is still in version 0.x.x).

So instead of setting up a testing matrix using both minimum and maximum requirements, what do you think of testing the minimum using requirements.txt, while also including upper bounds within `install_requires`?

mdeceglie:
Thanks for finding that @abshinn, I found it helpful.  

I think our project would be classified as reusable components that ship with an example application. For this case, the author recommends 

> Include a `requirements.txt` file for the example, so that at least there is a record of a set of specific versions that are known to work with your component.

This makes a lot of sense for us. I support adding a requirements.txt file to document the environment known to work with our example(s). Particularly useful for things like `matplotlib` which aren't required directly for `rdtools` but are needed for the example notebook.

I would still prefer to automatically test the min and max requirement ranges in setup.py, which seems like it should be possible. That leaves two open questions:  

1. How to get  @travis-ci to test the min and max requirements. I made a new branch illustrating the current problem `setup_CI_md`, the problem can be seen [here](https://travis-ci.org/NREL/rdtools/branches) (also, more info in issue #76). @mikofski, any ideas on this? If no one has any insights, perhaps I will open an issue directly with @travis-ci. 

2. The question of whether to include an upper limit in `install_requires`. The author of that post is a strong advocate for it, but I worry about the upper limit becoming out of date and wreaking havoc with peoples environments on install. For example I notice that [pvlib-python](https://github.com/pvlib/pvlib-python) doesn't include upper limits. And if we succeed with 1 above, we'll catch problems when other packages release.

mdeceglie:
Update: I resolved the problem with environment variables not getting set. The problem was spaces around `=` in the `env` section of `.travis.yml`

Builds are now passing for min and current requirements in the [`setup_CI_md`](https://github.com/NREL/rdtools/tree/setup_CI_md) branch.

mdeceglie:
I'm still not sure that we'd be using `requirements.txt` properly. The purpose of `install_requires` is explicitly to document minimum requirements. Check out https://packaging.python.org/discussions/install-requires-vs-requirements/  I'll bake an example in another branch and we can use that to guide further discussion.

abshinn:
@mdeceglie I am sorry, I think my wording in bullet number 1 of the proposal above was confusing. I edited the proposal. We can discuss over the phone if more efficient.

Here is my thought process on how we should use the two requirements lists for our purposes:

My thinking is that `requirements.txt` should be a complete list all dependencies for RdTools, with exact versions, in order to create a repeatable environment across TravisCI builds. It serves the same purpose as `min_reqs.txt` in your branch -- a snapshot of the environment that we know RdTools will work in.

Quoting the link you provided above:
> Whereas install_requires requirements are minimal, requirements files often contain an exhaustive listing of pinned versions for the purpose of achieving repeatable installations of a complete environment.

My interpretation of `install_requires` is to list the direct/minimal dependencies, allowing a range of versions (with at least a lower bound) to allow RdTools to exist as a reusable component within a user's environment (that may have any number of combinations of dependency versions). I like the idea of simultaneously testing the latest requirements (`pip install -U .`) so that we can be alerted if a dependency is updated on PyPi containing interface changes that break RdTools.

mikofski:
I agree with the previous comments that `requirements.txt` should use `pkg==X.Y.Z` and set just the direct dependencies for the `rdtools` project that are known to work. In my experience, setting up a consistently repeatable virtual environment of exact versions avoids a lot of problems, and this was my main reason for submitting #76.

Nice work @mdeceglie on setting up matrix builds in your branch! Make sure to add a commit that says, "closes #76" so the issue is closed by GitHub automatically when you merge #82.

I also agree it's not necessary to set upper limits for your `install_requires` unless you know for a fact that `rdtools` doesn't work with a newer version of a dependency. Unless there's an issue a specific version, just except somewhat on faith that most packages you use are stable and newer versions will be backward compatible. The one notable exception is pvlib.

IMO you probably should add the Jupyter notebook requirements somewhere, perhaps in `jupyter-reqs.txt` with a note in the docs. I think you need to help users a little to have a consistent experience, or they will get discouraged or loose confidence.

I think you guys have done tremendous work! Thanks!

mdeceglie:
Thanks for clarifying that @abshinn, and for weighing in @mikofski. I have a better understanding of the proposal now, and think it makes sense. Practically, does it make sense to keep the `requirements.txt` synced with the minimum version numbers specified in `setup.py`? (I guess this is where I got confused before).  If so, what should our practice be for keeping those in sync? One option would be to add something to setup.py, such that it could be called with `python setup.py --generate_requirements` and would just write `requirements.txt` with exact versions specified (`==`) then exit without installing anything. For example:
```python
if '--generate_requirements' in sys.argv:
    # Write requirement file and exit
    with open('requirements.txt', 'w') as f:
        for r in INSTALL_REQUIRES:
            f.write(r.replace('>=', '==') + '\n')
    sys.exit()
```
Performing that could be part of the release procedure? Or we could add an explicit unit test that would let us know when we need to manually take a look?

I'm also on board with a separate requirements file for running all our examples (hopefully someday we will have more than one).

mikofski:
@mdeceglie often people just use `pip freeze > requirements.txt` since this works well enough, true, it dumps indirect dependencies too, so perhaps not ideal. In practice, generating minimum requirements isn't something you have to do very often, so even though it's not DRY, IMO it's not a huge technical debt to do it manually. IMO they do seperate things, `requirements.txt` is for creating _exact_ environments, and `install_requires` are for distributing your package to _any_ environment, so although not dissimilar, they accomplish different goals. EG: it's not uncommon to have more than requirements file for different environments that you may use for testing, for docs, for your notebook, for different Python versions, etc. IMO, just don't overthink it.

wholmgren:
I'll throw in my 2 cents since pvlib was mentioned...

> The one notable exception is pvlib.

I agree that it's a good idea to restrict your pvlib versions to one 0.x release cycle, e.g. >= 0.5.0 and < 0.6. This is what I do for my own internal packages even though I control the pvlib releases! I wouldn't worry about it for more mature packages like numpy/pandas/etc.

I mostly copied MetPy and XArray when implementing pvlib's CI setup a few years ago. Those are my go-to libraries to see how other projects approach things. They are in a nice not too big, not too small, not too complicated space.

Finally, in my opinion, pytest is a huge improvement over the built-in unittest framework.

abshinn:
Much thanks for the advice @mikofski and @wholmgren! I will be sure to put an upper limit on pvlib, and start using pytest in PR #81.

mdeceglie:
Thanks everyone. @abshinn , I think this is ready to go please let me know if you see any outstanding issues.

mikofski:
Two more things:

1. add a [`deploy` to pypi section to your `.travis.yml`](https://docs.travis-ci.com/user/deployment/pypi/):

        deploy:
          provider: pypi
          user: "Your username"
            secure: "Your encrypted password"

   please don't commit any personal secrets to your github repo, so you shouldyou could [use the travis CLI to generate a secure key](https://docs.travis-ci.com/user/environment-variables#defining-encrypted-variables-in-travisyml), or use the [travis settings page online to set a hidden environment variable](https://docs.travis-ci.com/user/environment-variables#defining-variables-in-repository-settings), _eg_: `$PYPI_PASSWORD`, which you then use in your `.travis.yml`, you can use [`sunpower/pvmismatch/.travis.yml`](../../../SunPower/PVMismatch/blob/master/.travis.yml) or [`breakingbytes/simkit/.travis.yml`](../../../BreakingBytes/simkit/blob/master/.travis.yml) as examples

2. add the [build badge from travis](https://docs.travis-ci.com/user/status-images/) to your readme so that users will know that it's tested and the state of the build (pass/fail/unknown).

abshinn:
@mdeceglie -- thanks for updating the branch! Looks like the [latest build on TravisCI](https://travis-ci.org/NREL/rdtools/builds/417403534) is doing what we want it to (testing minimum and maximum requirements on both Python 2 and 3), so I think this is ready to go.

@mikofski -- totally agree. I will go ahead and add a badge with `branch=development` for now. When it comes time to release to master, I think we'll want the readme to show the status of `branch=master`. I will create a separate issue for setting up continuous deployment.

PR #87, opened 2018-08-21
Issues 78 79 80

mdeceglie:
This PR addresses a few bugs and documentation improvements.


rdtools/test/normalization_pvwatts_test.py
```diff
@@ -83,6 +83,12 @@ def test_normalization_with_pvw(self):
         # Test corrected energy is equal to 1.0
         self.assertTrue((corr_energy == 1.0).all())
 
+        # Test expected behavior when energy has no explicit frequency
```
abshinn: Great to test this. This test does not cover the case where freq is None and energy time series is not inferable or irregular. As mentioned above, a `with self.assertRaises(ValueError):` or `with pytest.raises(ValueError):` context would do the trick.

rdtools/normalization.py
```diff
@@ -271,11 +277,22 @@ def normalize_with_sapm(energy, sapm_kws):
 
 
 def delta_index(series):
-    # Takes a panda series as input and returns (time step sizes, average time step size)
-    # Length of each interval calculated by using 'int64' to convert to nanoseconds
+    '''
+    Takes a panda series as input and returns (time step sizes, average time step size) in hours
```
abshinn: Is it worth noting here that the pandas series index must be a DatetimeIndex?

rdtools/normalization.py
```diff
@@ -236,6 +240,8 @@ def normalize_with_sapm(energy, sapm_kws):
 
     if energy.index.freq is None:
         freq = pd.infer_freq(energy.index)
+        if freq is None:
+            raise ValueError('Could not infer frequency of energy, which must be a regular time series')
```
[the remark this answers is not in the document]
mdeceglie: Agreed on both counts, I have made the changes

rdtools/normalization.py
mdeceglie: I think that it is, I have added that.

rdtools/test/normalization_pvwatts_test.py
mdeceglie: I've added the test for both pvwatts and sapm functions

PR #91, opened 2018-09-19
fix outdated docstring

mdeceglie:
This fixes an outdated docstring in `degradataion_year_on_year()`


mdeceglie:
Thanks @cdeline Sorry for the double PR (the first was mistakenly based off a branch other than `development` making for a confusing commit summary.

PR #104, opened 2018-11-20
Issue 103

mdeceglie:
Fixes an issue wherein insolation was summing to zero when irradiance was NaN. The fix requires an update of the minimum `pandas` version to 0.22.0. 

I've tested the change on the example notebook, and the results are unaffected.


mdeceglie:
@abshinn Great, thanks for the review!

PR #105, opened 2018-11-21
Step 0 improvements

mdeceglie:
These changes make step 0 in the example more robust and add a function for calculating energy time series from power time series, `energy_from_power()`. 

The relevant lines in the old version of the example notebook (step 0: Import and preliminary calculations) are:
```
# And then set the frequency of the dataframe
df = df.resample(freq).median()

# Calculate energy yield in Wh
df['energy'] = df.power * pd.to_timedelta(df.power.index.freq).total_seconds()/(3600.0)
```
Have changed to:
```
# And then set the frequency of the dataframe
df = df.resample(freq).interpolate(method='time', limit=1)

# Calculate energy yield in Wh
df['energy'] = rdtools.energy_from_power(df.power, 0.25)
```

The change in resampling method from median to interpolate with a limit of 1 provides more controlled resampling. NaN will be filled in if more than one datapoint is missing at the expected frequency.  

`energy_from_power(power_series, max_interval_hours)` provides functionality that will be needed in may RdTools analyses. The docstring for this new function is:

>     Returns a right-labeled energy time series from an instantaneous power time series.
>     Energy is not returned when the gap between data points exceeds max_interval_hours.
>     power_series should be given in Watts, and max_interval_hours in hours.
> 
>     Parameters
>     ----------
>     power_series: Pandas Series with DatetimeIndex
>         Instantaneous time series of PV power generation in Watts
>     max_interval_hours: float
>         The maximum allowed gap between power measurements. If the gap between
>         consecutive power measurements exceeds max_interval_hours, no energy value
>         will be returned for that interval
> 
>     Returns:
>     --------
>     right-labeled energy pandas time series in Wh

Note these changes result in the following changes to the output of the example notebook:

|                  | Original notebook output | New notebook output |
|:----------------:|:------------------------:|:-------------------:|
|        Rd        |           -0.43          |        -0.42        |
|     Rd CI low    |           -0.53          |        -0.50        |
|    Rd CI high    |           -0.33          |        -0.34        |
|        P95       |           -0.55          |        -0.54        |
|    Clearsky Rd   |           -0.05          |        -0.12        |
|  Clearsky Rd low |           -0.22          |        -0.31        |
| Clearsky Rd high |           0.14           |         0.07        |
|   Clearsky P95   |           -0.31          |        -0.42        |


rdtools/normalization.py
```diff
@@ -381,3 +381,37 @@ def check_series_frequency(series, series_description):
...
+    right-labeled energy pandas time series in Wh
+
+    '''
+
+    if not isinstance(power_series.index, pd.DatetimeIndex):
+        raise ValueError('power_series must be a pandas series with a DatetimeIndex')
+
+    times = np.array(power_series.index.astype('int64'))
+    time_deltas = np.diff(times) / 10.0**9 / 3600.0
+
```
cdeline:
```suggestion
if max_interval_hours is None:
    max_interval_hours = np.median(time_deltas) * 4
```

cdeline: PS:   `max_interval_hours = None`  also needs to be added to the function input if you do this.

rdtools/normalization.py
```diff
@@ -381,3 +381,37 @@ def check_series_frequency(series, series_description):
     else:
         freq = series.index.freq
     return freq
+
+
+def energy_from_power(power_series, max_interval_hours):
```
cdeline:
```suggestion
def energy_from_power(power_series, max_interval_hours = None):
```

rdtools/normalization.py
cdeline: Ok, sure - we can set the default gap at the median timestep.  I'll approve this.

mdeceglie:
Thanks @cdeline I'll incorporate those along with an appropriate modification to the docstring.

rdtools/normalization.py
mikofski: [pep8 - no spaces around equals for default keyword args](https://pep8.org/#other-recommendations)
>Don’t use spaces around the = sign when used to indicate a keyword argument or a default parameter value.
>
>Yes:

```python
def complex(real, imag=0.0):
    return magic(r=real, i=imag)
```
>No:

```python
def complex(real, imag = 0.0):
    return magic(r = real, i = imag)
```

cdeline: Fixed, thanks!

cdeline:
Looks good to me now.  Thanks Mike!

mdeceglie:
@cdeline great, thanks for the review!

mdeceglie:
On closer investigation `df.resample(freq).interpolate(method='time', limit=1)` doesn't behave as I expected. This example illustrates the problem:

screen shot 2018-11-30 at 4 08 50 pm

https://gist.github.com/mdeceglie/1c641401e76ad4b10bfb6751fd22b899

mdeceglie:
I've updated this branch with a new, robust method for time series interpolation and resampling: `rdtools.interpolate()`. Using that in the example notebook (along with `energy_from_power()`) gives the following results relative to `master`:


|                  | Original notebook output | New notebook output |
|:----------------:|:------------------------:|:-------------------:|
|        Rd        |           -0.43          |        -0.43        |
|     Rd CI low    |           -0.53          |        -0.52        |
|    Rd CI high    |           -0.33          |        -0.36        |
|        P95       |           -0.55          |        -0.55        |
|    Clearsky Rd   |           -0.05          |        -0.12        |
|  Clearsky Rd low |           -0.22          |        -0.30        |
| Clearsky Rd high |           0.14           |         0.08        |
|   Clearsky P95   |           -0.31          |        -0.42        |


Here is an illustration of rdtools.interpolate() relative to other methods we have considered:
screen shot 2018-12-04 at 4 46 05 pm
full example: https://gist.github.com/mdeceglie/1c641401e76ad4b10bfb6751fd22b899 

This shows some of the problems with using `df.resample().median()` or `df.resample().interpolate()`. On the other hand `rdtools.interpolate()` can handle both arbitrary (including irregular) input and target indexes. It also provides explicit control over how large of a gap should be interpolated through.

@cdeline @dirkjordan Would you guys mind taking another look?

mdeceglie:
In chasing down the discrepancy on the clear-sky results I realized that the new `energy_from_power()` function should be utilized by the normalization functions, so I've made that change. I have also uncovered issue #107, which is now addressed in this branch with a reworked version of `energy_from_power()`. 

The summary of differences for the notebook is now:

|                  | Original notebook output | New notebook output |
|:----------------:|:------------------------:|:-------------------:|
|        Rd        |           -0.43          |        -0.43        |
|     Rd CI low    |           -0.53          |        -0.52        |
|    Rd CI high    |           -0.33          |        -0.36        |
|        P95       |           -0.55          |        -0.58        |
|    Clearsky Rd   |           -0.05          |        -0.15        |
|  Clearsky Rd low |           -0.22          |        -0.32        |
| Clearsky Rd high |           0.14           |         0.00        |
|   Clearsky P95   |           -0.31          |        -0.39        |

Most of the change comes from adoption of the new `energy_from_power()`. The use of the new interpolation function has negligible impact on the results.

Given the flaws in the prior version, I suggest we should adopt these changes even though the example notebook results shift slightly. The confidence intervals, even at the 68.2% level, still overlap substantially. @cdeline and @dirkjordan, please take a look and let me know what you think.

@abshinn I have also made minor adjustments to some of the tests for `normalize_with_pvwatts()` and `normalize_with_sapm()`. Please take a look and make sure they make sense.

Given the changes to the calculation, this would warrant a new major rdtools version.

rdtools/normalization.py
```diff
@@ -383,40 +375,270 @@ def check_series_frequency(series, series_description):
...
+    if target_frequency is None:
+        target_frequency = str(median_step_ns) + 'N'  # Pandas offset allias in ns
+
+    if max_timedelta is None:
+        max_interval_nanoseconds = median_step_ns
+    else:
+        max_interval_nanoseconds = max_timedelta.total_seconds() * 10.0**9
+
+    try:
+        freq_interval_size_ns = pd.tseries.frequencies.to_offset(target_frequency).nanos
```
cdeline: I'm getting an error here when I run the example notebook: 
`df['energy'] = rdtools.energy_from_power(df.power, max_timedelta=pd.to_timedelta('15 minutes'))`

Error message:

>ValueError: Invalid frequency: 3e+11N

rdtools/normalization.py
```diff
@@ -383,40 +375,270 @@ def check_series_frequency(series, series_description):
...
+    if not isinstance(time_series.index, pd.DatetimeIndex):
         raise ValueError('power_series must be a pandas series with a DatetimeIndex')
 
-    times = np.array(power_series.index.astype('int64'))
-    time_deltas = np.diff(times) / 10.0**9 / 3600.0
+    t_steps = t_step_nanoseconds(time_series)
+    median_step_ns = t_steps.median()
+
+    if target_frequency is None:
+        target_frequency = str(median_step_ns) + 'N'  # Pandas offset allias in ns
```
cdeline: replace with:
`target_frequency = str(int(median_step_ns)) + 'N'  # Pandas offset allias in ns`
to fix python 2.7 string casting bug.

rdtools/normalization.py
mdeceglie: `energy_from_power()` can be called with a single argument, resolving this suggestion.

rdtools/normalization.py
mdeceglie: Done. Thanks!

mdeceglie:
Thanks for the input on this everyone. In addition to the example notebook, @dirkjordan and I have been testing the changes on system 4 from [PVDAQ](https://maps.nrel.gov/pvdaq/). The new code changes answers slightly, especially in the clear-sky workflow, but confidence intervals still overlap substantially and we are generally comfortable with the updated calculations.

I've also added some more instruction about avoiding the use of interpolation for downsampling to the example notebook.

@dirkjordan and @cdeline, could you please indicate your approval at this stage?

@abshinn, please take a look at the changes to `normalization_pvwatts_test.py` and `normalization_sapm_test.py` and ensure you agree. (Comments on the rest of the PR are also welcome of course!).

cdeline:
Yep, this looks good.  I'm fine with merging this to development and moving on to the next thing.

dirkjordan:
I'm fine with a merge too.

mdeceglie:
Thanks @cdeline and @dirkjordan. I'll give @abshinn a chance to take a look at the modified tests, then hopefully proceed with the merge.

PR #111, opened 2019-03-08
Update to use pvlib>=0.6.0

mdeceglie:
Previously, RdTools required pvlib <0.6. pvlib 0.6.0 represents a major update and includes API changes. `rdtools.normalization` makes use of pvlib functions, but is unaffected by the pvlib update as far as I can tell.

The example notebook however uses `get_clearsky()`, and the default behavior for this function has changed with the addition of the `perez_enhancement` keyword. (More details [here](https://github.com/pvlib/pvlib-python/issues/435).) This change in default behavior changes the example notebook results for the clearsky workflow. We can recover the original results by using:  `clearsky = loc.get_clearsky(df.index, solar_position=sun, perez_enhancement=True)` in the 0th step of the clear sky workflow. 

Here is the original output:
original
And here is the new output without the `perez_enhancement` keyword:
updated

We should decide whether to preserve the original output, or adopt the new pvlib default behavior. My vote is for the latter. Since the perez enhancement can be problematic, I think we should guide users toward the more conservative approach.


mdeceglie:
I ran the notebook with the upstream changes from PR #105. The clear sky output  is now:

Screen Shot 2019-03-11 at 9 14 06 AM

mdeceglie:
Let's go forward with the new default behavior of `get_clearsky()`, because it is safer. I've added a note so people can understand the change.

PR #116, opened 2019-06-24
Analysis Pipeline

araval:
This simplifies the example notebook with the help "analysis_pipeline.py". We would like to calculate degradation for a large number of systems in our database, so having a pipeline is more efficient. With the help of this script, we now input system metadata and energy or power time-series and we get degradation values along with confidence intervals as output. 

See for example `docs/pipeline example.ipynb`.

### DIFF
--- a/analysis_pipeline.py
+++ b/analysis_pipeline.py
@@ -1,0 +1,270 @@
+from datetime import timedelta
+import pandas as pd
+import numpy as np
+import pvlib
+import rdtools
+import logging
+from timezonefinder import TimezoneFinder
+
+import logger
+logger.init()
+    
+def get_timezone(latitude, longitude):
+    """
+    returns: timezone (str)
+    """
+    tf = TimezoneFinder()
+    tz = tf.timezone_at(lat=latitude, lng=longitude)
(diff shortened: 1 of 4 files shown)


mdeceglie:
@araval Thanks for the PR! We're also working on some code to streamline the workflow. I'll push that and issue a PR by the end of the week. Then let's all plan to compare notes and figure out how to best put all the pieces together in a consistent way.

analysis_pipeline.py
```diff
@@ -0,0 +1,279 @@
...
+        '''
+        input_filename: csv with power time-series
+        system_metdata: dictionary with the following keys: 
+                        - systemid   (string. unique identifier) 
+                        - latitude
+                        - longitude
+                        - tilt       (degrees)
+                        - azimuth    (degrees)
+                        - pdc        (DC capacity in Watts)
+                        - tempco     (temperature coefficient)
```
cwhanse: I would suggest `gamma_pmp` as the keyword

araval: Thank you, done!

cdeline:
This content is being handled by #117 now.

Issue #36, opened 2017-08-22
Single-line analysis routines


mdeceglie:
As a longer-term goal, we would like to develop analysis routines that bundle normalization, filtering, aggregation, and analysis so they can be deployed with a single line.

mdeceglie:
We've now considered several different ways to accomplish this (#116 #117 #130 ). Let's use this issue to facilitate discussion of the best way forward. Of particular interest is the question of whether to require that the single line analysis routine API can be programmed by a separate "recipe file". This would have a few benefits. First, it would allow us to decouple changes in analysis best practices from RdTools major version releases. Second, it would provide analysts and researchers a way to compactly share their routines built with RdTools. If we took this approach I'm imagining the part that gets incorporated and version controlled with RdTools is more of a shell, and the detailed plumbing of the entire routine gets programmed in through the recipe file. 

No matter what, we intend to build this on the functional interface, so people can use that as a library to meet their own needs. 

@cdeline @cwhanse @GregoryKimball @abshinn @kanderso-nrel What do you think about this point?

cwhanse:
+1 to the "recipe file" concept. I'm not confident enough to say that one of #116, #117 or #130 is better. #130 adds a layer that is unfamiliar to me and thus I'm uncertain about being about to use, debug or contribute to that approach.

kanderso-nrel:
[PVMismatch](https://github.com/SunPower/PVMismatch) uses a property-based approach in its [PVCell](https://github.com/SunPower/PVMismatch/blob/master/pvmismatch/pvmismatch_lib/pvcell.py) class that is similar in spirit to the declarative approach in #130.  Here's a toy example in the context of RdTools:

``` python
class SystemAnalysis:
    
    def __init__(self, poa, power):
        self.poa = poa
        self.power = power
    
    @property
    def normalized_power(self):
        return rdtools.normalize(self.power, self.poa)

    @property
    def filtered_power(self):
        return rdtools.filter(self.normalized_power)

    @property
    def degradation_result(self):
        results = rdtools.degradation(self.filtered_power)
        return results

poa, power = ...
sa = SystemAnalysis(poa, power)
results = sa.degradation_results
```
[...]

PR #117, opened 2019-06-25
Model chains

mdeceglie:
This PR provides an updated object-oriented API implemented with `system_analysis` class that streamlines clear-sky- and sensor-based workflows for both soiling and degradation analysis. It also incorporates a plotting module and built in plotting methods. It is designed to be easy to use out of the box, but fully customizable for advanced analysis.

For an example see: https://github.com/NREL/rdtools/blob/model_chains/docs/System_analysis_example.ipynb 

To do:
- [x] Update `setup.py` and requirements for plotting
- [x] Add tests
- [x] Finalize name of module and class
- [x] Add documentation to index.rst
- [x] Align parameter names with rest of modules
- [x] Consider prepending _ to some methods as weak internal use indicator


cdeline:
Any idea why travis-ci tests are not passing?

docs/notebook_requirements.txt
```diff
@@ -18,7 +18,7 @@ jupyter-console==5.2.0
 jupyter-core==4.4.0
 kiwisolver==1.0.1
 MarkupSafe==1.0
-matplotlib==2.2.2
+matplotlib==3.0.3
```
cdeline: This seems to break python 2.7.  Any reason that this is required?

mdeceglie: I'll change it back to 2.2.2. We should abandon python 2 soon anyway, it is officially deprecated at the beginning of 2020.

rdtools/system_analysis.py
```diff
@@ -0,0 +1,586 @@
...
+from . import aggregation
+from . import degradation
+from . import soiling
+from . import clearsky_temperature
+from . import plotting
+
+
+class system_analysis():
+    '''
+    Class for end-to-end analysis
```
cdeline: Recommend moving __init__ docstring up to the class docstring such that help(system_analysis) and print(system_analysis.__doc__) presents useful information.  This is considered best practice per google:
https://stackoverflow.com/questions/37019744/is-there-a-consensus-what-should-be-documented-in-the-classes-and-init-docst

docs/notebook_requirements.txt
cdeline: Thanks.  Is there a specific advantage of 3.0.3 over 2.2.2, just so we can be aware for future use?  I would be ok with ditching python 2.7 support early if we do a major RdTools revision that requires it, but if it's not too much hassle to keep 2.7 going, then I figure why not...

mdeceglie: No specific advantage that I know of, it's just what I had in the environment I used to develop this :-)

rdtools/system_analysis.py
mdeceglie: Done, thanks!

mdeceglie:
Thanks for the review @cdeline! Before merging I still need to add tests to cover the new functionality and we should align with PR #116

rdtools/system_analysis.py
```diff
@@ -0,0 +1,585 @@
...
+        elif pv_input == 'energy':
+            self.pv_power = None
+            self.pv_energy = pv
+
+        if temperature_input == 'cell':
+            self.cell_temperature = temperature
+            self.ambient_temperature = None
+        elif temperature_input == 'ambient':
+            self.cell_temperature = None
+            self.ambient_temperature = temperature
```
araval: The logic in this block above could be simplified a bit by changing the arguments in the init function to be `cell_temperature` and `ambient_temperature`, instead of `temperature` and `temperature_input`. 
you could then just have two lines: 
self.cell_temperature = cell_temperature
self.ambient_temperature = ambient_temperature

rdtools/system_analysis.py
```diff
@@ -0,0 +1,585 @@
+'''
```
abshinn: Even if many of the underlying components are already covered by unit tests, the basic functionality of the analysis class in this module should also be covered by unit tests. I don't think it's necessary to try for complete coverage right away. Perhaps all that is needed for the near term is a simple check that this module can be imported and the main class can be instantiated.

rdtools/system_analysis.py
kevinsa5: Separate `cell_temperature` and `ambient_temperature` arguments also avoids the issue of misspelled `'cell'` or `'ambient'` strings passing through the `if/elif` blocks and causing mysterious errors later on.

rdtools/system_analysis.py
```diff
@@ -0,0 +1,587 @@
...
+        -------
+        None
+        '''
+        if times is None:
+            times = self.poa.index
+        if self.pvlib_location is None:
+            raise ValueError('pvlib location must be provided')
+        if self.pv_tilt is None or self.pv_azimuth is None:
+            raise ValueError('pv_tilt and pv_azimuth must be provded')
+        if times is not self.poa.index and rescale is True:
```
kevinsa5: Does `times is not self.poa.index` need to use `is`?  One could imagine a situation where `times` and `self.poa.index` are equivalent indexes but different objects and using `is` would reject them.

rdtools/system_analysis.py
```diff
@@ -0,0 +1,587 @@
...
+        bool_filter = True
+
+        if case == 'sensor':
+            poa = self.poa
+            cell_temp = self.cell_temperature
+        if case == 'clearsky':
+            poa = self.clearsky_poa
+            cell_temp = self.clearsky_cell_temperature
+
+        if 'normalized_filter' in self.filter_params.keys():
```
kevinsa5: Here and below, `key in my_dict` is semantically the same as `key in my_dict.keys()`, although the former avoids generating a temporary `dict_keys` object.

rdtools/system_analysis.py
```diff
@@ -0,0 +1,587 @@
...
+
+    def aggregate(self, normalized, insolation):
+        '''
+        Returns insolation-weighted normalized PV energy and the associated aggregated insolation
+
+        Parameters
+        ----------
+        normalized: Pandas Times Series (numeric), normalized pv energy time series
+        insolation: Pandas Time Series (numeric), Time series of insolation associated
+                    with each normalize_energy point
```
kevinsa5: should `normalize_energy` be `normalized`?

rdtools/system_analysis.py
```diff
@@ -0,0 +1,587 @@
...
+        Perform year-on-year degradation analysis on insolation-weighted aggregated energy yield.
+
+        Parameters
+        ----------
+        aggregated: Pandas Time Series of insolation-weighted aggregated normalized PV energy
+        kwargs: passed to degradation.degradation_year_on_year()
+
+        Returns
+        -------
+        dict or year-on-year results with keys:
```
kevinsa5: typo in `dict or year-on-year results with keys:`?

rdtools/system_analysis.py
```diff
@@ -0,0 +1,587 @@
...
+            'p50_rd': yoy_rd,
+            'rd_confidence_interval': yoy_ci,
+            'calc_info': yoy_info
+        }
+
+        return yoy_results
+
+    def srr_soiling(self, aggregated, aggregated_insolation, **kwargs):
+        '''
+        Performs stochastic rate and recovery soiling analysis.
```
kevinsa5: Inconsistent docstring style ("Performs" vs "Perform" as in `yoy_degradation`).  [PEP 257](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings) prefers the "command" style.  `aggregate` and the `plot_xyz` functions also use the "description" style.

rdtools/system_analysis.py
```diff
@@ -0,0 +1,587 @@
...
+
+        Returns
+        -------
+        dict or year-on-year results with keys:
+            'p50_sratio': The median year-on-year insolation-weighted soiling ratio (numeric)
+            'sratio_confidence_interval': lower and upper bounds of insolation-weighted soiling ratio
+                                          confidence interval(list)
+            'calc_info': Dict of detailed results (see soiling.soiling_srr() docs)
+        '''
+        if aggregated.index.freq != 'D' or aggregated_insolation.index.freq != 'D':
```
kevinsa5: Technically other frequency strings may be valid here as well -- `'24h'`, `'1440T'` etc.  Could instead check the output of `pd.infer_freq(aggregated.index)` to be safe.

rdtools/system_analysis.py
```diff
@@ -0,0 +1,587 @@
...
+
+    Instantiation parameters
+    ----------
+    pv: Right-labeled Pandas Time Series of PV energy or power. If energy, should *not* be cumulative, but only for preceding time step.
+    poa: Right-labeled Pandas Time Series of measured plane of array irradiance in W/m^2
+    temperature: Right-label Pandas Time Series of either cell or ambient temperature in Celsius
+    temperature_coefficient: The fractional PV power temperature coefficient (numeric)
+    aggregation_freq: The frequency with which to aggregate normalized PV data for analysis (Pandas frequency specification)
+    pv_input: 'power' or 'energy' to specify type of input used for pv parameter
+    temeprature_input: 'cell' or 'ambient' to specify type of input used for temperature parameter
```
mdeceglie: We should consider changing the options for temperature input to `module` and `ambient`. In practice it is a good approximation to use back of module temperature for temperature corrections. If we do make this change, we should also change the behavior and name of `system_analysis.calc_cell_temperature()` to return module temperature.

dirkjordan: I agree, both option would be permissible. Module temperature is more accurate but for degradation analysis consistency is important. In practice, I have seen a lot of back-of-the-module sensors peel off over many years in the field. This can bias the results. 
On a different note, the description of the temperature coefficient could be a little bit more specific. For example, if the temp coefficient is 0.40 %/year is the numeric input to be used 0.4 or 0.004, probably 0.004 but to a beginner this may not be clear.

rdtools/system_analysis.py
```diff
@@ -0,0 +1,587 @@
...
+    temperature_model: model parameter pvsystem.sapm_celltemp() used in calculating cell temperature from ambient
+    pv_azimuth: Azimuth of PV array in degrees from north (numeric)
+    pv_tilt: Tilt of PV array in degrees from horizontal
+    pv_nameplate: Nameplate DC rating of PV array in Watts (numeric)
+    interp_freq: The frequency to which to interpolate all Pandas Time Series passed at instantiation. We recommend using the natural
+                 frequency of the data, rather than up or down sampling. Analysis requires regular time series.
+    max_timedelta: The maximum gap in the data to be interpolated/integrated across when interpolating or calculating energy from power (Timedelta)
+    '''
+
+    def __init__(self, pv, poa=None, temperature=None, temperature_coefficient=None,
```
cdeline: The default temperature_coefficient input  = None results in an error during both sensor_analysis and clearsky_analysis.  I suggest that we replace this with a meaningful default value = -0.004

cdeline: .. or even better, set it and return a user warning.

rdtools/system_analysis.py
```diff
@@ -0,0 +1,585 @@
...
+from . import normalization
+from . import filtering
+from . import aggregation
+from . import degradation
+from . import soiling
+from . import clearsky_temperature
+from . import plotting
+
+
+class system_analysis():
```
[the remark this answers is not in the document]
kevinsa5: See also https://www.python.org/dev/peps/pep-0008/#class-names

rdtools/system_analysis.py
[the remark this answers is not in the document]
mdeceglie: Fixed with the updated temperature interface

rdtools/system_analysis.py
```diff
@@ -0,0 +1,585 @@
...
+    max_timedelta: The maximum gap in the data to be interpolated/integrated across when interpolating or calculating energy from power (Timedelta)
+    '''
+    
+    def __init__(self, pv, poa=None, temperature=None, temperature_coefficient=None,
+                 aggregation_freq='D', pv_input='power', temperature_input='cell', pvlib_location=None,
+                 clearsky_poa=None, clearsky_temperature=None, clearsky_temperature_input='cell', windspeed=0, albedo=0.25,
+                 temperature_model=None, pv_azimuth=None, pv_tilt=None, pv_nameplate=None, interp_freq=None, max_timedelta=None):
+        '''
+        Instantiates a system_analysis object
+        '''
```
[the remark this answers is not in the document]
mdeceglie: Done, thanks

rdtools/system_analysis.py
```diff
@@ -0,0 +1,585 @@
...
+               default value of 0 neglects the wind in this calculation
+    albedo: Albedo to be used in irradiance transposition calculations (numeric)
+    temperature_model: model parameter pvsystem.sapm_celltemp() used in calculating cell temperature from ambient
+    pv_azimuth: Azimuth of PV array in degrees from north (numeric)
+    pv_tilt: Tilt of PV array in degrees from horizontal
+    pv_nameplate: Nameplate DC rating of PV array in Watts (numeric)
+    interp_freq: The frequency to which to interpolate all Pandas Time Series passed at instantiation. We recommend using the natural
+                 frequency of the data, rather than up or down sampling. Analysis requires regular time series.
+    max_timedelta: The maximum gap in the data to be interpolated/integrated across when interpolating or calculating energy from power (Timedelta)
+    '''
```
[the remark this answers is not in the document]
mdeceglie: This is a good suggestion and extends beyond the additions made here. #122 has been opened.

rdtools/system_analysis.py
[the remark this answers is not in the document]
mdeceglie: Done, thank you

rdtools/system_analysis.py
```diff
@@ -0,0 +1,585 @@
...
+    def calc_clearsky_poa(self, times=None, rescale=True, model='isotropic', **kwargs):
+        '''
+        Calculate clearsky plane-of-array irradiance and store in self.clearsky_poa
+
+        Parameters
+        ----------
+        times: Pandas DateTimeIndex for which to calculate clearsky poa
+        rescale: Whether to attempt to rescale clearsky irradiance to measured (bool)
+        model: Model for pvlib.irradiance.get_total_irradiance() (str)
+        kwargs: key word arguments passed to pvlib.irradiance.get_total_irradiance()
```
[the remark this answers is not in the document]
mdeceglie: I'd lean toward including the entire function name e.g. `get_total_irradiance_kwargs`. But is that too verbose?

There are a number of places we pass kwargs to lower level functions. If we make this change we should change them all.

mdeceglie:
To do: The current names of methods and attributes don't leave much room for other analysis, We should rename things to leave room for things like capacity tests or performance ratio analyses.

rdtools/system_analysis.py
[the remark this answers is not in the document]
kevinsa5: Since the function signature uses the double-splat operator, the user can't directly specify a value for `kwargs`, they can only specify keyword arguments that will then get collected into `kwargs`.  I.e. code like this isn't valid: `calc_clearsky_poa(self, times=..., rescale=..., model='isotropic', kwargs={'key1': 'value1'})`.  So from the user's perspective, I think the only place they would ever see the name of that variable is if they were looking at the docstring, in which case it's pretty clear what it's used for.  I'd say leave the name as-is since `**kwargs` is idiomatic python.  

Also pvlib uses `kwargs` the way this PR currently does, to pass on to the lower-level function: https://pvlib-python.readthedocs.io/en/stable/generated/pvlib.pvsystem.PVSystem.get_irradiance.html#pvlib.pvsystem.PVSystem.get_irradiance

I do have some small preference for pvlib's wording though -- `Extra parameters passed to xyz` vs `key word arguments passed to xyz`.

rdtools/system_analysis.py
```diff
@@ -0,0 +1,585 @@
...
+
+    def calc_clearsky_poa(self, times=None, rescale=True, model='isotropic', **kwargs):
+        '''
+        Calculate clearsky plane-of-array irradiance and store in self.clearsky_poa
+
+        Parameters
+        ----------
+        times: Pandas DateTimeIndex for which to calculate clearsky poa
+        rescale: Whether to attempt to rescale clearsky irradiance to measured (bool)
+        model: Model for pvlib.irradiance.get_total_irradiance() (str)
```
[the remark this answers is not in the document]
kevinsa5: Python would throw a `TypeError: calc_clearsky_poa() got multiple values for argument 'model'` if you specified it twice, so no possibility of an issue there.  Though I think it is redundant to include the model in the signature, since `get_total_irradiance` already defaults to `isotropic` anyway.  Removing it from this signature would still let it get passed in through `kwargs` without any change in the user code.

rdtools/system_analysis.py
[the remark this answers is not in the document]
mdeceglie: Thanks @kevinsa5, I think that makes sense.

rdtools/system_analysis.py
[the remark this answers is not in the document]
mdeceglie: I think my past self was wanting to encode our preferred default behavior, but it does seem confusing, so I can remove it.

rdtools/system_analysis.py
```diff
@@ -0,0 +1,585 @@
...
+            pv_nameplate = 1.0
+        else:
+            renorm = False
+
+        if self.temperature_coefficient is None:
+            raise ValueError('Temperature coeffcient must be available to perform pvwatts_norm')
+
+        pvwatts_kws = {"poa_global": poa,
+                       "P_ref": pv_nameplate,
+                       "T_cell": cell_temperature,
```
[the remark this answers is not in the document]
mdeceglie: Good question, the reason is that sometimes we use the clearsky versions of those parameters and sometimes the sensor versions. The preprocess methods handle passing in the correct parameters. 

The alternative is to add a parameter for which parameter to pvwatts_norm() with some conditionals, but I'm not sure that ends up being cleaner. Please chime in if you think it's worth going that way.

rdtools/system_analysis.py
```diff
@@ -0,0 +1,585 @@
...
+        aggregated, aggregated_insolation = self.aggregate(normalized[self.sensor_filter], insolation[self.sensor_filter])
+        self.sensor_aggregated_performance = aggregated
+        self.sensor_aggregated_insolation = aggregated_insolation
+
+    def clearsky_preprocess(self):
+        '''
+        Perform clear-sky-based normalization, filtering, and aggregation work flow
+        '''
+        if self.clearsky_poa is None:
+            self.calc_clearsky_poa()  # kwargs?
```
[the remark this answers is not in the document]
mdeceglie: Fixed, thanks

setup.py
```diff
@@ -34,12 +34,13 @@
...
 INSTALL_REQUIRES = [
+    'h5py >= 2.7.1',
+    'matplotlib >= 2.2.2',
     'numpy >= 1.12',
     'pandas >= 0.23.0',
+    'pvlib >= 0.6.0, <0.7.0',
     'statsmodels >= 0.8.0',
     'scipy >= 0.19.1',
-    'h5py >= 2.7.1',
-    'pvlib >= 0.6.0, <0.7.0',
```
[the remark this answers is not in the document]
mdeceglie: I can't trace the reasoning for the order change, but, it's working now!

docs/System_analysis_example.ipynb
```diff
@@ -0,0 +1,637 @@
...
...gB2vq4Yh0IgChIgtkFVqaqaKPKVL0VZMXOWS5t7nD1/kQtXRnzo035dsbyjd3SNVXxKTs94o8cyxM4HWZxCPoUkBWJY68KlbTAprG9N6A57YJWkmDKqoKM1syZlNC7o9VOaumR1EYoKOkOY1bAQgS0hvA/IobPUo2xqFrGMC8daVRFHbfngcXDUCpmfBf468Id4r76l5UjsT5sPsh+Jvlb9Mq9wqUvyWgmKKVEgjCYF29vbnL8y4o8/MWJ774iJtSfMg/hgS4Wf5nYiv6b4OQ/CzsSvJ+Y51BUQwCCFroOFhQ5NWJCI0uv3CWzJrjMwzchrvx4ptiJMI6ppgSQBzvm2sVnlKyMJIE2gGINbgCuXM5IFuGd5gVkNVVVBrxWeOA6O+hPzN/DpPOdPcjAtrx32y/waB6Hxa4dXa6cFMIamaXAYmEekAyMYV3J5d0aeTTlz/iKjvOSzn4Y/aO5swKWPD7IkwOcvgG1gWkCawv2r0BuACpgIcNBJvEHrxFAXQAeyoqAXG0IxjCcZC72IqijB+DSeYDFCgW7omATQFJZeB9wMZjMoK0gjqBo4tQBFBpMclmMwarlvKSXtdNuAzDHxUvpW753kQFpeHdyob8vB51ej0iIE4jDzCCp4T9Kp93CCMEK08Z5Rlvtk7rJgPNrhwvo6v/fhhi189v+dMoyr+FpW8LmFNZDVsNoHIl+kkzsfVXbOG/44gjDykeQggLTToS5yOp0u01lGgbI5UqzWuBoKA0kX7KzGrC3TdzllXWBUqXNwBlwNdQA6haALYQBVBZ2B90rrYEggfo22Fbs9Hm6m5/imA5v/FPhVEflHHCpRVdXnTmhsLXcZ12t9+iKpsflzQef/vabEHRivz1hVFY0aXFUSRRF7oxGTvGayt83OtOLJM+t88nE/ha7u0L0+hK9fXgBWV6AqoFCghMWuV8exDXRS2NoBlqErsLYAKkJVKZ1UCKMUcRUuSMmnGWl3SKIZWQIdo6QLMYFamFiIDXU1Ja8tea7sTH1Qx9QwXIC9s2ANNCPYrGD1AZ9Ufs+9QwIq1EQ01hGGred4HNzMc3zmOvv+6qHttm/164jDSd0v2p6n4uz3aznY/MqL0HoDGoQRxtbklTKd7nJld8bFy+uc39zhmWfhsTE8fQfuL8R/mL8AWIuhG0O6CJHzCdqLFSwOIEngyh7MKljpwGDBT227Q
```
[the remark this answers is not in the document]
mdeceglie: fixed, thanks

rdtools/system_analysis.py
```diff
@@ -0,0 +1,585 @@
...
+    def filter(self, normalized, case):
+        '''
+        Calculate filters based on those in rdtools.filtering. Uses self.filter_params, which is a dict,
+        the keys of which are names of functions in rdtools.filtering, and the values of which are dicts
+        containing the associated parameters with which to run the filtering functions. See examples for
+        details on how to modify filter parameters.
+
+        Parameters
+        ----------
+        normalized: normalized PV energy (Pandas Time Series)
```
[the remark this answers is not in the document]
mdeceglie: Since we used `normalized` elsewhere, my vote is to keep consistency throughout the codebase.

rdtools/system_analysis.py
```diff
@@ -0,0 +1,585 @@
...
+
+        srr_results = {
+            'p50_sratio': sr,
+            'sratio_confidence_interval': sr_ci,
+            'calc_info': soiling_info
+        }
+
+        return srr_results
+
+    def sensor_preprocess(self):
```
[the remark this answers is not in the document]
mdeceglie: I've struggled with that question also. I think that more experienced analysts may want to use that method (as well as other lower level methods). For example you can preprocess, and then perhaps adjust filtering parameters and preprocess again. In the name of flexibility I decided to err on the side of naming all the methods without the underscore.

rdtools/system_analysis.py
```diff
@@ -0,0 +1,583 @@
...
+
+        return cell_temp
+
+    def calc_clearsky_tamb(self):
+        '''
+        Calculate clear-sky ambient temperature and store in self.clearsky_ambient_temperature
+        '''
+        times = self.clearsky_poa.index
+        if self.pvlib_location is None:
+            raise ValueError('pvlib location must be provided')
```
kanderso-nrel: This error message could be misinterpreted as telling the user to pass a pvlib location to this function instead of the original constructor of the SystemAnalysis object.  Maybe something like `'Missing location info, must pass pvlib_location to SystemAnalysis()'` and/or make a note in the docstring about the requirement?

rdtools/system_analysis.py
```diff
@@ -0,0 +1,583 @@
...
+        (Normalized pv energy, associated insolation)
+        '''
+        if self.pv_nameplate is None:
+            renorm = True
+            pv_nameplate = 1.0
+        else:
+            renorm = False
+
+        if self.temperature_coefficient is None:
+            raise ValueError('Temperature coeffcient must be available to perform pvwatts_norm')
```
kanderso-nrel: Maybe the same should be done here as with `calc_clearsky_tamb` and `pvlib_location` above.

rdtools/system_analysis.py
```diff
@@ -0,0 +1,583 @@
...
+                       "T_ref": 25,
+                       "gamma_pdc": self.temperature_coefficient}
+
+        normalized, insolation = normalization.normalize_with_pvwatts(self.pv_energy, pvwatts_kws)
+
+        if renorm:
+            # Normalize to the 95th percentile for convenience, this is renormalized out
+            # in the calculations but is relevant to normalized_filter()
+            x = normalized[np.isfinite(normalized)]
+            normalized = normalized / x.quantile(0.95)
```
kanderso-nrel: Should this 95th percentile normalization be mentioned in the docstring?

rdtools/system_analysis.py
```diff
@@ -0,0 +1,583 @@
...
+            bool_filter = bool_filter & f
+        if 'tcell_filter' in self.filter_params.keys():
+            if cell_temp is None:
+                raise ValueError('Cell temperature must be available to use tcell_filter')
+            f = filtering.tcell_filter(cell_temp, **self.filter_params['tcell_filter'])
+            bool_filter = bool_filter & f
+        if 'clip_filter' in self.filter_params.keys():
+            if self.pv_power is None:
+                raise ValueError('PV power (not energy) is required for the clipping filter. Either omit the clipping filter,'
+                                 'provide PV power at instantiation, or explicitly assign system_analysis.pv_power.')
```
kanderso-nrel: Great example of the type of error message I was trying to come up with above

rdtools/system_analysis.py
```diff
@@ -0,0 +1,583 @@
...
+        if case == 'clearsky':
+            if self.poa is None or self.clearsky_poa is None:
+                raise ValueError('Both poa and clearsky_poa must be available to do clearsky filtering with csi_filter')
+            f = filtering.csi_filter(self.poa, self.clearsky_poa, **self.filter_params['csi_filter'])
+            bool_filter = bool_filter & f
+
+        if case == 'sensor':
+            self.sensor_filter = bool_filter
+        elif case == 'clearsky':
+            self.clearsky_filter = bool_filter
```
kanderso-nrel: Would it be useful to keep track of the individual filter results separately and store them somewhere to allow inspection by the user?  I could imagine someone wanting to know which filter is removing certain data, or wanting to roll up things like "the tcell filter removed 50% of the dataset".

rdtools/system_analysis.py
kanderso-nrel: Just found another option: `aggregated.index.freq != pd.tseries.offsets.Day()`.  Not sure how far back in pandas history it would work though.

rdtools/system_analysis.py
```diff
@@ -0,0 +1,585 @@
...
+                self.calc_clearsky_tamb()
+            self.clearsky_cell_temperature = self.calc_cell_temperature(self.clearsky_poa, 0, self.clearsky_ambient_temperature)
+            # Note example notebook uses windspeed=0 in the clearskybranch
+        cs_normalized, cs_insolation = self.pvwatts_norm(self.clearsky_poa, self.clearsky_cell_temperature)
+        self.filter(cs_normalized, 'clearsky')
+        cs_aggregated, cs_aggregated_insolation = self.aggregate(cs_normalized[self.clearsky_filter], cs_insolation[self.clearsky_filter])
+        self.clearsky_aggregated_performance = cs_aggregated
+        self.clearsky_aggregated_insolation = cs_aggregated_insolation
+
+    def sensor_analysis(self, analyses=['yoy_degradation'], yoy_kwargs={}, srr_kwargs={}):
```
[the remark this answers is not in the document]
kanderso-nrel: I can imagine the user's perspective being that they are running a degradation analysis that happens to be sensor (or clearsky) based, which to me means that the API should reflect that as well.  I.e. the function name should be degradation or soiling and switching between clearsky and sensor should be a parameter, not the other way around.  Thoughts?

I think I like the idea of a single function with switches to control what gets run, although I think I would favor two kwargs: `sensor_analyses=['yoy_degradation'], clearsky_analyses=['yoy_degradation']` and both have the option to include `'srr_soiling'` as well.  The downside is that it doesn't allow differentiation between the `yoy_kwargs` and `srr_kwargs` passed to the clearsky and sensor analyses.  Is that a big deal?

rdtools/system_analysis.py
```diff
@@ -0,0 +1,583 @@
...
+        matplotlib figure containing two axes
+
+        '''
+
+        if result_to_plot == 'sensor':
+            results_dict = self.results['sensor']['yoy_degradation']
+            aggregated = self.sensor_aggregated_performance
+        elif result_to_plot == 'clearsky':
+            results_dict = self.results['clearsky']['yoy_degradation']
+            aggregated = self.clearsky_aggregated_performance
```
kanderso-nrel: Could add an `else` here to catch typos and throw a more informative error than `NameError: name 'results_dict' is not defined`.  The other plot functions as well

rdtools/system_analysis.py
```diff
@@ -0,0 +1,585 @@
...
+    interp_freq: The frequency to which to interpolate all Pandas Time Series passed at instantiation. We recommend using the natural
+                 frequency of the data, rather than up or down sampling. Analysis requires regular time series.
+    max_timedelta: The maximum gap in the data to be interpolated/integrated across when interpolating or calculating energy from power (Timedelta)
+    '''
+    
+    def __init__(self, pv, poa=None, temperature=None, temperature_coefficient=None,
+                 aggregation_freq='D', pv_input='power', temperature_input='cell', pvlib_location=None,
+                 clearsky_poa=None, clearsky_temperature=None, clearsky_temperature_input='cell', windspeed=0, albedo=0.25,
+                 temperature_model=None, pv_azimuth=None, pv_tilt=None, pv_nameplate=None, interp_freq=None, max_timedelta=None):
+        '''
```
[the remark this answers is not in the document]
mdeceglie: You raise a good point that the there are a lot of inputs. But I agree that it's somewhat necessary, even if we started grouping things as dicts we'd have to document those in the docstring. I'm inclinded to leave it as is. Users can always package their inputs into a dict and use ** to unpack the arguments.

rdtools/system_analysis.py
```diff
@@ -0,0 +1,587 @@
...
+    pvlib_location: pvlib location object used for calculating clearsky temperature and irradiance
+    clearsky_poa: Right-labeled Pandas Time Series of clear-sky plane of array irradiance
+    clearsky_temperature: Right-labeled Pandas Time Series of clear-sky cell or ambient temperature in Celsius
+    clearsky_temperature_input: 'cell' or 'ambient' to specify type of input used for clearsky_temperature parameter
+    windspeed: Right-labeled Pandas Time Series or numeric indicating wind speed in m/s for use in calculating cell temperature from ambient
+               default value of 0 neglects the wind in this calculation
+    albedo: Albedo to be used in irradiance transposition calculations (numeric)
+    temperature_model: model parameter pvsystem.sapm_celltemp() used in calculating cell temperature from ambient
+    pv_azimuth: Azimuth of PV array in degrees from north (numeric)
+    pv_tilt: Tilt of PV array in degrees from horizontal
```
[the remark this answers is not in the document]
mdeceglie: Docstrings have been updated to numpy style

rdtools/system_analysis.py
mdeceglie: yep, fixed

rdtools/system_analysis.py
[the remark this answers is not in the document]
kanderso-nrel: Just to be a stickler... looks like default values still need to be added where relevant

rdtools/system_analysis.py
mdeceglie: Fixed, thanks

rdtools/system_analysis.py
mdeceglie: Fixed, thanks

rdtools/system_analysis.py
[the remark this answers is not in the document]
kanderso-nrel: Also the docstrings in plotting.py (and filtering.py?)

rdtools/system_analysis.py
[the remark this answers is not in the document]
mdeceglie: The new classes and functions now conform to numpy style. Thanks.

rdtools/system_analysis.py
```diff
@@ -0,0 +1,585 @@
...
+        self.results = {}
+
+        # Initialize to use default filter parameters
+        self.filter_params = {
+            'normalized_filter': {},
+            'poa_filter': {},
+            'tcell_filter': {},
+            'clip_filter': {},
+            'csi_filter': {},
+            'ad_hoc_filter': None  # use this to include an explict filter
```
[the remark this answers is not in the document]
mdeceglie: Good suggestion. I've added some more to the revised docstring and also an example in the notebook.

rdtools/system_analysis.py
```diff
@@ -0,0 +1,672 @@
...
+            renorm = True
+            pv_nameplate = 1.0
+        else:
+            renorm = False
+
+        if self.temperature_coefficient is None:
+            raise ValueError('Temperature coeffcient must be available to perform pvwatts_norm')
+
+        pvwatts_kws = {"poa_global": poa,
+                       "P_ref": pv_nameplate,
```
kanderso-nrel: `pv_nameplate` is only defined if `self.pv_nameplate is None`, this throws an UnboundLocalVariable error if `pv_nameplate` is given to the constructor.

rdtools/system_analysis.py
```diff
@@ -0,0 +1,583 @@
...
+            if clearsky_poa is not None:
+                clearsky_poa = normalization.interpolate(clearsky_poa, interp_freq, max_timedelta)
+            if clearsky_cell_temperature is not None:
+                clearsky_cell_temperature = normalization.interpolate(clearsky_cell_temperature, interp_freq, max_timedelta)
+            if clearsky_ambient_temperature is not None:
+                clearsky_ambient_temperature = normalization.interpolate(clearsky_ambient_temperature, interp_freq, max_timedelta)
+            if isinstance(pv_azimuth, (pd.Series, pd.DataFrame)):
+                pv_azimuth = normalization.interpolate(pv_azimuth, interp_freq, max_timedelta)
+            if isinstance(pv_tilt, (pd.Series, pd.DataFrame)):
+                pv_tilt = normalization.interpolate(pv_tilt, interp_freq, max_timedelta)
```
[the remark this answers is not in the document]
mdeceglie: Another thing to consider is that we want to support cases where the weather data and the PV data are measured at different frequency. For example, if we have low frequency PV data (daily, monthly, weekly, etc.) we can still do a degradation analysis based on modeled energy at higher frequency from weather data. In this case, the analyst would leave `interp_freq=None`. But would be responsible for supplying regular time series (but of different frequency) to SystemAnalysis(). Supporting these cases is why everything gets kept as separate series.

abshinn:
I see quite a few commits since I last reviewed. Would it be helpful to have me and others take another look?

mdeceglie:
Thanks @abshinn , I think before getting back into the weeds on this particular PR, we all need to reach consensus on what the backend should look like. In particular, this approach isn't programable with a separate "recipe file". But the approach of #130 and/or the Dask approach referenced in that PR potentially are. Once we all agree at a high level  we can figure out the details. I'm rebooting the discussion in #36 to help us reach consensus.

rdtools/system_analysis.py
kanderso-nrel: Fixed by #133

rdtools/system_analysis.py
```diff
@@ -0,0 +1,673 @@
...
+        Returns
+        -------
+        None
+        '''
+        if times is None:
+            times = self.poa.index
+        if self.pvlib_location is None:
+            raise ValueError('pvlib location must be provided')
+        if self.pv_tilt is None or self.pv_azimuth is None:
+            raise ValueError('pv_tilt and pv_azimuth must be provded')
```
kanderso-nrel, line 174:
```suggestion
            raise ValueError('pv_tilt and pv_azimuth must be provided')
```

cdeline:
2.0-beta.0 merged into model_chains

cdeline:
Summarizing here our Aug 24 2020 meeting notes: 

- New functionality going into RdTools will include system availability and system performance index.  
- The existing system_analysis object is going to be limited to soiling and degradation. It operates on ~daily aggregated normalized inverter data.  Other functions work on different types of input, specifically multiple inverter and meter feeds.
- There already is a high-level organizing class for availability, and it doesn't need to go into system_analysis.
- system_analysis object should be renamed to something closer to "degradation_soiling_analysis" since it's not a "system analysis" per se.
- For future use of the SA object with satellite-based irradiance, we just recommend using a separate SA instance, rather than extending SA functionality to have sensor_analysis, clearsky_analysis AND nsrdb_analysis, etc.  This removes the need to have SA object tag or keep track of multiple power and irradiance channels.

rdtools/analysis.py
```diff
@@ -282,8 +282,10 @@ def pvwatts_norm(self, poa, cell_temperature):
                        "temperature_cell_ref": 25,
                        "gamma_pdc": self.temperature_coefficient}
 
+        # suppress RdTools 2.0 deprecation warning
+        warnings.filterwarnings("ignore")
         energy_normalized, insolation = normalization.normalize_with_pvwatts(self.pv_energy, pvwatts_kws)
-
+        warnings.resetwarnings()
```
kanderso-nrel, line 288: We should consider using non-deprecated functionality here instead of suppressing the warning -- we'll have to do it anyway in 3.0

rdtools/analysis.py
```diff
@@ -434,7 +436,12 @@ def srr_soiling(self, aggregated, aggregated_insolation, **kwargs):
                                           confidence interval
             'calc_info' : Dict of detailed results (see soiling.soiling_srr() docs)
         '''
+        # suppress RdTools experimental warning
+        import warnings
+        warnings.filterwarnings("ignore")
         from rdtools import soiling
+        warnings.resetwarnings()
```
kanderso-nrel, line 443: I think libraries should avoid `warnings.resetwarnings()` because it would reset any filters that were defined outside of rdtools in addition to the one set here.  Could we use `warnings.catch_warnings` instead?  It restores the previous list of filters when the context manager exits so any user-defined filters would still be in place after this function runs. 

```python
with warnings.catch_warnings():
    warnings.simplefilter('ignore')
    from rdtools import soiling
```

kanderso-nrel (commented):
Thanks for moving this forward @cdeline, a couple comments below

kanderso-nrel:
I have a case where the irradiance data is unusable and the filtering step removes the entire dataset.  Then the empty normalized signal becomes an empty daily aggregated signal and handed off to `degradation_year_on_year`, which raises a not-very-helpful error `ValueError: min() arg is an empty sequence` from here:  https://github.com/NREL/rdtools/blob/model_chains/rdtools/degradation.py#L225-L226

I'd like a more informative error message to be raised.  I see a few places it could happen:
1) With a minor code reordering in `degradation_year_on_year`, it could be part of the "less than two years of data" check.  I guess it's an improvement because it's an RdTools error instead of an opaque pandas error, but it's still less useful than I'd like; the user is left thinking "I *gave* it two years of data, what's it talking about?" because the error doesn't mention that the filtering step got rid of it all.  I think it's important to be clear whether the error is because they didn't pass in enough data or because too much got removed in the filtering. 
2) At the end of `RdAnalysis.filter`, raise an error if the filter will remove too much data.  Not sure what constitutes "too much" in this context.  Maybe just check if the 2 year threshold is satisfied after applying the filter.  However this seems a little out of scope for `filter` -- it just calculates the filter, which is distinct from deciding if the filter will have downstream consequences. 
3) In the `X_preprocess` methods.  I think this is what I would support -- it makes conceptual sense in that those methods are orchestrating the workflows, so checking validity along the way seems in scope.  One downside is duplicate code because both the sensor and clearsky methods would need to do the check. 

Any preference or other ideas?  I'm imagining something like this in the two preprocess methods:

```python
post_filter = energy_normalized[self.sensor_filter]
if post_filter.empty or post_filter.index[-1] - post_filter.index[0] < pd.Timedelta('730d'):
    raise ValueError("Less than two years of data left after filtering")
```

rdtools/system_analysis.py
mdeceglie: Since the functional API still used cell temperature, I suggest we put a pin in this (open an issue) and proceed with cell temperature for now.

rdtools/system_analysis.py
mdeceglie: Leaving `temperature_coefficient=None` raises a `UserWarning` no temperature correction will be performed. This follows the behavior of `normalize_with_pvwatts` which can be run with no temperature correction, so the current behavior makes sense to me.

rdtools/system_analysis.py
mdeceglie: Changed, thanks!

rdtools/plotting.py
```diff
@@ -0,0 +1,180 @@
...
+
+    fig, (ax1, ax2) = plt.subplots(1, 2, figsize=(10, 3))
+    ax2.hist(yoy_values, label='YOY', bins=len(
+        yoy_values) // 40, color=plot_color)
+    ax2.axvline(x=yoy_rd, color='black', linestyle='dashed', linewidth=3)
+
+    ax2.set_xlim(hist_xmin, hist_xmax)
+
+    ax2.annotate(u' $R_{d}$ = %.2f%%/yr \n confidence interval: \n %.2f to %.2f %%/yr'
+                 % (yoy_rd, yoy_ci[0], yoy_ci[1]), xy=(0.5, 0.7), xycoords='axes fraction',
```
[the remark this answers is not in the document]
mdeceglie: agreed, opened #235 though because that feels out of scope here.

rdtools/system_analysis.py
```diff
@@ -0,0 +1,583 @@
...
+        -------
+        calculated cell temperature (numeric)
+        '''
+        if self.temperature_model is None:
+            cell_temp = pvlib.pvsystem.sapm_celltemp(poa, windspeed, ambient_temperature)
+        else:
+            cell_temp = pvlib.pvsystem.sapm_celltemp(poa, windspeed, ambient_temperature, model=self.temperature_model)
+        cell_temp = cell_temp['temp_cell']
+
+        return cell_temp
```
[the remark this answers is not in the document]
mdeceglie: That's a good point. I wonder if we should make more methods private. As a starting point I suggest adding `_` as a weak internal user indicator to:
`aggregate`
`filter`
`pvwatts_norm`
`calc_clearsky_tamb`
`calc_cell_temperatre`
`calc_clearsky_poa`

This leaves 2-layers of methods for suggested external use.

rdtools/system_analysis.py
mdeceglie: #233 addressed this

rdtools/system_analysis.py
mdeceglie: I think this is a good idea and can be done in a purely additive way. To do it right with testing and examples would take a bit of work and I don't want to further delay this PR, so I opened #236

rdtools/system_analysis.py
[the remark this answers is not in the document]
cdeline: Agreed.  Make it so!

mdeceglie:
I noticed that both the pvdaq4 examples now use 15 minutes `max_timedelta` in the interpolate step. I'd suggest aligning them both to avoid that argument and use the default behavior. My thinking is that our best practice is encoded in the default behavior, including that key word in the example may cause some users to copy paste and run with it without pausing to understand what it does and make a more appropriate choice. If I take that parameter out of both notebooks, I can get alignment. What do you think @cdeline @kanderso-nrel ?

For posterity there are a few of other changes to the functional API PVDAQ 4 notebook proposed here. I think they are all ok, but want to document them:

1. Soiling signal is applied early on and present for both degradation and soiling analysis (this makes it easier to compare with the object oriented example)
2. Soiling results change slightly, I proved to myself that this is due to a combination of the max_timedelta change noted above and 1.
3. Some inconsequential changes to the names of columns in the working dataframe.

rdtools/analysis.py
mdeceglie: Thinking this through in a little more detail perhaps we want to continue to emit warnings when experimental submodules are used. Someone accessing those through this class should still be alerted their answers and API might change.

mdeceglie:
@cdeline  what do you think about removing the object-oriented example notebook `RdAnalysis_example.ipynb` built on the DKSC data? It parallels the PVDAQ notebook and I'm not clear it brings enough extra value to justify maintaining both in parallel. But what do you think?

rdtools/analysis.py
mdeceglie: Screen Shot 2020-11-17 at 12 46 47 PM
Confirmed that removing the code to catch warnings, that the warnings only get emitted when the experimental functionality is accessed.

rdtools/analysis_chains.py
```diff
@@ -0,0 +1,829 @@
...
+            raise ValueError(
+                "Less than two years of data left after filtering")
+
+    def aggregate(self, energy_normalized, insolation):
+        '''
+        Return insolation-weighted normalized PV energy and the associated aggregated insolation
+
+        Parameters
+        ----------
+        normalized : pandas.Series
```
kanderso-nrel, line 444:
```suggestion
        energy_normalized : pandas.Series
```

rdtools/analysis_chains.py
```diff
@@ -0,0 +1,829 @@
...
+    def srr_soiling(self, energy_normalized_daily, insolation_daily, **kwargs):
+        '''
+        Perform stochastic rate and recovery soiling analysis.
+
+        Parameters
+        ---------
+        energy_normalized_daily : pandas.Series
+            Time Series of insolation-weighted aggregated normalized PV energy
+        insolation_daily : pandas.Series
+            Time Series of insolation, aggregated at same level as `aggregated`
```
kanderso-nrel, line 506: `same level as aggregated` -- change aggregated to whatever the updated parameter name lands on

rdtools/analysis_chains.py
```diff
@@ -0,0 +1,829 @@
...
+        -------
+        dict
+            Soiling results with keys:
+            'p50_sratio' : The median insolation-weighted soiling ratio
+            'sratio_confidence_interval' : list of lower and upper bounds of
+                                          insolation-weighted soiling ratio
+                                          confidence interval
+            'calc_info' : Dict of detailed results (see soiling.soiling_srr() docs)
+        '''
+        # suppress RdTools experimental warning
```
kanderso-nrel, line 520: not needed anymore

rdtools/analysis_chains.py
```diff
@@ -0,0 +1,829 @@
...
+        '''
+        Calculate clearsky plane-of-array irradiance and stores in self.poa_global_clearsky
+
+        Parameters
+        ----------
+        times : pandas.DateTimeIndex
+            times on for which to calculate clearsky poa
+        rescale : bool
+            Whether to attempt to rescale clearsky irradiance to measured
+        model : str
```
kanderso-nrel, line 201: Don't think this actually exists.  The model can be passed through via the `**kwargs`, so I think it's ok to just delete this docstring parameter entry.

rdtools/analysis_chains.py
```diff
@@ -457,14 +460,14 @@ def aggregate(self, energy_normalized, insolation):
 
         return aggregated, aggregated_insolation
 
-    def yoy_degradation(self, aggregated, **kwargs):
+    def yoy_degradation(self, energy_normalized, **kwargs):
```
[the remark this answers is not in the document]
kanderso-nrel: Although technically it doesn't have to be daily, it's down to the `aggregation_freq` value...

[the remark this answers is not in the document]
mdeceglie: My thinking was to match the parameter names in `rdtools.degradation`

rdtools/analysis_chains.py
```diff
@@ -491,15 +494,15 @@ def yoy_degradation(self, aggregated, **kwargs):
 
         return yoy_results
 
-    def srr_soiling(self, aggregated, aggregated_insolation, **kwargs):
+    def srr_soiling(self, energy_normalized_daily, insolation_daily, **kwargs):
```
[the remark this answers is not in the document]
mdeceglie: Here again trying to match the parameter names in `rdtools.soiling`

rdtools/analysis_chains.py
```diff
@@ -73,22 +76,22 @@ class TrendAnalysis():
...
-                    ambient_temperature, interp_freq, max_timedelta)
+            if poa_global is not None:
+                poa_global = normalization.interpolate(
+                    poa_global, interp_freq, max_timedelta)
+            if temperature_cell is not None:
+                temperature_cell = normalization.interpolate(
+                    temperature_cell, interp_freq, max_timedelta)
+            if temperature_ambient is not None:
+                temperature_ambient = normalization.interpolate(
+                    temperature_ambient, interp_freq, max_timedelta)
```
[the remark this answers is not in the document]
mdeceglie: Good catch! Note this changes the notebook, but brings it into perfect alignment with the functional notebook

kanderso-nrel (approved):
LGTM 🎉

cdeline:
woot!

Issue #122, opened 2019-10-30
clean up documentation and set up readthedocs


kevinsa5:
Sphinx is the main way to use RTD, but the default sphinx (or rather restructuredtext) docstring format is not necessarily very pretty.  The Google and NumPy docstring formats are supported through the napoleon sphinx extension (pvlib uses this): https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html

Should this be done prior to (or as a part of) cutting the next big release?

mikofski:
My vote is for numpy style docs

```
Parameters
----------
var1 : type
    description
var2 : type
    description
```

this is [the doc style used by pvlib](https://pvlib-python.readthedocs.io/en/latest/contributing.html#documentation)

kanderso-nrel:
The real question is what Sphinx theme to use... http://www.sphinx-doc.org/en/master/usage/theming.html#builtin-themes

The RTD theme isn't built in but seems like the obvious choice.

mdeceglie:
I vote for RTD!

kanderso-nrel:
Closed by #125

mdeceglie:
Let's keep this open until it ends up in `Master`. I'll tag it with "fixed in development branch" so we can keep tabs of changes for the new release.

kanderso-nrel:
Oops, thanks.  Will keep that in mind.

PR #125, opened 2019-11-07
Docstring standardization and pep8 formatting

kanderso-nrel:
Referencing #122.  This PR has gotten a lot bigger than I anticipated.  High-level overview: 

1) Reformatted all docstrings to be compatible with the numpy docstring format for use with the [Napoleon Sphinx extension](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html) in anticipation of setting up RdTools on readthedocs.io.  

2) I also reformatted much of the code to not be over 80 characters long in anticipation of setting up a pep8 linter.  In most cases this just involved breaking code over several lines, but in a few cases it involved "real" code changes like introducing temporary variables.  

3) Added a bunch of sphinx files so that fancy RTD-style docs are autogenerated from the code:
    - The main page is generated from README.md.  Had to make a few formatting changes to allow that translation to work)
    - An example page generated from the example ipython notebook.  Had to make all the titles except for the first one smaller (level 2 instead of level 1).  Will be easy to add additional notebooks later on as other branches get merged.  
    - An API reference that includes submodule summaries, function summaries, and full function documentation.  

To-Do lists
-----------
Docstrings:

- ~clearsky_temperature.get_clearsky_tamb has undocumented parameters~
- ~degradation.* function descriptions lack detail~
- All module descriptions lack detail
- ~filtering.poa_filter is undocumented~
- ~filtering.tcell_filter is undocumented~
- ~normalization.check_series_frequency is poorly documented~
- ~normalization.delta_index is poorly documented~
- ~normalization.ConvergenceError is undocumented~
- ~typo in multiple right-labeled notices~
- ~typo in irradiance_rescale/method~

Sphinx configuration:

- ~rdtools\docs\sphinx\source\modules.rst: WARNING: document isn't included in any toctree~
- ~WARNING: html_static_path entry 'rdtools\docs\sphinx\source\_static' does not exist~
- ~What should the copyright be?  Currently have `© Copyright 2019, National Renewable Energy Laboratory, kWh Analytics, Github contributors`~
- make sure the makefile is equivalent to the custom build script
- ~spruce up the index page, maybe copy over the README.md content.  linking `README.md` with `index.rst` almost works well with a redirect rst file and the `m2r` package, but images don't work because the paths don't get translated.  Might need to rewrite the readme in RST to make it work.~
- add link to https://www.nrel.gov/pv/rdtools.html?
- ~make a page for each module~
- explanations of the functional and OO interfaces -- I vote to move this to a new PR or #117
- ~most of the links in README.md don't work after translation.  Probably there's a way to modify them to work in both the MD file on github and the generated RST on RTD.~


kanderso-nrel:
@mdeceglie @cdeline I think this is ready for review.  It's pretty huge though so I'd guess that it'd be easier to talk through everything in person.  Let me know what you think.

cdeline:
make HTML is working for me.  The docs look really good.

cdeline:
I'm getting an import error for statsmodels when importing from requirements.txt.  This is a known issue that appears to be fixed in 0.10.1 of statsmodels (?)  https://github.com/statsmodels/statsmodels/issues/4635

cdeline:
Alright, I got TravisCI happy again, but readthedocs.io is still timing out on the build.  https://readthedocs.org/projects/rdtools

cdeline:
At the moment I've been focusing on getting things building with requirements.txt, but the suggestion is to set up a separate requirements textfile for building documentation.  It can obviously be more streamlined to set up documentation rather than actually running RdTools.

cdeline:
Details on the resources available for buildthedocs are here. Other suggestions on reducing build time is on page 58 https://buildmedia.readthedocs.org/media/pdf/docs/latest/docs.pdf#page=34&zoom=100,0,584

kanderso-nrel:
I don't understand that output of the RTD builds -- it says the total time was 15 minutes, but the commands listed only total to a handful of seconds.  Maybe the issue is something like the job isn't configured to actually build the documentation, so the job never completes?

cdeline:
I believe it - trying to install all of those packages from a bare environment can take some time.  There's an option to use pre-installed packages on the server with a config.yml file.  PVlib takes that approach - I'll copy their config file and see if that helps anything.  Also, I'll set up a secon requirements.txt file just for the documentation - we can probably streamline the required packages a tad.

kanderso-nrel:
"master doc" failures might be because we aren't specifying it in `conf.py` -- see https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-master_doc

`index.rst` contains our root toctree, so I suppose it should be set to `'index'`.

cdeline:
Docs passing!
https://readthedocs.org/projects/rdtools/builds/9952713/

kanderso-nrel:
Very cool :) Only issue I see is that the screenshots aren't showing up on the landing page.  The problem is that the translation of README.md from markdown to RST doesn't update the relative paths to those images, so they don't work in the built HTML.  The build script I was using had a copy command at the end to bring the screenshots to the right place, but the makefile doesn't currently have that.  It might be possible to rewrite the paths in the markdown file so that they work in both the README on github and after translation on RTD, or maybe the build process should just copy them over like I was doing.

kanderso-nrel:
I hadn't noticed before, but the same issue (relative paths not getting translated) broke the links to example notebooks and such as well.  Maybe an easy fix for those links would just be to change them to point at RTD now that it exists?

cdeline:
Kevin - if you create a username on ReadTheDocs, I can add you as an administrator.  Right now it says kanderso-nrel does not exist when I try to add you as an administrator

kanderso-nrel:
Done, same account name.

cdeline:
Kevin - I was able to fix this by adding a symbolic link to sphinx/build/html back to /screenshots.  You don't need to change anything in readme.md

kanderso-nrel:
I was worried that a symlink wouldn't play nicely with Windows dev environments, but if it works then that's fine.  The latest commit (renaming the screenshots folder to _static and changing doc links from Github to RTD) appears to work as well.

cdeline:
OK, let's do it your way.  The symlinks are working weird with git.

Issue #127, opened 2019-11-11
Feature: add functions to calculate performance ratio


kanderso-nrel:
Add the ability to calculate performance ratio (eg https://www.nrel.gov/docs/fy13osti/57991.pdf) to quantify system performance.

PR #128, opened 2019-11-11
Feat/performance ratio

kanderso-nrel:
References #127.  Step one of the bigger loss categorization project is to add the ability to calculate performance ratio.  The method chosen for now is the temperature-adjusted metric defined in [1].  I've put the code in a new file called `losses.py` in anticipation of future loss-quantification functions going there as well.  

The main user-facing function `performance_ratio` differs from the spec in two ways:
- optionally allowing the user to filter out times where the expected power would be above the system's clipping limit
- defaulting to 25 C as the reference cell temperature instead of the POA-weighted Tcell value calculated from a system's 8760.  The POA-weighted Tcell version is intended to give the same annual value as the standard (non-temp-adjusted) PR, which results in the values being centered around 0.8 or so depending on the system.  Using 25 as the reference shifts the values to centered around 1.0.  

The other function `calculate_pr` is more generic and allows a user to calculate a PR using a generic expected energy signal and filtering mask.  

[1] "Weather-Corrected Performance Ratio" - https://www.nrel.gov/docs/fy13osti/57991.pdf

### DIFF
--- a/rdtools/__init__.py
+++ b/rdtools/__init__.py
@@ -13,6 +13,8 @@ from rdtools.filtering import poa_filter
 from rdtools.filtering import tcell_filter
 from rdtools.filtering import clip_filter
 from rdtools.soiling import soiling_srr
+from rdtools.losses import performance_ratio
+from rdtools.losses import calculate_pr
 
 from ._version import get_versions
 __version__ = get_versions()['version']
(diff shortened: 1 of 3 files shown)


kanderso-nrel:
Current Travis failures are from the py2.7 jobs, FYI

mdeceglie:
Closing this one, it's been stale for a while

Issue #102, opened 2018-10-25
Generalized PV model interface


mdeceglie:
Currently `rdtools.normalization` uses wrappers for some specific pvlib models. To maintain a clean, interoperable ecosystem for PV modeling and analysis, it would make sense to enable RdTools to work more generally with arbitrary models from pvlib. It seems to me that the logical way to achieve this is to develop normalization functions that take pvlib `ModelChain` objects as an input.

@wholmgren and @cwhanse, I'm interested to hear any thoughts (or warnings) you have at the outset.

cwhanse:
@mdeceglie are you referring, for example, to `rdtools.normalization.sapm_dc_power`? This function operates on a `pvlib` `LocalizedPVSystem` object, and performs a sequence of `pvlib.PVSystem` object methods (solar position, POA, cell temp, DC output). In that sense the `rdtools` function is doing what `ModelChain` object was designed for. It may be that `ModelChain` isn't as flexible as `rdtools` may want, since most of the development of `ModelChain` was motivated by making the modeling sequence convenient.

mdeceglie:
Considering the example of `rdtools.normalization.sapm_dc_power`, it implements a specific model. I'm interested in providing users the opportunity to specify arbitrary models to serve as the basis for normalization and subsequent analysis. Perhaps it's best to develop a prototype for what I'm thinking to help fuel discussion, but I wanted to see if there were any show stoppers I'm missing.

wholmgren:
I agree that it sounds a lot like what `ModelChain` was designed for but flexibility could be a concern. Let us know if you run into issues. I'm not that familiar with `rdtools` but it looks like `normalize_with_sapm(energy, sapm_kws)` could be `normalize_with_modelchain(energy, mc)` without too much effort.

A little off topic, but I think `LocalizedPVSystem` was a mistake, so use with caution.

mdeceglie:
To lay the groundwork for this (or potentially even close the issue), we should probably refactor `normalization.py` such that the there is a function which accepts measured and modeled PV, on potentially different time axes, and performs the normalization. Currently this logic is buried in each of the specific normalization functions.

PR #130, opened 2019-11-20
Feature: Model plugins

kanderso-nrel:
Two modelchain PRs didn't seem like enough so here's another one...  

This implementation follows a declarative approach where modeling steps are described in relation to each other without prescribing a specific control flow.  This allows automated analysis of the model chain itself and allows a higher level of customization in the overall RdTools analysis while still providing a high-level API that can run a lot of code without much work on the user's part.  

The idea is that each basic step in the RdTools pipeline (tcell calculation, normalization, aggregation etc) is a "model" that has a set of inputs and a set of outputs.  Instead of directly hooking up the output of the tcell model to the input of the normalization model, this approach decouples the relationship into the more abstract concepts "model X produces tcell" and "model Y requires tcell".  Since the normalization model now needs an abstract input "tcell" instead of the explicit output of a specific tcell model, this allows a user to swap in a new tcell model that will get used in place of the original.  When a given value is required, the provider of that value is looked up and any of its requirements are recursively evaluated.  For instance, when running the sensor-based degradation analysis, aggregated performance is required, which requires normalized performance, which requires cell temperature, which requires... Because each model has the metadata of what it requires and what it provides, this dependency chain can be automatically determined and evaluated from the base inputs.  Instead of telling RdTools to "run analysis X", it's "give me result X".  

Features:
- The default RdTools analysis can be run with only two function calls -- the SystemAnalysis init, and a call to `sa.calculate('result_of_interest')`, e.g. `sa.calculate('sensor_degradation_results')`
- This "plug-in" architecture allows any of the modeling steps in the RdTools analysis process to be customized as needed through custom code, not just kwargs
- New models can be added into the model graph by the user, so the RdTools calculations are extendable
- The individual outputs of each model are available to the user, allowing somewhat more granular inspection than the current ModelChains branch.  e.g. each filter series is available in addition to the overall `sensor_filter`. 
- Intermediate results are cached so things aren't calculated more than once. 

Here's the ported version of the ModelChains example notebook: https://github.com/NREL/rdtools/blob/model_plugins/docs/model_plugins_example.ipynb

And some preliminary documentation:
https://rdtools.readthedocs.io/en/model_plugins/system_analysis.html
https://rdtools.readthedocs.io/en/model_plugins/api.html#system-analysis

To-Do:
- [ ] Enforce pep8 everywhere
- [ ] Add tests/port tests from other model chains branches
- [ ] Add specific examples of how to customize a system analysis


kanderso-nrel:
A relevant issue: #102

wholmgren:
This is interesting! Have you considered using something like dask graphs? I got a dask version of pvlibs modelchain working last year but didn’t push it.

kanderso-nrel:
Thanks for the suggestion, dask graphs certainly look promising.  I read through some of the docs and came away with these thoughts:

- Dask plays nicely with parallelized graph computation (not sure that RdTools's use case would really benefit from this, but doesn't hurt to have around).
- Dask has caching, but since it's geared towards large datasets it can purge older results and might be less effective than ours.  Maybe not an issue since RdTools analyses fit in memory, so it won't have to dump older results.  
- Dask has nice graph visualization through GraphViz.
- Dask graph construction is more readable -- just a dict of function names to inputs/ouputs.  Graph construction code is separate from the suite of functions instead of a jumble of functions and decorators. 
- I'm not seeing evidence that `deferred` inputs are possible with Dask.  I don't think `dask.delayed` is a replacement, but could be wrong.  Not necessarily a deal breaker either way since the main use case for deferral is to avoid going down impossible code paths, which can also be done by just constructing the graph such that those code paths don't exist.  
- The graph is still customizable by users that want to replace a node with their own function -- instead of registering a plugin, you just overwrite that entry in the graph dictionary.  And we could wrap it and present a nicer API to the user. 
- Avoiding unnecessary calculations is possible via the `dask.optimize` functions, in particular `cull`.  
- Functions with multiple return values (eg our `aggregate` function, which returns both aggregated energy and insolation) are kind of clumsy -- I think the way to handle it is to leave `aggregate`'s return value as a tuple and then have two more tasks that just index the two elements out of the tuple. 

Overall I wish I had known about dask a couple weeks ago, maybe I wouldn't have reinvented this baby version of it.  @wholmgren why did you end up not using it?

wholmgren:
Great summary. 

Here's a link to my dask proof of concept for pvlib's ModelChain: https://github.com/wholmgren/pvlib-python/blob/dask/pvlib/mcdask.py

I prototyped it after some spirited discussions about the pvlib API. I didn't go through the effort of proposing it for pvlib because I didn't want to put in the work to convince the community that it was a good replacement/alternative to ModelChain. That you're now looking at similar ideas has me wondering if there's new interest and we could make some progress. Wouldn't blame you for wanting to stick with the ideas you've already invested in, though.

To be clear, my group does use a lot of dask, but the calls to pvlib are at a lower level than where we're using dask. [scipy talk](https://youtu.be/tYGcicSruck) and [AMS poster](https://forecasting.energy.arizona.edu/media/posters/Lorenzo_AMS2017.pdf) if you're interested.

kanderso-nrel:
Thanks for the links.  I'm curious, what was your motivation to use a graph-based approach for PVlib?  I see two main benefits of using a model graph instead of a standard imperative approach for RdTools:  

First, there are some ideas for RdTools to provide several different branches of analysis (degradation and soiling of course, availability, performance metrics, ...) out of the box.  These branches would share some common dependencies (normalized production, for instance) and it seemed like a mild annoyance to write code to make sure each of these dependencies is satisfied without wasting time on inputs for the other branches.  A model graph would resolve those dependencies optimally without any hand-tuning.  

Second, I didn't want to limit the scope of a system analysis to what we choose to include in RdTools.  Since there are a million ways to slice and dice operational data, it's easy for me to imagine a user wanting to add their own extensions (not just replacements) to the built-in model set.  To toss out some examples, maybe a given user is interested in calculating measured monthly transposition gain/poa-weighted cell temp/daily albedo profile/lost revenue from soiling/...  PVlib's [user-defined models](https://pvlib-python.readthedocs.io/en/stable/modelchain.html#user-defined-models) is a step in this direction in that it allows customization of the existing modeling steps, but (I think) there's no way to add brand new steps into the analysis.  A flexible model graph removes any implicit structural restrictions and puts user modeling logic on the same level as the built-in logic.  Whether or not this additional flexibility is worth the additional complexity, I'm not sure...  users can always just run their custom analyses themselves using our outputs.

kanderso-nrel:
FYI to any interested parties, I put together a function that converts the model chain's internal graph to a dask task graph.  Dask gives the same results for the test case in the notebook, see here for the proof (towards the bottom of the notebook): https://github.com/NREL/rdtools/blob/model_plugins/docs/model_plugins_example.ipynb

I wouldn't expect a different result -- just a proof of concept that dask works for at least our basic analysis case.

cwhanse:
> Thanks for the links. I'm curious, what was your motivation to use a graph-based approach for PVlib?

I don't know that this answers the question you are asking, but my $0.02: pvlib is organized around a fairly linear process diagram (at www.pvpmc.org) that describes the steps to predict power and energy from solar power systems. ModelChain was conceived as a way to record the selections and to simplify executing the selected models for this use case (predict power). Forecasting power (the dask application) follows a similar, basically linear flow in two stages: forecast weather, convert weather to power.

wholmgren:
@kanderso-nrel you've pretty much described my motivations for exploring a graph. Along with the graph, I also liked that the dask scheduler gives you the ability to cache results and distribute workloads. 

I agree with @cwhanse about the organization of pvlib and the motivation of ModelChain. But while pvlib might be organized to follow that linear process, it doesn't require anyone to use it that way. ModelChain does.

An imperfect analogy that comes to mind is that ModelChain is like a walled garden, while maybe a graph based version of it could be more like an open bazaar. Perhaps each have their places.

>  PVlib's user-defined models is a step in this direction in that it allows customization of the existing modeling steps, but (I think) there's no way to add brand new steps into the analysis.

Yes, this is a major problem. I suspect that a grand total of 1 person (me) has ever used ModelChain's support for user defined functions. The kind of person that writes a user defined function for ModelChain is the kind of person that is happy to roll their own sequence of basic pvlib function calls for their bespoke analysis. But would that kind of person be interested in building on a pvlib-supplied graph if it gave them all of the other benefits of dask? Seems more compelling than trying to fit into the walled garden of ModelChain.

mikofski:
This discussion is super interesting. I won't pretend that I understand all of the details, but I will just add, that **IMHO it's better to use existing tools then to try to roll your own**.

I think that creating a plugin framework, task scheduler, or any of the things that Dask and other existing frameworks already provide is going to be difficult. Even if what you want right now seems simple, maintaining it will be a challenge. IMO if there's a feature you want that's not in Dask, open a PR there and add it, then vendorize your feature locally until it's merged. Or better yet, try to see if you can reframe what you want to fit the existing framework as is.

I don't mean to discourage you, because I know you could probably do it, but IMO our skills are better put to solving solar, storage, and weather challenges. That's my 2&cent;

mikofski:
BTW: check out @cedricleroy 's [pyungo](../../../cedricleroy/pyungo) and [pyungo-viz](../../../cedricleroy/pyungo-viz) - I think they may be very close to what you're doing here with the `@mc.plugin(requires...)` decorators.

kanderso-nrel:
> IMHO it's better to use existing tools

I completely agree; I should have included a bullet point in my list that said "A big plus is that Dask is maintained by someone else, and they do it better".  And thanks for the pyungo links... I have learned my lesson that before I ever write another line of code, I should ask you guys if there's already a library to do it 😄 

For what it's worth at this point I do think Dask is the better choice.  I'll keep fleshing out the Dask-based version and make sure no blockers pop up.

mikofski:
I should have tempered my comments. These types of open investigations are an important part of learning and exploring the bigger picture outside of our specific field, even if the outcome is just to choose which existing tool to use. Thanks for sharing!

mikofski:
In addition to Dask, you should check out [AirFlow](https://airflow.apache.org/), another task scheduler, which let's you schedule any task, not just Python. For example read [how the Auth0 data team uses R and Python](https://auth0.com/blog/how-the-auth0-data-team-uses-r-and-python/). Probably overkill, but it's another dimension to the puzzle

kanderso-nrel:
I've got a preliminary Dask-based version on the `feat/model_chain_dask` branch.  The code is definitely simpler using Dask, although now there's 200 lines of dictionary definitions.  I've only run into two real drawbacks so far:

- Dask error messages aren't as informative as I'd like in that they don't provide any additional information beyond the python default.  Not a big deal, I added better logging over top that reports where in the graph the error occurred.  
- Caching (via dask.cache.Cache) is done globally instead of at the SystemAnalysis level.  This means that multiple SA objects in the same scope would share the same cache, and they'll all use the same variable names, so results from one will leak into the others.  The good news is that using this cache isn't necessary if you know ahead of time all the outputs you're interested in and calculate them all at once with a single `sa.calculate` call.  The bad news is that this isn't compatible with the "persistent results stored on the object" goal.  This could be solved by implementing our own SA-level cache or changing keys to be unique to each SA. 

Here's the updated example notebook: 
https://github.com/NREL/rdtools/blob/feat/model_chain_dask/docs/model_plugins_example.ipynb

and implementation:
https://github.com/NREL/rdtools/blob/feat/model_chain_dask/rdtools/system_analysis.py

mikofski:
Sorry to add gas to the fire, but check out [xarray-simlab](
https://github.com/benbovy/xarray-simlab)

mdeceglie:
We've decided to move on with #117 for this functionality. This is a potentially powerful approach, so maybe we'll refactor everything in this style someday.

PR #131, opened 2019-11-22
Feat/inverter availability

kanderso-nrel:
Code for inverter availability calculations.  Two main functions:
- is_online, to detect when inverters are offline
- downtime_losses, to estimate lost production

Detecting offline inverters should be easy, except communication outages make it tricky.  The code assumes that meter-level data is reliable and never has comms outages but is mostly robust to inverter comms outages.  I believe the only thing it will fail at (assuming good meter data) is when some inverters are truly offline and some are only non-communicative.  In that case the code says that they're all offline.

For lost production, the preferred method is inter-inverter comparison -- examining how inverters have performed relative to each other in the past and assuming the same relative performance can be used to calculate lost production when some inverters are offline.  If all inverters are determined to be offline, an expected_power signal provided by the user is used to determine lost energy.  If there is no expected power data for the outage (eg the entire system is offline, so there's no weather data), a user-provided 12x24 typical production profile is used instead.  

Documentation links:
- https://rdtools.readthedocs.io/en/feat-inverter_availability/availability_example.html
- https://rdtools.readthedocs.io/en/feat-inverter_availability/generated/rdtools.losses.html#module-rdtools.losses

---

Edit for anyone reading this -- the above info is mostly about the original implementation and is largely out of date.

- [x] Updated changelog ~(waiting for #193 to get merged with the whatsnew file)~


kanderso-nrel:
Picking this PR up again to implement the methods in Anderson & Blumenthal "Overcoming communications outages in inverter downtime analysis", PVSC 47

kanderso-nrel:
This is getting close.  Things that could use some input if anyone is interested:
- I'm not wild about the names of the three new functions: `availability.loss_from_energy`, `availability.loss_from_power`, `plotting.availability_cumulative_plot`.  Here are some phrases that come to mind for inspiration:
    + `loss_from_power`: "peer to peer comparison", "component analysis", "parts of a whole", "partial outage"
    + `loss_from_energy`: "uncertainty by duration", "bridging the gap", "full outage"
- `loss_from_energy` is fairly complex and might benefit from being wrapped in a class like the soiling SRR code -- it would be nice to have access to the private intermediate variables for inspection if desired
- Should availability have a section on the [docs landing page](https://rdtools.readthedocs.io/en/latest/) like degradation and soiling do?
- It would be good to allow the user to configure the thresholds in `loss_from_energy`, either through optional kwargs that default to the 1st and 99th percentile, or by adding a new column in the return value that specifies where in the error distribution the outage value fell (rather than just the boolean of whether it's below the 1st percentile or not).

Still need to do more thorough testing on real datasets; could be that more tweaks are coming next week.

kanderso-nrel:
OK, finally ready for review.  After trying it out on many real datasets, I am happy with its behavior when the input timeseries are decent quality.  I've also added some checks to try to give reasonable output when the input doesn't obey the laws of physics, but there's only so much you can do to get around "garbage in, garbage out".  Here are the questions that are still open about the API and docs:

- I'm not wild about the names of the three new functions: `availability.loss_from_energy`, `availability.loss_from_power`, `plotting.availability_cumulative_plot`.  Here are some phrases that come to mind for inspiration:
    + `loss_from_power`: "peer to peer comparison", "component analysis", "parts of a whole", "partial outage"
    + `loss_from_energy`: "uncertainty by duration", "bridging the gap", "full outage"
- `loss_from_energy` is fairly complex and may benefit from being wrapped in a class like the soiling SRR code -- it would be nice to have access to the private intermediate variables for inspection if desired.  Also I would guess that the typical use case would involve both methods, so maybe an overall availability class that runs both is the way to go?
- Should availability have a section on the [docs landing page](https://rdtools.readthedocs.io/en/latest/) like degradation and soiling do?

cdeline:
Looking good @kanderso-nrel . 
- I do think that a paragraph and a figure on the RdTools docs landing page is appropriate.  this is more of an advertisement of this capability than detailed description, so a short description and an attention-grabbing picture would work here.
- I actually don't have any problem with the function names that you've proposed.  Maybe we can discuss IRL.
- I'd have to take a closer look at `the loss_from_energy` function to comment.

mdeceglie:
I agree that adding some info about this to the landing page would be good. This is a fundamentally different workflow, we should try to give it equal weight to the degradation/soiling workflow.

rdtools/plotting.py
```diff
@@ -230,3 +231,39 @@ def soiling_rate_histogram(soiling_info, bins=None):
...
+    """
+    fig, ax = plt.subplots()
+    cumulative_energy.plot(ax=ax, label='Cumulative Energy')
+    for i, row in outage_info.iterrows():
+        start, end = row[['start', 'end']]
+        start_energy = row['start_energy']
+        expected_production = row['expected_energy']
+        lo, hi = np.abs(expected_production - row[['ci_lower', 'ci_upper']])
+        raw_expected_energy = expected_energy.loc[start:end].sum()
+        scale_factor = expected_production / raw_expected_energy
```
mdeceglie, line 262: I think this scaling is bit different than the scaling calculation in `loss_from_energy()` would it make sense to align them? If so, perhaps make the scaling function stand alone so it can be reused. If you go with an OO approach, it could also be got from the object.

rdtools/test/availability_test.py
```diff
@@ -0,0 +1,314 @@
...
+
+import pvlib
+import pandas as pd
+import numpy as np
+import itertools
+import datetime
+
+# Values to parametrize power tests across.  One test will be run for each
+# combination. Can't be careless about expanding this list because of
+# combinatorial explosion.
```
mdeceglie, line 18: Sounds like a blast

mdeceglie (commented):
This is some powerful new functionality. It will be exciting to have it in RdTools!

1. I suggest adding a note in the markdown below the first figure in the example to the effect of "The solid lines show the data that would be available in our example. The dashed lines show the true underlying behavior (that we normally wouldn't know)"

2. In `loss_from_power` I suggest adding a validation that `subsystem_power` and `system_power` have the same index, and raising a descriptive exception if they don't

3. In the example for `loss_from_power` how come the the reported+estimated sometimes deviates from the the satellite. For example on the 1st and  7th, it exceeds the satellite, and on the 4th and 5th, it is low relative to satellite.

4. In `loss_from_power` let's consider using NaN isntead of 0 for instances where the method fails to estimate a loss. It is helpful to distinguish a true estimae of 0 lost power from not having an estimate.

6. I suggest adding a legend to `availability_cumulative_plot`

7. Would it make sense to add a function for the final system availability calculation in the notebook?

8. In that final availability calculation, why is the actual production not based on df['meter_energy']?

9. In the final calculation, are the two methods gurateed not to double count?

10. Consider providing and option for known relative sizing in loss_from_power

11. Please check that all new function parameters align with pvterms where helpful

mdeceglie:
I think we should consider building the whole module around a single availability class. It might improve a number of things. Like the fact that the first step of `loss_from_energy` is to run `loss_from_power`. It would also help address the fact that the outputs of the two functions are so drastically different. For example you could also produce a dataframe with every timestep  and columns "outage type" and "power lost" that takes into account both legs of the analysis. Yet another advantage would be the final calculation in the notebook being a natural method. Finally intermediate values being available for inspection as you note (and the plotting function for that matter).

rdtools/availability.py
```diff
@@ -0,0 +1,334 @@
...
+    Returns
+    -------
+    outage_info : pd.DataFrame
+        A dataframe of records about each detected outage, one row per outage.
+        The primary columns of interest are ``type``, which can be either
+        ``'real'`` or ``'comms'`` and reports whether the outage was determined
+        to be a real outage with lost production or just a communications
+        interruption with no production impact; and ``loss`` which reports
+        the estimated production loss for the outage.
+
```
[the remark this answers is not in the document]
kanderso-nrel: Documented in the attribute description now that `outage_info` is an attribute: 
https://rdtools--131.org.readthedocs.build/en/131/generated/rdtools.availability.AvailabilityAnalysis.html#rdtools.availability.AvailabilityAnalysis.outage_info

rdtools/availability.py
```diff
@@ -0,0 +1,334 @@
...
+    df['Expected Energy'] = rdtools.energy_from_power(df['Expected Power'])
+    df['Meter_kWh_interval'] = rdtools.energy_from_power(df['Meter_kW'])
+
+    df_subset = df.loc[subset, :]
+
+    # window length is "number of daytime intervals".
+    # Note: the logspace bounds are currently kind of arbitrary
+    window_lengths = np.logspace(np.log10(12),
+                                 np.log10(0.75*len(df_subset)),
+                                 10).astype(int)
```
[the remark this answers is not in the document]
kanderso-nrel: Switched to `window_lengths = 2**np.arange(1, int(np.log2(len(df_subset))), 1)`.  I think this solves the issue, but I could easily be convinced that an even better alternative exists.

rdtools/availability.py
```diff
@@ -0,0 +1,334 @@
...
+    # functions mapping number of intervals (outage length) to error bounds
+    predict_upper = lambda x: float(interp(upper)(x))
+    predict_lower = lambda x: float(interp(lower)(x))
+
+    # Calculate boolean series to indicate full outages. Considerations:
+    # - Multi-day outages need to span across nights
+    # - Full outages don't always take out communications, so the cumulative
+    #   meter can either drop out or stay constant depending on the case
+    # During a full outage, no inverters will report production:
+    low_threshold = subsystem_power.quantile(0.99) / 1000
```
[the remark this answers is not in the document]
kanderso-nrel: `low_threshold` not needed here anymore because it reuses the mask from `loss_from_power` now.

rdtools/availability.py
```diff
@@ -0,0 +1,334 @@
...
+            'error_upper': predict_upper(daylight_intervals),
+            'expected_energy': df_outage['Expected Energy'].sum(),
+            'start_energy': df.loc[start_minus_one, 'Meter_kWh'],
+            'end_energy': df.loc[end, 'Meter_kWh'],
+        }
+        outage_data.append(data)
+
+    df_outages = pd.DataFrame(outage_data)
+    # pandas < 0.25.0 sorts columns alphabetically.  revert to dict order:
+    df_outages = df_outages[data.keys()]
```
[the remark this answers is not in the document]
kanderso-nrel: I think it's unofficially guaranteed in 3.6 (at least for CPython) but it was convenient to fix this while making sure the dataframe still has all columns in the case of a dataset without any outages.

rdtools/plotting.py
```diff
@@ -230,3 +231,39 @@ def soiling_rate_histogram(soiling_info, bins=None):
     ax.set_ylabel('Count')
 
     return fig
+
+
+def availability_cumulative_plot(outage_info, cumulative_energy,
```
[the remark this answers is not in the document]
kanderso-nrel: It's all `energy_cumulative` now

kanderso-nrel:
@mdeceglie thanks for these detailed notes.  I think I've addressed almost all of them -- some responses below:

> I suggest adding a note in the markdown below the first figure in the example to the effect of "The solid lines show the data that would be available in our example. The dashed lines show the true underlying behavior (that we normally wouldn't know)"

Done

> In loss_from_power I suggest adding a validation that subsystem_power and system_power have the same index, and raising a descriptive exception if they don't

> I suggest explicitly validating [that the input timeseries are continuous] on the input and raising an exception if the assumption isn't met.

The first of these is easy enough (but not yet done), but I'm not sure what to think about the second.  These methods don't require a regular timeseries, just a continuous one -- so if it checked for large holes in the timeseries, there would have to be a `max_timedelta`-style parameter to define what "large" means. 

> In the example for loss_from_power how come the the reported+estimated sometimes deviates from the the satellite. For example on the 1st and 7th, it exceeds the satellite, and on the 4th and 5th, it is low relative to satellite.

This actually isn't plotted anymore, but it was just because the "satellite" data had noise added to it to be a little more realistic.

> In loss_from_power let's consider using NaN isntead of 0 for instances where the method fails to estimate a loss. It is helpful to distinguish a true estimae of 0 lost power from not having an estimate.

I've added two timeseries attributes, `loss_system` and `loss_subsystem` that reflect the estimated interval power loss from the two methods.  I also added `loss_total` that is the sum of the two.  The way I'm thinking about it, a value of zero in `loss_subsystem` during a full outage is expected because there would be a nonzero value in `loss_system`.  I like the idea of reserving NaN for "something is unexpectedly wrong" in the sense that the user should check their input data for validity.  Thoughts?

> I suggest adding a legend to availability_cumulative_plot

The plot has changed in other ways, but legends are now present where it seemed appropriate.

> Would it make sense to add a function for the final system availability calculation in the notebook?

Added, see `AvailabilityAnalysis._combine_losses()`

> In that final availability calculation, why is the actual production not based on df['meter_energy']?

That calculation is changed so it's no longer an issue, but it was a lazy way of making units line up.  The loss was calculated from lost power, and so therefore the point of comparison should be power too.

> In the final calculation, are the two methods gurateed not to double count?

I'm not 100% sure, but I think so.  The `loss_from_energy` process detects outages based on `looks_offline = ~self.reporting_mask.any(axis=1)`, where the reporting_mask variable is a boolean matrix of inverters reporting production. So it only runs when no inverters are reporting production, which is the exact failure condition for `loss_from_power`.

Maybe it should do an xor-style check between the two loss timeseries and raise a warning if there is any overlap? 

> Consider providing and option for known relative sizing in loss_from_power

Added via optional `relative_sizes` parameter to `AvailabilityAnalysis`.

> Please check that all new function parameters align with pvterms where helpful

Thanks for the reminder.  I did a pass, feel free to point out any that I missed.

rdtools/availability.py
```diff
@@ -0,0 +1,334 @@
...
+    This method's ability to detect short-duration outages is limited by the
+    resolution of the system data. For instance, 15-minute averages would not
+    be able to resolve the rapid power cycling of an intermittent inverter.
+    Additionally, the loss at the edges of an outage may be underestimated
+    because of masking by the interval averages.
+
+    The outage detection routine assumes that the input timeseries are
+    continuous, so any data gaps should be filled with nan before using
+    this function.  See :py:func:`rdtools.normalization.interpolate`.
+
```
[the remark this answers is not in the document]
mdeceglie: @kanderso-nrel In reference to your response below. I'm not sure I follow the requirement. How would you describe a non-continuous time series?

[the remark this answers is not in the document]
kanderso-nrel: It might be better described as a continuous datetime index.  For example this hourly dataset is discontinuous because it is missing the 01:00 row:
```
2019-01-01 00:00:00    1.0
2019-01-01 02:00:00    3.0
2019-01-01 03:00:00    4.0
```
So to make the index continuous, add a nan row:
```
2019-01-01 00:00:00    1.0
2019-01-01 01:00:00    NaN
2019-01-01 02:00:00    3.0
2019-01-01 03:00:00    4.0
```
It's okay for measured values to be missing -- the issue is only with missing timestamps/rows.

rdtools/availability.py
```diff
@@ -0,0 +1,565 @@
...
+        |                      | duration.                                    |
+        +----------------------+----------------------------------------------+
+        | `energy_start`       | System cumulative production at the outage   |
+        |                      | start.                                       |
+        +----------------------+----------------------------------------------+
+        | `energy_end`         | System cumulative production at the outage   |
+        |                      | end.                                         |
+        +----------------------+----------------------------------------------+
+        | `energy_actual`      | System production during the outage (*i.e.*, |
+        |                      | `end_energy - start_energy`).                |
```
mdeceglie, line 129:
```suggestion
        |                      | `energy_end - energy_start`).                |
```

rdtools/availability.py
```diff
@@ -0,0 +1,565 @@
...
+        These values must be recorded at the device itself (rather than summed
+        by a downstream device like a datalogger or DAS provider) to preserve
+        its integrity across communication interruptions. Units must match
+        ``power`` integrated to hourly energy (e.g. if ``power`` is in kW then
+        ``energy`` must be in kWh).
+
+    power_expected : pd.Series
+        Expected system power data with the same index as the measured data.
+        This can be modeled from on-site weather measurements if there is no
+        risk of instrument calibration or data gaps. However, because full
```
mdeceglie, line 42: Can we rephrase slightly? Maybe "This can be modeled from on-site weather measurements if instruments are well calibrated and there is no risk of data gaps"

rdtools/availability.py
```diff
@@ -0,0 +1,565 @@
...
+    results : pd.DataFrame
+        Rolled-up production, loss, and availability metrics. The index is
+        a datetime index of the period passed to
+        :py:meth:`AvailabilityAnalysis.run`. The columns of the dataframe are
+        as follows:
+
+        +----------------------+----------------------------------------------+
+        | Column Name          | Description                                  |
+        +======================+==============================================+
+        | `lost_production`    | Production loss from outages                 |
```
mdeceglie, line 58: Let's add a note about units

rdtools/availability.py
```diff
@@ -0,0 +1,565 @@
...
+        +----------------------+----------------------------------------------+
+        | `actual_production`  | System energy production.                    |
```
mdeceglie, line 60: Same suggestion re:units

rdtools/availability.py
```diff
@@ -0,0 +1,565 @@
...
+        """
+        self._loss_from_power(low_threshold, relative_sizes,
+                              power_system_limit)
+        self._error_distributions(quantiles)
+        self._loss_from_energy()
+        self._combine_losses(rollup_period)
+
+    def plot(self):
+        """
+        Create a figure summarizing the availability analysis results.
```
mdeceglie, line 556: Though it may be obvious to experienced python users, for someone without much experience in OOP, it may not be obvious that you need to run `AvailabilityAnalysis.run()` first. I suggest adding a note here and a descriptive exception in case someone tries to run this without having run the analysis first.

rdtools/availability.py
kanderso-nrel: "no risk of instrument calibration" haha.  What a nightmare it would be if the instruments were calibrated

rdtools/availability.py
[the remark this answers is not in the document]
cwhanse: I've seen the term "serially complete" used to describe a time series where time points are regular and there are no missing values.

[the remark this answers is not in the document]
mdeceglie: Is it required that each value in the index be evenly spaced? If so I might describe it as requiring "a regular index of evenly spaced timestamps" to be explicit about it.

If that is the requirement, then we could check that `index.freq is not None` Although I believe it is possible to have evenly spaced points and `freq==None`, perhaps the most conservative thing to do is raise the error in that case.

rdtools/availability.py
```diff
@@ -0,0 +1,565 @@
...
+            skipped.
+
+        quantiles : 2-element tuple, default (0.01, 0.99)
+            The quantiles of the error distribution used for the expected
+            energy confidence interval. The lower bound is used to classify
+            outages as either (1) a simple communication interruption with
+            no production loss or (2) a power outage with an associated
+            production loss estimate.
+
+        rollup_period : pandas offset string, default 'm'
```
mdeceglie, line 545: Will any [DateOffset](https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#dateoffset-objects) object work here? If so I suggest calling this a "pandas DateOffset or alias".

On a related note, I was surprised to see it lower case. Though it works, perhaps it would be best to adopt the official capitalized offset alias throughout the code and example.

rdtools/availability.py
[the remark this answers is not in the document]
cwhanse: `index.freq` is not always filled in when an index is created.

```
import pandas as pd
dt = pd.date_range(start='2020-09-22 00:00:00', end='2020-09-22 23:00:00', periods=24)
dt.freq is None
```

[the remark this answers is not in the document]
kanderso-nrel: The NSRDB is frequently advertised as being serially complete.  I don't think we can use it here because even spacing is not a requirement.  This code detects outages by looking at the data values provided to it (numbers or nan), not by trying to find jumps in the index.  The intent is to prevent outages from being overlooked because they were hidden by jumps in the index.  

How about this?

> This function expects outages to be represented in the timeseries by nan, zero, or very low values.  If your DAS does not report data with missing values (so an outage would create a jump in the reported timestamps), you should insert those missing rows before using this function.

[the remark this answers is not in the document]
mdeceglie: That makes sense to me, assuming I have understood correctly that your function doesn't assume regular intervals. E.g. It would work with an irregular/uneven time series as long as outages are represented by nan/zero/low

mdeceglie:
@kanderso-nrel  what do you think about rotating the x labels on the right hand plot like this (but within the plotting function)? I'm not well versed in the pandas plotting interface, so I can't explain why they get rotated on the left plots but not the right. Perhaps we should explicitly rotate both? 

Screen Shot 2020-09-22 at 3 39 13 PM

rdtools/availability.py
mdeceglie: Thanks! I suggest d->D in the notebook too

kanderso-nrel:
> why they get rotated on the left plots but not the right

It's the difference between `series.plot(ax=ax)` and `ax.plot(series)`.  Pandas has a default set of matplotlib formatting params it uses when you do `series.plot(ax=ax)`.  The function now uses pandas plotting everywhere so that the formatting is applied consistently: https://github.com/NREL/rdtools/blob/feat/inverter_availability/docs/system_availability_example.ipynb

rdtools/availability.py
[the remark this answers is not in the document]
cwhanse: It's not a major point, but I think that revision trades one ambiguity for a different one. How are you defining a "jump" in the timestamps if the timestamps are not on regular intervals? And if regular intervals are not required, does the algorithm still work if the timestamps with missing data are simply omitted?

mdeceglie:
@kanderso-nrel  I suggest adding the reference for your work on the availability algorithm.

kanderso-nrel:
> @kanderso-nrel I suggest adding the reference for your work on the availability algorithm on the overview docs page

Done (despite brain fart of "soiling" in the commit message).  I also took the liberty of adding one for the degradation section.

rdtools/availability.py
```diff
@@ -0,0 +1,576 @@
...
+        relative_sizes : dict or pd.Series, optional
+            The production capacity of each subsystem, normalized by the mean
+            subsystem capacity. If not specified, it will be estimated from
+            power data.
+
+        power_system_limit : float or pd.Series, optional
+            An optional maximum system power used as an upper limit for
+            (power_system + lost_power) so that the maximum system capacity or
+            interconnection limit is not exceeded. If omitted, that check is
+            skipped.
```
mdeceglie, line 540: This could use a little more explanation in the new OO context. In particular it isn't immediately clear what `lost_power` is and that this check is in the subsystem part of the routine. On a related note, would it make sense to add this check to the system-level part of the analysis?

rdtools/availability.py
```diff
@@ -0,0 +1,576 @@
...
+import pandas as pd
+import numpy as np
+from scipy.interpolate import interp1d
+
+
+class AvailabilityAnalysis:
+    """
+    A class to perform system availability and loss analysis.
+
+    This class follows the analysis procedure described in [1]_.
```
mdeceglie, line 17:
```suggestion
    This class follows the analysis procedure described in [1]_, and implements two
    distinct algorithms. One for partial (subsystem) outages and one for system-wide outages. 
    The `run()` method executes both algorithms and combines their results.
```

rdtools/availability.py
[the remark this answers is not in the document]
kanderso-nrel: I guess the point I'm trying to document is "don't expect the algorithm to find any outages that are shorter than your data intervals, and by the way your data intervals might be way longer than you expect if any rows got dropped during outages".

I'm expecting most users of this code to have a semi-curated dataset from their DAS provider with regular timestamps except when rows get dropped.  In that case, a simple pandas reindex or resample is what is needed before using this code.  For datasets with irregular measurement timing, I see the point that a "jump" is poorly defined, but I'm having trouble coming up with something more precise.  Suggestions are welcome!

mdeceglie (commented):
I think this is looking good. Here are a couple suggestions.

rdtools/availability.py
```diff
@@ -0,0 +1,576 @@
...
+        )
+
+        lost_production = rdtools.energy_from_power(self.loss_total)
+        df = pd.DataFrame({
+            'lost_production': lost_production.resample(rollup_period).sum(),
+            'actual_production': actual_production,
+        })
+        loss_plus_actual = df['lost_production'] + df['actual_production']
+        df['availability'] = 1 - df['lost_production'] / loss_plus_actual
+        self.results = df.tz_localize(None)
```
mdeceglie, line 511: What's the reasoning for this line? Why not just use the same timezone as the original input data?

kanderso-nrel: That line is a holdover from when this was internal code.  The reasoning was that, for monthly results, the timezone is not meaningful and was just getting in the way of merging the roll-up datasets with different timezones.  The merging aspect is less relevant here, and I still think it doesn't really make sense for these timestamps to have timezones, but maybe that's too presumptuous and we should let the user drop the tz afterwards if they want to.

rdtools/availability.py
kanderso-nrel: The missing context here is large systems with plant controllers and dynamic inverter setpoints.  For example, a system with an interconnection limit of 20MW might have 14x 1.5MW inverters for a total of 14 * 1.5 = 21MW total capacity, but the inverters are normally operated with a clipping point of 20/14=1.428MW so that the system doesn't go over its 20MW limit.  But if one inverter trips offline, the other inverters are allowed to go all the way up to their hard limit of 1.5MW, meaning the system produces 13*1.5=19.5MW in full sun, so really only 0.5MW is lost instead of 1.5 (or 1.428) and naively estimating loss at 1.5 or 1.428 MW would be incorrect.

With that out of the way, I agree that the parameter description needs improving.  Do you think it's worthwhile to describe the use case like the above paragraph?  Or should we just keep it simple and be clearer about what it does. 

> would it make sense to add this check to the system-level part of the analysis?

I don't think the system-level method is really susceptible to the error that this check tries to correct.  The time series system-level power loss (`loss_system`) is created from a scalar loss that is distributed in proportion to the time series expected power signal because that's really the only guidance we have on how to distribute the overall production loss across the outage duration.  So while the system-level loss time series might sometimes exceed the system capacity, that isn't indicative of an error in the loss estimate, just an error of how the loss estimate got distributed across the outage.  I do have a way of distributing production while obeying a maximum capacity, but the code wasn't very elegant and it wasn't always numerically stable, so I didn't pursue it here.

rdtools/availability.py
mdeceglie: I would err on the side of not stripping out the time zone information. I'm trying to think generally. For example, folks could be interested in their hourly availability, in which case time zone information may be desirable to understand what is going on across the fleet.

rdtools/availability.py
mdeceglie: Thanks that helps me understand. Let's work through the system-level question before deciding on new docs for it. On that system-level point, I see how the original intent of that parameter isn't really relevant. But, should the same parameter have influence? For example, should it be used to clip power in the calculation of `energy_expected_rescaled`?

kanderso-nrel: Hmm, interesting point that the clipping point isn't preserved when rescaling the expected energy to match total actual production.  As long as the expected energy is reasonably accurate (a handful of percent error), it wouldn't be a big error.  If we did make the rescaling operation preserve the system power limit, I think we would need to use some sort of iterative convergent method to still have the property that `energy_expected_rescaled` matches actual production in aggregate during normal production times, which is what I was hoping to avoid.  I would rather just document its limitations and encourage the user to pass in a better expected energy signal if it's a concern. 

Do you see another way to make use of it?

rdtools/availability.py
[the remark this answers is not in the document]
kanderso-nrel: Revised sentence: "If your DAS does not record data from outages (e.g., a three-hour outage results in three hours of omitted timestamps), you should..."

rdtools/availability.py
kanderso-nrel: > Maximum allowable system power. This parameter is used to account for cases where online subsystems can partially mitigate the loss of an offline subsystem, for example a system with a plant controller and dynamic inverter setpoints.  This constraint is only applied to the subsystem loss calculation.

mdeceglie (approved):
LGTM Thanks for putting this together @kanderso-nrel !

mdeceglie:
Should we use double backticks in the tables describing dataframes?

kanderso-nrel:
@mdeceglie I added an "other uses" section in the notebook with an example like we discussed in the meeting earlier today, of running the entire thing at the inverter level instead of at the system level: https://rdtools--131.org.readthedocs.build/en/131/system_availability_example.html#Other-use-cases

The other option (omitting the cumulative energy data) would require some more in-depth changes to get working, so I think we should leave it as a possible future improvement. 

Also made a small change to how the outage bounds are reported; the end of the outage was previously being reported as the timestamp of the first non-outage value rather than the last outage value.  This was introducing a small error into the calculated actual energy during the outage, because it was including that single interval of energy after the outage.

Issue #132, opened 2020-01-10
Model_Chains pv_nameplate error


cdeline:
Error results when passing in 'pv_nameplate' input into SystemAnalysis object.

minimum example:
sa = rdtools.SystemAnalysis(df['power_E'], df['poa'], 
                             cell_temperature=df['Tcell'],
                             pvlib_location=loc,
                             pv_tilt=meta['tilt'],
                             pv_azimuth=meta['azimuth'],
                             pv_nameplate = meta['pdc']
                             )#

Error:
---------------------------------------------------------------------------
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-125-55293c7e94e3> in <module>()
----> 1 sa_E.sensor_analysis(analyses=['yoy_degradation'])

c:\users\cdeline\documents\python scripts\kwh_rdtools\rdtools\system_analysis.py in sensor_analysis(self, analyses, yoy_kwargs, srr_kwargs)
    479         '''
    480 
--> 481         self.sensor_preprocess()
    482         sensor_results = {}
    483 

c:\users\cdeline\documents\python scripts\kwh_rdtools\rdtools\system_analysis.py in sensor_preprocess(self)
    438         if self.cell_temperature is None:
    439             self.cell_temperature = self.calc_cell_temperature(self.poa, self.windspeed, self.ambient_temperature)
--> 440         normalized, insolation = self.pvwatts_norm(self.poa, self.cell_temperature)
    441         self.filter(normalized, 'sensor')
    442         aggregated, aggregated_insolation = self.aggregate(normalized[self.sensor_filter], insolation[self.sensor_filter])

c:\users\cdeline\documents\python scripts\kwh_rdtools\rdtools\system_analysis.py in pvwatts_norm(self, poa, cell_temperature)
    257 
    258         pvwatts_kws = {"poa_global": poa,
--> 259                        "P_ref": pv_nameplate,
    260                        "T_cell": cell_temperature,
    261                        "G_ref": 1000,

UnboundLocalError: local variable 'pv_nameplate' referenced before assignment

Solution: has to do with pv_nameplate and self.pv_nameplate naming error.  Will submit pull request momentarily.

kanderso-nrel:
Closed by #133?

cdeline:
Yes, PR #133 fixes it in model_chains

PR #133, opened 2020-01-10
Fixes system_analysis bug #132 when pv_nameplate is passed into syste…

cdeline:
…m_analysis object.
pretty straightforward- sets pv_nameplate = self.pv_nameplate if the value exists.

Also, unrelated edit adding optional input to plotting.degradation_summary_plots(hist_nbins=200). Custom definition of the number of bins in the Rd histogram.  For data sets with small numbers of YoY pairs, the default is to only plot 1/40th of the datapoints.  This sets the default number of bins to 200, but allows it to be manually set.

### DIFF
--- a/rdtools/plotting.py
+++ b/rdtools/plotting.py
@@ -2,9 +2,9 @@ import matplotlib.pyplot as plt
 
 
 def degradation_summary_plots(yoy_rd, yoy_ci, yoy_info, normalized_yield,
-                              hist_xmin=None, hist_xmax=None,  hist_nbins=200, 
-                              scatter_ymin=None, scatter_ymax=None, plot_color=None, 
-                              summary_title=None, scatter_alpha=0.5):
+                              hist_xmin=None, hist_xmax=None, scatter_ymin=None,
+                              scatter_ymax=None, plot_color=None, summary_title=None,
+                              scatter_alpha=0.5):
     '''
     Description
     -----------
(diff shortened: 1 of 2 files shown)


cdeline:
I checked this out, and tests are passing.  Example notebooks are only slightly changed - the default histogram plot uses 75 bins which is close to the value used in our example notebook.

rdtools/plotting.py
```diff
@@ -2,9 +2,9 @@
 
 
 def degradation_summary_plots(yoy_rd, yoy_ci, yoy_info, normalized_yield,
-                              hist_xmin=None, hist_xmax=None, scatter_ymin=None,
-                              scatter_ymax=None, plot_color=None, summary_title=None,
-                              scatter_alpha=0.5):
+                              hist_xmin=None, hist_xmax=None,  bins=75, 
+                              scatter_ymin=None, scatter_ymax=None, plot_color=None, 
+                              summary_title=None, scatter_alpha=0.5):
```
mdeceglie: I'm generally in favor with adding this control of the bin count. This does also change the default behavior from `yoy_values//40` to 75. I want to make sure everyone agrees. @dirkjordan ? 

If not we can make the default `None` with a conditional that sets bins to `yoy_values//40`

mdeceglie:
Let's revert the plotting changes, they are handled in #138 , and merge just the bug fix

kanderso-nrel:
@mdeceglie @cdeline PTAL

cdeline:
OK, so this removes all of the plotting changes and leaves the one-line bug fix to set `pv_nameplate = self.pv_nameplate` which is a pretty trivial change.  I'd review and approve this myself, but I can't.  :)

PR #134, opened 2020-01-30
Add Changelog to sphinx docs

kanderso-nrel:
I think it makes sense to merge this one first so that the other 2.0 PRs can incrementally add their own changes.  Built docs for this branch are here: https://rdtools.readthedocs.io/en/changelog/changelog.html


mdeceglie:
It looks like  a python 2.7 build is failing. Should we target merging #135 as our first step?

mdeceglie:
I suggest adding a section now for API changes.

kanderso-nrel:
I think this and #135 should be merged before others, but no preference on which of the two goes first -- a failing test doesn't prevent us from merging.  I just added the API changes section.  For future changelogs we might not worry about having all the sections included at the start, but I think it's a good idea here to hopefully minimize merge conflicts.  Are we good to merge this now?

PR #135, opened 2020-01-30
Drop py2.7 and add 3.7 and 3.8 to testing

kanderso-nrel:
Testing out changes to the travis config

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,9 @@ sudo: false
 language: python
 
 python:
-  - "2.7"
   - "3.6"
+  - "3.7"
+  - "3.8"
 
 # Test two environments:
 # 1) dependencies with pinned versions from requirements.txt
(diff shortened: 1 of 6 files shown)


kanderso-nrel:
@mdeceglie @cdeline -- test failures are because requirements.txt specifies `numpy==1.13.3` and pypi doesn't have `numpy==1.13.3` wheels for 3.7 and 3.8 because they didn't exist when 1.13.3 was released.  

[NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html) recommends numpy 1.15 as the minimum numpy version for this point, which does have 3.7 wheels on pypi.  The oldest numpy release with 3.8 wheels is 1.17.3.  What do you think about bumping up the numpy version in requirements.txt to 1.17.3 so that we can test on 3.7 and 3.8?  Or bumping to 1.15 and not testing on 3.8?  

Also pandas 1.0.0 just got released a few hours ago and broke `get_clearsky_tamb`, that's why I've excluded it here.

mdeceglie:
I have no objection to increasing the numpy minimum version.

kanderso-nrel:
Tests pass on 3.8 now with updated requirements.txt entries for other packages as well:
- numpy==1.13.3 -> 1.17.3
- pandas==0.23.0 -> 0.25.2
- scipy==1.2.2 -> 1.3.2
- statsmodels==0.10.1 -> 0.10.2

PR #136, opened 2020-01-31
Add developer notes section to sphinx docs

kanderso-nrel:
This PR:
- Adds a sphinx page with instructions on running the test suite and building docs locally
- Deletes the sphinx_requirements.txt file and replaces it with an `extras_require` section called `doc` in setup.py.  This lets you do `pip install rdtools[doc]` to install both the normal rdtools requirements and the doc requirements in one go. 

Any objections to the second bullet?  I think specifying requirements in setup.py is generally preferred over requirements.txt files.  Suggestions for the "developer notes" page are welcome as well of course.

Built version here: https://rdtools.readthedocs.io/en/developer_notes/developer_notes.html

### DIFF
--- a/docs/sphinx/source/developer_notes.rst
+++ b/docs/sphinx/source/developer_notes.rst
@@ -100,6 +100,21 @@ And even a single test function:
 
     python -m pytest rdtools/test/soiling_test.py::test_soiling_srr
 
+You can also evaluate code coverage when running the test suite using the 
+`coverage <https://coverage.readthedocs.io>_` package:
+
+::
+
+    coverage run -m pytest
+    coverage report
+
+The first line runs the test suite and keeps track of exactly what lines of
+code were run during test execution.  The second line then prints out a
+summary report showing how much much of each source file was
+executed in the test suite.  If a percentage is below 100, that means a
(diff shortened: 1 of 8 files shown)


docs/sphinx/source/developer_notes.rst
```diff
@@ -21,15 +21,65 @@ you'll need to clone the RdTools source repository from Github with e.g.
...
+This will install the latest official release of RdTools.  If you want to work
+with a development version and you have cloned the Github repository to your
+computer, you can also install RdTools and dependencies by navigating to the
+repository root and running:
+
+::
+
+    pip install .
+
+This will install the development version of RdTools along with the current
```
mdeceglie: Technically it will install whatever branch you are on, which may not be development

docs/sphinx/source/developer_notes.rst
```diff
@@ -50,29 +100,28 @@ And even a single test function:
 
     python -m pytest rdtools/test/soiling_test.py::test_soiling_srr
```
mdeceglie: I believe you can just run the tests with:
`pytest`

PR #139, opened 2020-02-03
Normalized filter cherrypick

kanderso-nrel:
This cherry-picks out the normalized_filter function present in 2.0.0-alpha, cleans up the docstring, and incorporates into the sphinx API reference.


kanderso-nrel:
Noticing that many of the existing filters are basically the same function, it might make sense to consolidate them into generic filtering functions.  The idea of separating the "best-practice" analysis from the low-level functions would support that idea because the current default arguments are really part of the best-practice methodology, so removing them from the functional interface is consistent with that separation.

kanderso-nrel:
Let's keep the functions as-are for now, can always make them more general later.  Need to update the lower default limit to 0.01 from 0.0.

kanderso-nrel:
Updated lower bound to 0.01.  Also changed the notebook filter variables from `nz_mask` to `normalized_mask`, and the same for the clear-sky versions.

rdtools/test/filtering_test.py
```diff
@@ -73,5 +73,11 @@ def test_clip_filter_upper(self):
         self.assertTrue((expected_result == filtered).all())
 
 
+def test_normalized_filter_default():
```
mdeceglie, line 76: Suggest updating this test or adding another that would fail if we change the default behavior. Perhaps something like: 
`pd.testing.assert_series_equal(normalized_filter(pd.Series([0.01 - 1e-16, 0.01 + 1e-16, 1e308])), pd.Series([False, True, True]))`

kanderso-nrel: Good idea, done.

PR #140, opened 2020-02-03
Move documentation in README.md to index.rst in sphinx docs

kanderso-nrel:
Mike and I talked about moving the library usage documentation in README.md to RTD and making README.md more geared towards library development. 

- [x] Need to remember to remove `m2r` from doc requirements once #136 is (or isn't) merged.


docs/sphinx/source/developer_notes.rst
```diff
@@ -185,8 +185,11 @@ If you get an error like ``Pandoc wasn't found``, you can install it with conda:
 The built documentation should be in ``rdtools/docs/sphinx/build`` and opening
 ``index.html`` with a web browser will display it.
 
-Code requirements
------------------
+Contributing
+------------
+
+Community participation is welcome!  New contributions should be based on the
+``development`` branch as the ``master`` branch is used only for releases.
```
kanderso-nrel, line 192: Should we take this out?

setup.py
```diff
@@ -46,7 +48,6 @@
 EXTRAS_REQUIRE = {
```
mdeceglie, line 48: My doc building exercise failed without `m2r`. Also had to install `pandoc` (I had to install pandoc with conda rather than pip for some reason in my virtual environment...)

docs/sphinx/source/developer_notes.rst
mdeceglie: I thought our discussion with @cdeline ended with "let's keep the development branch workflow, but be better about smaller more frequent releases"

mdeceglie: If I have that right, we can leave it for now

setup.py
mdeceglie: Now I see we already have a note about pandoc. But should it also be in setup.py? Or is it technically a sub-requirement?

docs/sphinx/source/index.rst
```diff
@@ -6,8 +6,227 @@
...
+
+Two workflows are available for system performance ratio calculation,
+and illustrated in an example notebook. The sensor-based approach
+assumes that site irradiance and temperature sensors are calibrated and
+in good repair. Since this is not always the case, a 'clear-sky'
+workflow is provided that is based on modeled temperature and
+irradiance. Note that site irradiance data is still required to identify
+clear-sky conditions to be analyzed. In many cases, the 'clear-sky'
+analysis can identify conditions of instrument errors or irradiance
+sensor drift, such as in the above analysis.
```
mdeceglie, line 67: Suggest adding:

> The clear-sky analysis tends to provide less stable results than sensor-based analysis when details such as filtering are changed. We generally recommend that the clear-sky analysis be used as a check on the sensor based results, rather than as a stand-alone analysis.

setup.py
kanderso-nrel: Ah, I removed `m2r` from `setup.py` but not `conf.py`.  I'll fix that, good catch.

pandoc is used internally by `nbsphinx` for the notebook to rst conversion, so technically it's an indirect requirement, but it's on us to provide it to nbsphinx so it's not really black and white.  Is it even possible to include conda packages in setup.py?

mdeceglie: pandoc is on pypi, but for whatever reason installing it with pip in the virtual environment doesn't allow me to build the docs. I vote we continue to leave it out of setup.py since it is indirect.

docs/sphinx/source/index.rst
```diff
@@ -66,6 +66,11 @@ clear-sky conditions to be analyzed. In many cases, the 'clear-sky'
 analysis can identify conditions of instrument errors or irradiance
 sensor drift, such as in the above analysis.
 
+The clear-sky analysis tends to provide less stable results than sensor-based
+analysis when details such as filtering are changed. We generally recommend
+that the clear-sky analysis be used as a check on the sensor based results,
```
mdeceglie, line 71: sensor-based needs a hyphen here

docs/sphinx/source/developer_notes.rst
cdeline: Yeah, I'm fine with keeping this.  Doing merges from development during releases is handy.

PR #142, opened 2020-02-05
Workaround for reindexing bug in Pandas 1.0.0

kanderso-nrel:
Calling `df.reindex(times, method='nearest')` with tz-aware indexes raises an error in `pandas 1.0.0` (see https://github.com/pandas-dev/pandas/issues/26683 and https://github.com/pandas-dev/pandas/pull/31511).  This works around it by dropping and restoring the timezones so that the reindex is performed on tz-naive indexes.  The intermediate UTC conversion is to avoid DST issues when relocalizing.

FYI after making this change, tests all pass for me with pandas 1.0.0 installed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,7 @@ TESTS_REQUIRE = [
 
 INSTALL_REQUIRES = [
     'numpy >= 1.12',
-    'pandas >= 0.23.0, <1.0.0, >=1.0.2',  # exclude 1.0.0 & 1.0.1 for GH142
+    'pandas >= 0.23.0,!=1.0.0,!=1.0.1',  # exclude 1.0.0 & 1.0.1 for GH142
     'statsmodels >= 0.8.0',
     'scipy >= 0.19.1',
     'h5py >= 2.7.1',
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,7 @@ TESTS_REQUIRE = [
 
 INSTALL_REQUIRES = [
     'numpy >= 1.12',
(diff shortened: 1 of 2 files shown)


cdeline:
Increased pandas version in setup.py to include Pandas 1.0.0 in our Travis pytests when upgrade-strategy=eager

kanderso-nrel:
FYI looks like the fix for this will be in pandas `1.0.2`.  It isn't released yet, but we could omit 1.0.0 and 1.0.1 from our allowed versions when it comes out.

setup.py
```diff
@@ -35,7 +35,7 @@
 
 INSTALL_REQUIRES = [
     'numpy >= 1.12',
-    'pandas >= 0.23.0, <1.0.0',
+    'pandas >= 0.23.0, <=1.0.0',
```
kanderso-nrel:
```suggestion
    'pandas >= 0.23.0, <1.0.0, >=1.0.2',  # exclude 1.0.0 & 1.0.1 for GH142
```
Pandas 1.0.2 was released yesterday with the fix for this issue.  Should we forget the workaround and just exclude 1.0.0 and 1.0.1 like this?  I'd guess that people will move on from 1.0.0 and 1.0.1 pretty quickly, so IMHO it's not worth maintaining weird code to support those versions.

mdeceglie: I agree and support requiring 1.0.2

cdeline:
I concur.  No need to support 1.0.0.

Issue #84, opened 2018-07-19
filtering.clip_filter() also filters low power


mdeceglie:
`clip_filter()` filters clipping, but also low powers, which aren't really related to clipping. The later functionality isn't really conveyed by the function name. This functionality should be implemented in a separate low power filter function.

PR #144, opened 2020-02-06
Remove low power cutoff in filtering.clipping_filter

kanderso-nrel:
See #84


kanderso-nrel:
Technically no, there isn't an exact replacement elsewhere.  v2.0.0-alpha.0's system_analysis class filters out normalized_power < 0.0 with the normalized filter and follows it with a poa < 200 filter: https://github.com/NREL/rdtools/blob/2.0.0-alpha.0/rdtools/system_analysis.py#L251

So I think it's unlikely that removing the clipping low_cutoff will make a practical difference.  I'd be in favor of increasing the lower limit on the the normalized filter to some epsilon > 0 to filter out very low values to keep with the spirit of this filter, though.

kanderso-nrel:
Thanks for the reminder.  For completeness, here's the changes I found:

Sensor-based:
- Degradation ci:  (-0.52, -0.33) -> (-0.52, -0.35)
- degradation P95 exceedance level: -0.58%/yr -> -0.57%/yr
- P50 insolation-weighted soiling ratio: 0.973 -> 0.972
- insolation-weighted soiling ratio ci: 0.965–0.978 -> 0.965–0.979

Clearsky-based:
- degradation P95 exceedance level: -0.35%/yr -> -0.34%/yr

Since the notebook is acting as a de-facto system test, it might be worth making a pytest file that at least does the easy checks in the notebook -- degradation/soiling rates, CIs, etc.  Separate issue.

kanderso-nrel:
Actually, those differences were comparing against an older version of the notebook.  Comparing this PR against the current notebook in the development branch here: https://github.com/NREL/rdtools/blob/development/docs/degradation_and_soiling_example.ipynb

Sensor-based:
- degradation P95 exceedance level: -0.58%/yr -> -0.57%/yr
- degradation CI: (-0.55, -0.34) -> (-0.52, -0.35)
- soiling ratio CI: 0.970–0.981 -> 0.964–0.978

Clearsky-based:
- Rd: -0.10%/yr -> -0.09%/yr
- degradation P95 exceedance level: -0.40%/yr -> -0.35%/yr
- degradation CI: (-0.29, 0.08) -> (-0.25, 0.06)

kanderso-nrel:
With the normalized > 0.01 change, here are the updated notebook differences from `development`:

Sensor-based:
- degradation CI: (-0.55, -0.34) -> (-0.52, -0.33)
- soiling ratio CI: (0.970, 0.981) -> (0.965, 0.979)

Clearsky-based:
- Rd: -0.10%/yr -> -0.09%/yr
- degradation P95 exceedance level: -0.40%/yr -> -0.35%/yr
- degradation CI: (-0.29, 0.08) -> (-0.25, 0.06)

mdeceglie:
I'm seeing almost the same changes to the results in #164 which tells me the things causing the change are lurking in `development` and we just didn't update the notebook.

PR #150, opened 2020-03-04
Implementing the CODS algorithm in RdTools development

asmunds:
This is an initial attempt at a pull request for the CODS algorithm.

Most of the relevant functions and classes are placed in soiling.py. I have added a class called `CODSAnalysis`, in which the core functionality live. Some supporting functions (with routines that are not specific to CODS) are placed outside of this class (still in soiling.py).

In degradation.py, I have added the `naive_YOY()`-function, which does the same as `degradation_year_on_year()`, except for the Monte Carlo procedure. This is done because CODS has its own internal method for finding confidence intervals through bootstrapping, and skipping the Monte Carlo procedure in the YOY-method speeds it up quite a bit.

There are a few new package requirements and updated package versions.

Please let me know if there is anything I could structure differently, or if anything else is needed. I could for instance write an example notebook, if it is desired. 
I don't know how to write tests, so I have not added that.


asmunds:
I have now updated the PR with the latest version of the development branch, and added tests and a jupyter demo script.

I don't know why Travis CI fails on filterpy. @kanderso-nrel or @mdeceglie, do you have any ideas on this?

asmunds:
Okay, including filterpy in the setup-file did it. Is there anything else I could do to make this PR more likely to be adopted?

asmunds:
I finally found time to add the time series bootstrap module (as we have discussed @mdeceglie), which I have put in the new file `bootstrap.py`. I have added a function in `degradation.py` (called `bootstrap_YOY`) that calls `degradation_year_on_year` and constructs confidence intervals based on the bootstrap module.

I have also removed naive_YOY and replaced it by a boolean parameter in `degradation_year_on_year` (called `full_calc`) which determines whether or not confidence intervals should be constructed. Feel free to suggest other ways of structuring this.

kanderso-nrel:
I had a quick look just at the documentation; here are some suggestions:

- Functions can be cross-linked in the built documentation on ReadTheDocs by using ```:py:func:`func_name` ``` (if the function is in the same module) or ``` :py:func:`rdtools.module_name.func_name` ``` for functions in other modules.  Mild preference for this over just using ``` `func_name` ``` in docstrings. 
- For parameter and return types -- `pd.Series` is consistent with the other docstrings currently on development, but #258 will require `pandas.Series` (and the same elsewhere, e.g. `numpy.array` instead of `np.array`)
- Rather than embedding citation in text, I suggest extracting them out to `Reference` sections.  Also, add an underscore after the brackets to get sphinx to recognize it as a link to the references (i.e., change `[1]` to `[1]_`).  Also reference entries need a leading `.. `, e.g.: https://github.com/NREL/rdtools/blob/master/rdtools/availability.py#L191
- We make some effort to comply with pv-terms (#175), so `energy_normalized_daily` is preferred over `daily_energy_normalized`
- If it's not a big hassle, it's better to use immutable types for default parameter values (e.g. tuple instead of list) to avoid any chance of running into this issue: https://web.archive.org/web/20200221224620/http://effbot.org/zone/default-values.htm
- For docstring equations (like the one in iterative_signal_decomposition), consider using RST to improve its formatting -- maybe use `.. math::`? https://docutils.sourceforge.io/docs/ref/rst/directives.html#math
- For functions with complex `Returns` sections, I've always found sphinx's handling of nested lists (like documenting the columns in a dataframe) awkward at best.  Check out https://rdtools--150.org.readthedocs.build/en/150/generated/rdtools.soiling.CODSAnalysis.iterative_signal_decomposition.html as an example.  The issue might be having a non-bulleted line before the bullets -- here's an example that renders nicely: https://github.com/NREL/rdtools/blob/master/rdtools/soiling.py#L547
- I'm not sure about the `knob_alternatives` parameter... it seems awkward to use as someone unfamiliar with this code.  Would it make sense to break it apart into individual parameters?
- I'm not sure we've followed this rule to a T, but: functions and methods should either (1) have a full docstring (with complete Parameters sections and such) and be listed in `api.rst`, or (2) be made "private" by putting an underscore at the beginning of the function name.  I'm looking at the `set_control_input`, `forward_pass` etc methods, and the new top-level functions in `soiling.py`.
- You can add the new jupyter notebook to the example gallery -- here are some notes about how to do that: https://github.com/NREL/rdtools/wiki/Example-gallery-notes

Type annotations are interesting, and not something used in RdTools before.  I'm interested in what advantages type annotations have over normal docstring types (e.g. can sphinx use them?  should we start using mypy?).

mdeceglie:
@asmunds It's awesome to see this moving along! I tried running CODS on our PVDAQ-4 example which is now our preferred example and I am running into `Soiling signal is small relative to the noise. Iterative decomposition not possible. Degradation found by RdTools YoY` This is one where we have imposed a synthetic soiling signal, and to the eye, the soiling signal is apparent. Any ideas why CODS isn't working on this example?

To reproduce, you just need to run the following at the end of `degradation_and_soiling_example_pvdaq_4.ipynb` 
```
from rdtools.soiling import CODSAnalysis
CODS = CODSAnalysis(daily)
CODS.run_bootstrap(reps=16, verbose=True)
```

rdtools/soiling.py
```diff
@@ -999,3 +1012,1287 @@ def monthly_soiling_rates(soiling_interval_summary, min_interval_length=14,
...
+            if verbose:
+                print('Soiling signal is small relative to the noise')
+            small_soiling_signal = True
+            df_out.SR_high = 1.0
+            df_out.SR_low = 1.0 - SR_amp
+
+        return df_out, degradation, soiling_loss, residual_shift, RMSE, \
+            small_soiling_signal, adf_res
+
+    def run_bootstrap(self, reps=512, verbose=False,
```
mdeceglie, line 1355: It would be great to give `run_bootstap()` a `confidence_level` parameter so users can adjust the confidence level used for confidence intervals, similar to `degradation_year_on_year()` and `bootrap_YOY()`

rdtools/bootstrap.py
```diff
@@ -0,0 +1,106 @@
...
+    Generate bootstrap samples based a time series signal and its model fit
+    using circular block bootstrapping.
+
+    Parameters
+    ----------
+    signal : pd.Series
+        The time series signal that you want to make bootstrap samples of
+    model_fit : pd.Series
+        A model fit to the signal
+    sample_nr : int, default 10
```
mdeceglie, line 21:
```suggestion
    sample_nr : int, default 1000
```

asmunds: Thanks, well spotted

rdtools/soiling.py
asmunds: Good idea. I've included this in the coming commits.

asmunds:
> 
> 
> @asmunds It's awesome to see this moving along! I tried running CODS on our PVDAQ-4 example which is now our preferred example and I am running into `Soiling signal is small relative to the noise. Iterative decomposition not possible. Degradation found by RdTools YoY` This is one where we have imposed a synthetic soiling signal, and to the eye, the soiling signal is apparent. Any ideas why CODS isn't working on this example?
> 
> To reproduce, you just need to run the following at the end of `degradation_and_soiling_example_pvdaq_4.ipynb`
> 
> ```
> from rdtools.soiling import CODSAnalysis
> CODS = CODSAnalysis(daily)
> CODS.run_bootstrap(reps=16, verbose=True)
> ```

Unfortunately, CODS does not handle that amount of noise well. I usually run a small outlier filter before I run the CODS algorithm on field-data.

asmunds:
Okay @kanderso-nrel and @mdeceglie, thanks for great feedback. I have pushed a few changes now.

@kanderso-nrel, thanks for the comments on the documentation. I agree with all your suggestions, and have tried to comply with them (except adding the example notebook to the gallery. I'm planning to do it, though). It doesn't seem as if sphinx can handle type annotations, at least not out-of-the-box, so I decided to drop them.

Here are some thoughts I have that I would love some feedback on:
* I have put the bootstrap module in api.rst. I am not sure if we would want these functions to be visible to everyone.
* I am not sure how to document class attributes that are not initialized in `__init__`. At this point they're just listed under the header "Attributes", but it does not compile nicely in sphinx. I see that the same thing happens in `availability.AvailabilityAnalysis`, so I guess this is something you guys will have to figure out anyway.
* The test-module for the CODS-methods might be lacking. I have written tests for the top-level methods, but not for all the low-level ones. I am planning to do that too, but I am not sure how to best deal with coverage, as the top-level methods depend on the lower-level ones, and we don't want to much redundancy.
* As is apparent in the two runs of the example notebooks that you have made, @mdeceglie, the CODS algorithm is not as robust to noise as I would hope. That is, if there is too much noise in the data, it concludes too easily that the soiling signal is insignificant. The sensitivity to noise is something that can be tuned, and we should maybe look into this in the future. I've already made a slight change (Comparing the 80 percentile span of the soiling signal vs. the noise, instead of the 95 percentile span). However, the performance of the model is better (in the sense that the model fits the data better, judging by the eye) if we just add an outlier filter before running CODS. I've added an outlier filter in the example notebook. 
* I've added a run of the CODS algorithm in `degradation_and_soiling_example_pvdaq_4.ipynb`. I assume we will remove this in the future, but I put it there to make it easier for you to run the example and see the effect of the outlier filter.

All in all, this is still work in progress. For instance, I have not added the CODS example notebook to the example gallery yet. But at least there has been some progress, and I hope we are getting closer to merging. Thanks again for the feedback.

rdtools/test/soiling_cods_test.py
```diff
@@ -92,13 +92,13 @@ def test_soiling_cods(cods_normalized_daily):
...
+    assert 0.962207 == pytest.approx(sr, abs=0.5),\
         'Soiling ratio different from expected value'
-    assert np.array([0.96662419, 0.95692131]) == pytest.approx(sr_ci, abs=1e-1),\
+    assert np.array([0.96662419, 0.95692131]) == pytest.approx(sr_ci, abs=0.5),\
         'Confidence interval of SR different from expected value'
-    assert 0.09 == pytest.approx(deg, abs=0.2),\
+    assert 0.09 == pytest.approx(deg, abs=0.5),\
         'Degradation rate different from expected value'
-    assert np.array([-0.17143952,  0.39313724]) == pytest.approx(deg_ci, abs=0.2),\
+    assert np.array([-0.17143952,  0.39313724]) == pytest.approx(deg_ci, abs=0.5),\
```
asmunds, line 101: I had to relax the tolerance of the numerical checks in this test to make it pass in Python 3.8. It seems there is still some randomness involved, even if I seed numpy.random. In itself, not getting the same number every time is not a problem. But if this means we are unable to test the integrity of the algorithm, it might be a problem. On the other hand, `iterative_signal_decomposition` is not stochastic in nature, and we are able to test that method with a high accuracy. If `iterative_signal_decomposition` works, `soiling_cods`should work as well.

mdeceglie:
It's great to have that boostrap capability available for the yoy analysis. I'd suggest incorporating into the base `bootstrap_YOY` function via a key word rather than keeping it as a separate function. We should retain the current default behavior. This would go something like this:
```
def degradation_year_on_year(energy_normalized, recenter=True,
                             exceedance_prob=95, confidence_level=68.2,
                             uncertainty_method='simple_bootstrap', block_length=30)
```
But with another available value `uncertainty_method = 'circular_block_bootstrap'` I haven't fully thought through whether we can get rid of the `full_calc` parameter. Is it used anywhere other than the existing `bootstrap_YOY()`?

asmunds:
I think that is a good idea. I will try to merge the two functions as you suggest.

Its important to have the possibility of not doing bootstrapping. It saves quite a bit of time, which makes a big difference in the bootstrap phase of CODS. However, we could trigger that functionality through `uncertainty_method` keyword, for instance by typing None or 'none'.

asmunds:
Okay, so the current version has `uncertainty_method` as a keyword in `degradation_year_on_year`, and the cods_example-notebook is included in the example gallery.

asmunds:
Is it possible to require scipy>=1?

kanderso-nrel:
> Is it possible to require scipy>=1?

OK with me.  pvlib has required `scipy>=1.2.0` since v0.8.0, so `scipy>=1.0` is already an indirect requirement for anyone using a recent version of pvlib.  It's also within the guidelines of NEP-29, which I think is a good rule of thumb even if we don't officially follow it.

mdeceglie:
I'm not able to reproduce the python 3.6 requirements-min failure locally (it fails on my machine for a different reason...). Does anyone have insights into what is causing the failure, or perhaps we should we just update the minimum requirements? Perhaps push through a scaled back #302 to stop testing on 3.6 as well?

kanderso-nrel:
I can reproduce them locally. Three of the failures are because of the index inequality check on this line: https://github.com/asmunds/rdtools/blob/69b3d57bbf2eeeb5fb51cf3eee379ab3ecfab6ad/rdtools/soiling.py#L1998

I'm having trouble understanding the intent of that check, so I'm not sure what to suggest.  In at least one of the test functions, `prescient_cleaning_events.index` is a DatetimeIndex while `zs_series.index` is a RangeIndex, maybe that's the issue?  Otherwise I don't think it makes sense to compare them...

The last one is because of `to_list` here:  https://github.com/asmunds/rdtools/blob/69b3d57bbf2eeeb5fb51cf3eee379ab3ecfab6ad/rdtools/soiling.py#L2007

Looks like `to_list` was only added in [0.24.0](https://pandas.pydata.org/docs/whatsnew/v0.24.0.html?highlight=to_list#other-api-changes).  I think changing that to `tolist` will get it working, but I did not confirm it myself.

asmunds:
> I'm having trouble understanding the intent of that check, so I'm not sure what to suggest.

The intent of the check is to make sure that the two time series that are input to the function (zs_series and prescient_cleaning_events) have the same indices. If the indices are different in the test function, that's a problem with the test function. I'll look into it.

rdtools/soiling.py
```diff
@@ -1052,3 +1065,1532 @@ def monthly_soiling_rates(soiling_interval_summary, min_interval_length=14,
     monthly_soiling_df['interval_count'] = relevant_interval_count
 
     return monthly_soiling_df
+
+
+class CODSAnalysis():
+    '''
+    Container for the Combined Degradation and Soling (CODS) algorithm
```
kanderso-nrel, line 1072:
```suggestion
    Container for the Combined Degradation and Soiling (CODS) algorithm
```

rdtools/soiling.py
```diff
@@ -1052,3 +1065,1532 @@ def monthly_soiling_rates(soiling_interval_summary, min_interval_length=14,
...
+        +------------------------+----------------------------------------------+
+        | 'residuals'            | The residuals of the model fit, i.e.         |
+        |                        | PI / (SR * SC * Rd)                          |
+        +------------------------+----------------------------------------------+
+        | 'SR_low'               | lower bound of 95 % conf. interval of SR     |
+        +------------------------+----------------------------------------------+
+        | 'SR_high'              | upper bound of 95 % conf. interval of SR     |
+        +------------------------+----------------------------------------------+
+        | 'rates_low'            | lower bound of 95 % conf. interval of SR     |
+        |                        | soiling rates                                |
```
kanderso-nrel, line 1118:
```suggestion
        | 'rates_low'            | lower bound of 95 % conf. interval of        |
        |                        | soiling rates                                |
```

rdtools/soiling.py
```diff
@@ -1052,3 +1065,1532 @@ def monthly_soiling_rates(soiling_interval_summary, min_interval_length=14,
...
+
+    def iterative_signal_decomposition(
+            self, order=('SR', 'SC', 'Rd'), degradation_method='YoY',
+            max_iterations=18, detection_tuner=.5, convergence_criterium=5e-3,
+            pruning_iterations=1, pruning_tuner=.6, soiling_significance_knob=.75,
+            process_noise=1e-4, renormalize_SR=None, ffill=True, clip_soiling=True,
+            verbose=False):
+        '''
+        Description
+        -----------
```
kanderso-nrel, line 1187: Let's remove this header, it doesn't get displayed in the rendered sphinx output and it messes up the single-line description of this function

rdtools/soiling.py
```diff
@@ -1052,3 +1065,1532 @@ def monthly_soiling_rates(soiling_interval_summary, min_interval_length=14,
...
+        consists of a degradation trend, a seasonal component, and a soiling
+        signal (defined as 1 if no soiling, decreasing with increasing soiling
+        losses). I.e.: PI = degradation_trend * seasonal_component * soiling_ratio *
+        residuals, or:
+
+        .. math::
+
+            PI = Rd * SC * SR * R
+
+        The function has a huristic for detecting whether the soiling signal is
```
kanderso-nrel, line 1200:
```suggestion
        The function has a heuristic for detecting whether the soiling signal is
```

rdtools/soiling.py
```diff
@@ -1052,3 +1065,1532 @@ def monthly_soiling_rates(soiling_interval_summary, min_interval_length=14,
...
+
+        * The degradation trend is obtained using the native RdTools Year-On-Year
+          method [1]_.
+        * The seasonal component is derived with statsmodels STL [2]_.
+        * The soiling signal is derived with a Kalman Filter with a cleaning
+          detection heuristic [3]_.
+
+        Parameters
+        ----------
+        order : tuple, defualt ('SR', 'SC', 'Rd')
```
kanderso-nrel, line 1213:
```suggestion
        order : tuple, default ('SR', 'SC', 'Rd')
```

rdtools/soiling.py
```diff
@@ -1052,3 +1065,1532 @@ def monthly_soiling_rates(soiling_interval_summary, min_interval_length=14,
...
+            components, so three iterations are needed to fit all three components.
+        detection_tuner : float, default .5
+            Higher value gives lower cleaning event detection sensitivity.
+            Should be between 0.1 and 2
+        convergence_criterium : float, default 5e-3
+            the relative change in the convergence metric required for
+            convergence
+        pruning_iterations : int, default 1
+        pruning_tuner : float, default .6
+            Should be between 0.1 and 2
```
kanderso-nrel, line 1234: Would be good to explain a bit what these two `pruning` parameters are for.  Same for `process_noise` below.

rdtools/soiling.py
```diff
@@ -1052,3 +1065,1532 @@ def monthly_soiling_rates(soiling_interval_summary, min_interval_length=14,
...
+        ffill : bool, default True
+            Whether to use forward fill (default) or backward fill before
+            doing the rolling median for cleaning event detection
+        clip_soiling : bool, default True
+            Whether or not to clip the soiling ratio at max 1 and minimum 0.
+        verbose : bool, default False
+            If true, prints a progress report
+        ...
+
+        Returns
```
kanderso-nrel, line 1249: Seven return values is quite a lot.  We should think about potential alternatives, for example a dictionary or some kind of `Result` object?

rdtools/soiling.py
```diff
@@ -1052,3 +1065,1532 @@ def monthly_soiling_rates(soiling_interval_summary, min_interval_length=14,
...
+                      degradation_method='YoY',
+                      process_noise=1e-4,
+                      order_alternatives=(('SR', 'SC', 'Rd'),
+                                          ('SC', 'SR', 'Rd')),
+                      detection_tuner_alternatives=(.25, .75),
+                      pruning_tuner_alternatives=(1/1.5, 1.5),
+                      forward_fill_alternatives=(True, False),
+                      **kwargs):
+        '''
+        Bootstrapping of CODS alforithm for uncertainty analysis, inherently accounting
```
kanderso-nrel, line 1522:
```suggestion
        Bootstrapping of CODS algorithm for uncertainty analysis, inherently accounting
```

rdtools/soiling.py
```diff
@@ -1052,3 +1065,1532 @@ def monthly_soiling_rates(soiling_interval_summary, min_interval_length=14,
...
+            The size of the confidence intervals to return, in percent
+        verbose : bool, default False
+            Wheter or not to print information about progress
+        degradation_method : string, default 'YoY'
+            Either 'YoY' or 'STL'. If anything else, 'YoY' will be assumed.
+            Decides whether to use the YoY method [3] for estimating the
+            degradation trend (assumes linear trend), or the STL-method (does
+            not assume linear trend). The latter is slower.
+        order_alternatives : tuple of tuples, default (('SR', 'SC', 'Rd'), \
+            ('SC', 'SR', 'Rd'))
```
kanderso-nrel, line 1555: This renders strangely -- I think this would fit all on one line without angering the linter (see https://github.com/NREL/rdtools/blob/master/.flake8), so probably that's the easiest solution.

rdtools/soiling.py
```diff
@@ -1052,3 +1065,1532 @@ def monthly_soiling_rates(soiling_interval_summary, min_interval_length=14,
...
+    def run_bootstrap(self,
+                      reps=512,
+                      confidence_level=68.2,
+                      verbose=False,
+                      degradation_method='YoY',
+                      process_noise=1e-4,
+                      order_alternatives=(('SR', 'SC', 'Rd'),
+                                          ('SC', 'SR', 'Rd')),
+                      detection_tuner_alternatives=(.25, .75),
+                      pruning_tuner_alternatives=(1/1.5, 1.5),
```
kanderso-nrel, line 1518: These two sets of `alternatives` are different from the defaults listed in the docstring below

rdtools/soiling.py
```diff
@@ -1052,3 +1065,1532 @@ def monthly_soiling_rates(soiling_interval_summary, min_interval_length=14,
...
+            df_out.SR_high = 1.0
+            df_out.SR_low = 1.0 - SR_amp
+
+        return df_out, degradation, soiling_loss, residual_shift, RMSE, \
+            small_soiling_signal, adf_res
+
+    def run_bootstrap(self,
+                      reps=512,
+                      confidence_level=68.2,
+                      verbose=False,
```
kanderso-nrel, line 1512: `iterative_signal_decomposition` has `verbose` as the last parameter, which I think would make sense to do here too.

rdtools/soiling.py
```diff
@@ -1052,3 +1065,1532 @@ def monthly_soiling_rates(soiling_interval_summary, min_interval_length=14,
...
+            +------------------------+----------------------------------------------+
+            | 'residuals'            | The residuals of the model fit, i.e.         |
+            |                        | PI / (SR * SC * Rd)                          |
+            +------------------------+----------------------------------------------+
+            | 'SR_low'               | lower bound of 95 % conf. interval of SR     |
+            +------------------------+----------------------------------------------+
+            | 'SR_high'              | upper bound of 95 % conf. interval of SR     |
+            +------------------------+----------------------------------------------+
+            | 'rates_low'            | lower bound of 95 % conf. interval of SR     |
+            |                        | soiling rates                                |
```
kanderso-nrel, line 1606:
```suggestion
            | 'rates_low'            | lower bound of 95 % conf. interval of        |
            |                        | soiling rates                                |
```

rdtools/soiling.py
```diff
@@ -1052,3 +1065,1532 @@ def monthly_soiling_rates(soiling_interval_summary, min_interval_length=14,
...
+                        [0, rate_std**2]])
+        f.Q = Q_discrete_white_noise(dim=2, dt=dt, var=process_noise**2)
+        f.x = np.array([initial_slope[1], initial_slope[0]])
+        f.B = np.zeros(f.F.shape)
+        f.B[0] = 1
+        f.R = measurement_noise
+        return f
+
+
+def soiling_cods(energy_normalized_daily,
```
kanderso-nrel, line 2248: Same comments here:
- consider moving `verbose` to the end of the parameter list
- `alternatives` defaults don't match docstring defaults
- `order_alternatives` renders incorrectly, so condense to a single line
- delete incorrect `SR` in the `rates_low` table entry

kanderso-nrel (commented):
Here's a very nitpicky review focused on the documentation.  I haven't yet looked at the code or tests :)

docs/sphinx/source/api.rst
```diff
@@ -165,4 +170,15 @@ Plotting
    soiling_interval_plot
    soiling_rate_histogram
    availability_summary_plots
-   tune_filter_plot
\ No newline at end of file
+   tune_filter_plot
+
+
+Bootstrap
```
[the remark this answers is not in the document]
asmunds: You're right, no need to include them there

rdtools/soiling.py
```diff
@@ -1052,3 +1065,1530 @@ def monthly_soiling_rates(soiling_interval_summary, min_interval_length=14,
...
+                           columns=range(int(sample_nr*len(list_of_SCs))))
+    # From each fitted signal, we will generate new seaonal components
+    for i, signal in enumerate(list_of_SCs):
+        # Remove beginning and end of signal
+        signal_mean = signal.mean()
+        unique_years = signal.index.year.unique()  # Unique years
+        # Make a signal matrix where each column is a year and each row a date
+        year_matrix = pd.concat([pd.Series(signal.loc[str(year)].values)
+                                 for year in unique_years],
+                                axis=1, ignore_index=True)
```
kanderso-nrel, line 2491: I think there is a bug here when the input series is not a sequence of complete calendar years.  For example if the first year starts on Dec 1, `year_matrix` will align the first year's December with the second year's January.  Here's a quick demonstration and a possible alternative:

<details>
  <summary>Click to expand!</summary>

```python
In [1]: import pandas as pd
   ...: signal = pd.Series(0, index=pd.date_range('2019-06-01', freq='d', periods=1000))
   ...: unique_years = signal.index.year.unique()
   ...: year_matrix = pd.concat([pd.Series(signal.loc[str(year)].values)
   ...:                          for year in unique_years],
   ...:                         axis=1, ignore_index=True)

In [2]: year_matrix
Out[2]: 
       0  1    2    3
0    0.0  0  0.0  0.0
1    0.0  0  0.0  0.0
2    0.0  0  0.0  0.0
3    0.0  0  0.0  0.0
4    0.0  0  0.0  0.0
..   ... ..  ...  ...
361  NaN  0  0.0  NaN
362  NaN  0  0.0  NaN
363  NaN  0  0.0  NaN
364  NaN  0  0.0  NaN
365  NaN  0  NaN  NaN

[366 rows x 4 columns]

In [3]: signal.rename('values').to_frame().assign(
   ...:     doy=signal.index.dayofyear,
   ...:     year=signal.index.year
   ...: ).pivot(index='doy', columns='year', values='values')
Out[3]: 
year  2019  2020  2021  2022
doy                         
1      NaN   0.0   0.0   0.0
2      NaN   0.0   0.0   0.0
3      NaN   0.0   0.0   0.0
4      NaN   0.0   0.0   0.0
5      NaN   0.0   0.0   0.0
..     ...   ...   ...   ...
362    0.0   0.0   0.0   NaN
363    0.0   0.0   0.0   NaN
364    0.0   0.0   0.0   NaN
365    0.0   0.0   0.0   NaN
366    NaN   0.0   NaN   NaN

[366 rows x 4 columns]
```
</details>

mdeceglie: Thanks for catching this @kanderso-nrel let's make sure we do something sensible about leap years.

rdtools/soiling.py
```diff
@@ -849,7 +862,7 @@ def annual_soiling_ratios(stochastic_soiling_profiles,
     Parameters
     ----------
     stochastic_soiling_profiles : list
-        List of pd.Series representing profile realizations from the SRR monte carlo.
+        List of o representing profile realizations from the SRR monte carlo.
```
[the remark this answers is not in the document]
asmunds: No

rdtools/soiling.py
asmunds: That's true. Perhaps we just replace the whole list with a dictionary?

setup.py
```diff
@@ -41,16 +41,18 @@
...
-    'statsmodels >= 0.9.0',
+    'statsmodels >= 0.11.1',
     'scipy >= 1.1.0',
     'h5py >= 2.8.0',
     'plotly>=4.0.0',
     'xgboost >= 1.3.3',
     'pvlib >= 0.7.0, <0.10.0',
     'scikit-learn >= 0.22.0',
+    'arch >= 4.11, <4.18',
+    'filterpy >= 1.4.5'
```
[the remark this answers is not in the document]
asmunds: Yup, I checked, and arch > 4.18 and filterpy above 1.4.2 works, so I changed this

asmunds:
Okay, I finally made time to get back to this (sorry for the delay). Thanks @kanderso-nrel for all the comments. I have now implemented all your suggestions. I hope you find the solutions satisfying :)

I'm not sure how to fix the readthedocs build-error. Any suggestions on that would be greatly appreciated.

kanderso-nrel:
Thanks @asmunds!  

> I'm not sure how to fix the readthedocs build-error. Any suggestions on that would be greatly appreciated.

I think this was fixed in #317, so if you merge `development` back into this branch I think it should work again.

asmunds:
> I think this was fixed in #317, so if you merge `development` back into this branch I think it should work again.

Of course. Thanks xD

The only thing I miss now is a good rendering of the attributes of the CODSAnalysis class. According to [the numpydoc Style guide](https://numpydoc.readthedocs.io/en/latest/format.html#documenting-classes), attributes can be documented under an Attributes header, but this does not render nicely in your readthedocs...

kanderso-nrel:
I agree it's not very pretty, but that's how the other classes are too (e.g. [here](https://rdtools--150.org.readthedocs.build/en/150/generated/rdtools.availability.AvailabilityAnalysis.html)), so I think it's okay.  I suspect it's a sphinx theme problem, not an issue with the docstring itself, so no need to fix it here IMHO.  Maybe someday we can figure out how to make them all render nicely.  

One other thing is a changelog entry in [pending.rst](https://github.com/NREL/rdtools/blob/development/docs/sphinx/source/changelog/pending.rst) :)

@mdeceglie have you found anything in your CODS tests that you want addressed here?

asmunds:
(GitHub is amazing - I did the latter two commits on my phone while brushing my teeth)

rdtools/soiling.py
```diff
@@ -1052,3 +1065,1552 @@ def monthly_soiling_rates(soiling_interval_summary, min_interval_length=14,
...
+                errors.append([b, ve])
+
+            # Print progress
+            if verbose:
+                _progressBarWithETA(b+1, reps, time.time()-t0, bar_length=30)
+
+        # Reweight and save weights
+        weights = 1 / np.array(RMSEs) / (1 + np.array(SR_is_1))
+        weights /= np.sum(weights)
+        self.knobs_n_weights = pd.concat(
```
mdeceglie, line 1849: note this brings up `FutureWarning: In a future version of pandas all arguments of concat except for the argument 'objs' will be keyword-only
  concat_tot_mod = pd.concat([kdf.total_model for kdf in bt_kdfs], 1)`

mdeceglie (commented):
This is looking great! Thanks @kanderso-nrel  for the detailed feedback and @asmunds for your help in finishing this up. I made a couple minor changes. I think most of my suggestions here can be addressed in a follow up PR rather than here.

rdtools/soiling.py
```diff
@@ -1052,3 +1065,1552 @@ def monthly_soiling_rates(soiling_interval_summary, min_interval_length=14,
...
+            iterative decomposition
+        degradation_method : string, default 'YoY'
+            Either 'YoY' or 'STL'. If anything else, 'YoY' will be assumed.
+            Decides whether to use the YoY method [3] for estimating the
+            degradation trend (assumes linear trend), or the STL-method (does
+            not assume linear trend). The latter is slower.
+        max_iterations : int, default 18
+            Max number of iterations to perform. Each iteration fits only one of the
+            components, so three iterations are needed to fit all three components.
+        detection_tuner : float, default .5
```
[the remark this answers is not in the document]
mdeceglie: Can wait for a follow up PR

rdtools/soiling.py
```diff
@@ -1052,3 +1065,1552 @@ def monthly_soiling_rates(soiling_interval_summary, min_interval_length=14,
...
+            Higher value gives lower cleaning event detection sensitivity.
+            Should be between 0.1 and 2
+        convergence_criterion : float, default 5e-3
+            the relative change in the convergence metric required for
+            convergence
+        pruning_iterations : int, default 1
+            Number of iterations when pruning (removing) cleaning events
+        pruning_tuner : float, default .6
```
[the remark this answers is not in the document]
mdeceglie: Can wait for a follow up PR

rdtools/soiling.py
```diff
@@ -1052,3 +1065,1552 @@ def monthly_soiling_rates(soiling_interval_summary, min_interval_length=14,
...
+
+    CODS.run_bootstrap(
+        reps=reps,
+        confidence_level=confidence_level,
+        verbose=verbose,
+        degradation_method=degradation_method,
+        process_noise=process_noise,
+        order_alternatives=order_alternatives,
+        detection_tuner_alternatives=detection_tuner_alternatives,
+        pruning_tuner_alternatives=pruning_tuner_alternatives,
```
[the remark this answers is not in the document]
mdeceglie: Can wait for a follow up PR

rdtools/degradation.py
```diff
@@ -267,31 +291,70 @@ def degradation_year_on_year(energy_normalized, recenter=True,
...
+    elif uncertainty_method == 'circular_block_bootstrap':
+        # Number of bootstrap repetitions
+        reps = 1000
+
+        # Construct degradation trend time series
+        N = len(energy_normalized)
+        numeric_index = np.arange(N)
+        days_per_index = \
+            (energy_normalized.dt.iloc[-1] - energy_normalized.dt.iloc[0]).days / N
+        degradation_trend = 1 + (Rd_pct / 100 / 365.24 * numeric_index
```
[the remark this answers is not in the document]
mdeceglie: We should figure this one out before merging

rdtools/soiling.py
mdeceglie: Can wait for a follow up PR

rdtools/degradation.py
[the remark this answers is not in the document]
asmunds: Well spotted! Yes it does, but that is easily sorted out by doing the same change wherever 365.24 is used in soiling.py.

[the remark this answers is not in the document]
asmunds: Though, I wonder, would it make sense to use the actual [tropical year](https://www.timeanddate.com/astronomy/tropical-year.html), i.e. ~8766 hours, as the reference year length when calculating the degradation rate? In my understanding, this would give a more strictly correct degradation rate. (The current estimate is only off by 0.066 %. But accumulating over a 20 year period, that would amount to a performance offset of 1.3 %, which may not be negligible.) Perhaps consider in a future PR?

rdtools/soiling.py
```diff
@@ -1052,3 +1065,1552 @@ def monthly_soiling_rates(soiling_interval_summary, min_interval_length=14,
...
+        complete overlap with input pi
+    RMSE : float
+        Root Means Squared Error of total model vs input pi
+    small_soiling_signal : bool
+        Whether or not the signal is deemed too small to infer anything
+        about it
+    adf_res : list
+        The results of an Augmented Dickey-Fuller test (telling whether the
+        residuals are stationary or not)
+    knobs_n_weights : pandas.DataFrame
```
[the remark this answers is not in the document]
asmunds: Made it private

asmunds:
Great, thanks @mdeceglie. I concur on the renaming of the tuner-attributes, but it can wait for a future PR. Should we create an Issue that adresses these attribute names?

mdeceglie:
Thanks for the updates @asmunds! Yes we should open an issue with some items for future PRs. Here is the list I have:

1. `FutureWarning` from `pd.concat`
2. Rename parameters with `_tuner` in the name
3. Update the CODS example to use PVDAQ4 data

Am I missing anything @kanderso-nrel  or @asmunds? Otherwise I think this is ready to go at long last, thank you!

kanderso-nrel:
A couple other things, can be follow-ups:
- Changelog entries for both new functionality and dependencies
- Consider removing `bootstrap` from the `Submodules` heading in api.rst since it doesn't have any public content

mdeceglie:
Thanks again @asmunds  for sticking with this. Happy to finally get it merged!

asmunds:
Alright, I'm so excited to see this merged after 2 years! Thank you @mdeceglie and @kanderso-nrel for reviewing and even contributing directly - that is greatly appreciated.

PR #152, opened 2020-03-06
Update normalization.irradiance_rescale and create irradiance_rescale_test.py

### LINKED ISSUE #151: normalization.irradiance_rescale is not tested
About half of the untested lines of code in the dev branch are in `normalization.irradiance_rescale`.  Covering those with a new test would get coverage to ~94% which would put us in pretty good shape.  

```
$ coverage run -m pytest
...
$ coverage report -m
Name                              Stmts   Miss  Cover   Missing
---------------------------------------------------------------
rdtools\__init__.py                  18      0   100%
rdtools\aggregation.py                3      0   100%
rdtools\clearsky_temperature.py      78     10    87%   51-54, 89, 91, 128, 131, 135, 140
rdtools\degradation.py              118     10    92%   120, 226, 232, 241, 270, 341, 343, 353-356
rdtools\filtering.py                 10      0   100%
rdtools\normalization.py            218     52    76%   260-266, 335-401, 499, 514-517, 720
rdtools\soiling.py                  222      7    97%   40, 44, 49, 220, 409, 420, 426
---------------------------------------------------------------
TOTAL                               667     79    88%
```

Relevant: https://github.com/NREL/rdtools/blob/development/rdtools/test/clearsky_temperature_test.py#L39-L40

kanderso-nrel:
Addresses #151.  Coverage of `normalization.py` goes from 76% to 94%:

```
Name                              Stmts   Miss  Cover
-----------------------------------------------------
rdtools\__init__.py                  18      0   100%
rdtools\aggregation.py                3      0   100%
rdtools\clearsky_temperature.py      78     10    87%
rdtools\degradation.py              118     10    92%
rdtools\filtering.py                 10      0   100%
rdtools\normalization.py            218     12    94%
rdtools\soiling.py                  222      7    97%
-----------------------------------------------------
TOTAL                               667     39    94%
```


kanderso-nrel:
A little background: 

```
@pytest.fixture
def simple_irradiance():
    times = pd.date_range('2019-06-01 12:00', freq='15T', periods=5)
    time_series = pd.Series([1,2,3,4,5], index=times)
    return time_series
```
Fixtures are objects that are used in multiple tests, but pytest does some clever stuff to make sure a fresh copy of the object is passed to each function.  So this is just making some dummy test data to be used in the actual test functions.  Pytest matches up fixtures with test functions by matching the fixture name with test function parameter names. 

```
@pytest.mark.parametrize("method", ['iterative', 'single_opt'])
def test_rescale(method, simple_irradiance):
    # test basic functionality
    ...
```
The `parametrize` decorator is basically just a fancy for loop; it calls the marked test for each of the specified parameter values.  So this function will get called twice -- once with `method='iterative'` and once with `method='single_opt'`, and both invocations will be given `simple_irradiance` according to the fixture's return value.  It's handy because each parameter combination is counted as a separate test:

```
(base) C:\Users\KANDERSO\projects\rdtools>pytest rdtools\test\irradiance_rescale_test.py
...
rdtools/test/irradiance_rescale_test.py::test_rescale[iterative] PASSED      [ 20%]
rdtools/test/irradiance_rescale_test.py::test_rescale[single_opt] PASSED     [ 40%]
rdtools/test/irradiance_rescale_test.py::test_warn PASSED                    [ 60%]
rdtools/test/irradiance_rescale_test.py::test_max_iterations PASSED          [ 80%]
rdtools/test/irradiance_rescale_test.py::test_non_convergence PASSED         [100%]
```

mdeceglie:
Should we change the default behavior regarding method in 2.0?

rdtools/test/irradiance_rescale_test.py
```diff
@@ -0,0 +1,51 @@
...
+def test_rescale(method, simple_irradiance):
+    # test basic functionality
+    modeled = simple_irradiance
+    measured = 1.05 * simple_irradiance
+    rescaled = irradiance_rescale(measured, modeled, method=method)
+    expected = measured
+    assert_series_equal(rescaled, expected, check_exact=False)
+    
+
+def test_warn(simple_irradiance):
```
mdeceglie: If we don't change the default behavior, a more descriptive name for this test would be helpful.
e.g. test_method_equals_none_warns
Or include a comment explaining the test

rdtools/test/irradiance_rescale_test.py
```diff
@@ -0,0 +1,51 @@
...
+    with pytest.warns(UserWarning,
+                      match="underlying calculations for irradiance_rescale"):
+        _ = irradiance_rescale(simple_irradiance, simple_irradiance)
+
+
+def test_max_iterations(simple_irradiance):
```
mdeceglie: This raises UnboundLocalError, a subclass of NameError. This works, but curious why you decided to test for NameError?

mdeceglie: Is this test redundant with `test_non_convergence()`?

kanderso-nrel: Good catch!  I used NameError because I had been running the body of `irradiance_rescale` manually in a console to explore its behavior -- since it was executing at global scope, it raised NameError instead of UnboundLocalError.

kanderso-nrel: I suppose the original intent is redundant since both of them verify that the `max_iterations` parameter changes function behavior, and `test_non_convergence()` is the better test.  

This makes me wonder if the specific behavior `test_max_iterations()` is testing should be changed.   One could imagine a rescale function that returns the input unchanged (or raises ConvergenceError) when `max_iterations=0` and it seems the reason this function doesn't do that is just an implementation detail rather than intended design.  Should we change it?  If not, I think we should document that max_iterations must be greater than zero.

rdtools/test/irradiance_rescale_test.py
kanderso-nrel: I'd say making `method='iterative'` the default with no warning is appropriate in 2.0.  Happy to add that to this PR and remove this test if people agree.

kanderso-nrel:
@mdeceglie do you want to change the default behavior for `method` in this PR?

rdtools/test/irradiance_rescale_test.py
mdeceglie: Thanks for pointing this out. Let's take this opportunity to think through rescale and perhaps align its behavior more closely with `scipy.optimize.minimize` for example. I suggest we:

- allow it to work with `max_iterations=0`
- Raise a warning rather than an error when convergence isn't reached within `max_iterations`
-expose `convergence_threshold` as an optional parameter

What do you think of these suggestions?

kanderso-nrel: +1 to 1 and 3.  For changing the error to a warning, I could go either way.  I'm pretty desensitized to numerical warnings and would likely not pay attention to something similar from rdtools -- the way I use python, it's unusual that a warning is actually about something I care about.  Usually it's about divide-by-zero at night or something that is fine to ignore.

IMO issuing a warning but returning the rescaled data anyway is some form of endorsement that the results are still useful, but I'd expect non-convergence to be associated with something funky with the data that ought to be fixed before analyzing.  With `max_iterations` and `convergence_threshold` exposed, do you think that returning non-converged rescalings is useful?

mdeceglie: I could also be convinced either way on 2. And it is more conservative to raise an error, so let's stick with the error on that one.

kanderso-nrel:
@mdeceglie I think I've addressed all the discussion points, PTAL

Issue #129, opened 2019-11-19
Bug: SRR analysis/_calc_daily_df error


kanderso-nrel:
The following line of code raised an error because `run[run.pi_norm > 0]` was of length zero and as such its index had no element 0:

https://github.com/NREL/rdtools/blob/development/rdtools/soiling.py#L196

Maybe the loop should include a check for `0 <= len(run) < 2`.  Unfortunately I don't have a reproducible example -- my code was broken in other ways and has since been fixed, so maybe the problem was because I was feeding it bad data.  I could probably recreate the broken code from git history, let me know if that would be useful.

cwhanse:
Just a suggestion, to test for length zero `run.empty` is probably better than `len(run)`

kanderso-nrel:
A little more info, the SystemAnalysis example notebook in the model_chains branch will cause this error if you modify the sa.clearsky_analysis call to include `srr_soiling`.

mdeceglie:
Based on this information, the problem seems to occur when a `run` i.e. soiling interval, consists of no normalized performance values above 0, this could occur for example if the whole run is NaN. In this case, I think we want to treat that interval as not valid, and give it the default invalid row in the result dataframe which summarizes soiling intervals. In addition if there are a few <0 or NaN points at the beggining or end of an otherwise valid interval, it looks like the current version will give incorrect start and end dates . The solution might look something like:
```python
for r in res_loop:
    run = daily_df[daily_df['run'] == r]
    length = (run.day[-1] - run.day[0])
    start_day = run.day[0]
    end_day = run.day[-1]
    start = run.index[0],
    end = run.index[-1]
    run_filtered = run[run.pi_norm > 0]
# use the filtered version if it contains any points
# otherwise use the unfiltered version to populate a
# valid=False row
    if not run_filterd.empty:
        run = run_filtered  
    result_dict = {
        'start': start,
        'end': end,
        'length': length,
        'run': r,
        'run_slope': 0,
        'run_slope_low': 0,
        'run_slope_high': 0,
        'max_neg_step': min(run.delta),
        'start_loss': 1,
        'clean_wo_precip': run.clean_wo_precip[0],
        'inferred_start_loss': run.pi_norm.mean(),
        'inferred_end_loss': run.pi_norm.mean(),
        'valid': False
    }
```

PR #154, opened 2020-03-12
SRR fixes

kanderso-nrel:
Noticed a couple small things to fix with the SRR code.  Could include #129 here as well, but I'm not very familiar with that code so I assigned it to @mdeceglie.


mdeceglie:
For these minor changes, I'm fine to merge without @cdeline's review.

PR #156, opened 2020-03-17
better GitHub links in documentation

kanderso-nrel:
The "edit on GH" links on generated pages in our documentation don't actually point to the source code they should.  See for example the links here:
https://rdtools.readthedocs.io/en/latest/generated/rdtools.soiling.srr_analysis.run.html
and
https://rdtools.readthedocs.io/en/latest/example.html

This is a port of the fix in https://github.com/pvlib/pvlib-python/pull/913.  The approach is to override the default links and insert custom links based on our code/docs structure.


docs/sphinx/source/conf.py
```diff
@@ -84,4 +85,98 @@
...
+    provides templates with a built-in `pagename` variable that is the path
+    at the end of the URL, without the extension.  For instance,
+    https://rdtools.rtfd.io/en/latest/generated/rdtools.soiling.soiling_srr.html
+    will have pagename = "generated/rdtools.soiling.soiling_srr".
+    """
+
+    # RTD automatically sets READTHEDOCS_VERSION to the version being built.
+    if os.environ.get('READTHEDOCS_VERSION', None) == 'stable':
+        branch = 'master'
+    else:
```
mdeceglie: Let's  be explicit about all the cases here. Otherwise, if we change the readthedocs configuration, we could end up with links pointing to development when they shouldn't. Not sure what the correct behavior should be for PR and local builds though. I'd almost rather it be broken than point to the wrong branch.

kanderso-nrel: Handling PR build correctly seems like a huge headache for not much benefit, and I'm not even sure what a local build should point to.  What do you think about only generating the link if it's stable/master or latest/development?  For other builds it would just omit that HTML element.

kanderso-nrel:
Changed to only include GH links when it thinks it's on RTD and building development or master.  You can use `set READTHEDOCS_VERSION=stable&& make html` for testing out the behavior locally.  Is this getting too complicated?

kanderso-nrel:
It occurs to me that this logic won't be future-proof because old documentation versions will always link to the most recent master and development.  But also since we're planning to stop using the development branch and instead use the same branch/merge style as pvlib, the URL logic here needs updating anyway.  I'm guessing its RTD versioning is what ours will look like over time: https://readthedocs.org/projects/pvlib-python/versions/

The correct logic might be:

- if `rtd_version == 'latest'`, point to master on github
- elif `rdtools.__version__` is a clean version, use it to point the the appropriate tag on github
- else don't generate the link 

So as long as RTD rebuilds `latest` on every merge to master and other builds are always associated with a tagged commit, I think this logic should suffice...  Are there any cases this wouldn't handle correctly?

kanderso-nrel:
As a (possibly) temporary workaround to the above-mentioned complexity, we can just change the link text to say "View most recent on github" or something and not worry about linking to the correct version.

kanderso-nrel:
"View most recent on github" isn't necessarily appropriate to use for links to both master and development; this updates the link to display `View on github/master` or `View on github/development` depending on the docs version.  Other builds (local, PR etc) won't have a link.  @mdeceglie does that sound good?

mdeceglie:
That sounds like a good approach, thanks!

PR #164, opened 2020-04-16
Remove conflicting reqs

mdeceglie:
Some packages were specified in both notebook_requirements.txt and requirements.txt. In some cases versions conflicted.


kanderso-nrel:
Tracking dependencies in both `requirements.txt` and `notebook_requirements.txt` seems annoying, and just removing the overlap like this might be a good solution.  Are we okay with having `notebook_requirements.txt` act as the full environment spec only when paired with `requirements.txt`?  Seems reasonable to me, but could consider mentioning this in the example notebook since it only says to install `notebook_requirements.txt`.

mdeceglie:
I ran the notebook. I'm seeing very similar changes to the output as #144

mdeceglie:
Given that we let `requirements.txt` and `notebook_requirements.txt` diverge once already, I'm in favor of removing the repeat packages.

mdeceglie:
- This includes the bug fix from #166 
- Adds `np.random.seed(0)` to the notebook to ensure consistency

requirements.txt
```diff
@@ -1,11 +1,14 @@
+cycler==0.10.0
 h5py==2.10.0
-numpy==1.17.3
+kiwisolver==1.2.0
```
kanderso-nrel, line 3: Looks like in addition to resolving conflicting versions, some packages were moved from notebook_requirements.txt to requirements.txt (kiwisolver, cycler, patsy).  `pip show` tells me they are all required by matplotlib, which isn't just used in the notebook, so it makes sense to include them in the main requirements file.  Is that the reasoning?

docs/sphinx/source/changelog/v2.0.0.rst
```diff
@@ -4,7 +4,7 @@ v2.0.0 (MONTH DAY, YEAR)
 
 API Changes
 -----------
-* Remove `low_power_cutoff` parameter in `filtering.clipping_filter` (:issue:`84`).
+* Remove ``low_power_cutoff`` parameter in ``filtering.clipping_filter`` (:issue:`84`).
```
kanderso-nrel, line 7:
```suggestion
* Remove ``low_power_cutoff`` parameter in :py:func:`~rdtools.filtering.clip_filter` (:issue:`84`).
```

thanks for cleaning this up

docs/sphinx/source/changelog/v2.0.0.rst
```diff
@@ -30,6 +30,12 @@ Documentation
 Requirements
 ------------
 * Drop support for Python 2.7, minimum supported version is now 3.6 (:pull:`135`).
+* Update requirements.txt and notebook_requirements.txt to avoid conflicting specifications. Taken together,
+  they represent the complete environment for the notebook example.
```
kanderso-nrel, line 34:
```suggestion
  they represent the complete environment for the notebook example (:pull:`164`).
```

docs/sphinx/source/changelog/v2.0.0.rst
```diff
@@ -30,6 +30,12 @@ Documentation
...
+
+Example Updates
+---------------
+* Seed ``numpy.random`` to ensure repeatable results
```
kanderso-nrel, line 38:
```suggestion
* Seed ``numpy.random`` to ensure repeatable results (:pull:`164`).
```

requirements.txt
mdeceglie: Yep, that is the reasoning. Following https://blog.miguelgrinberg.com/post/the-package-dependency-blues, let's keep indirect requirements in `requirements.txt`

PR #166, opened 2020-04-20
Update the normalized filter in the clearsky workflow

mdeceglie:
The normalized filter in the clear sky workflow should be adjusted to match that in the sensor analysis. Fixes bug from #144. Doesn't affect clear sky results.


kanderso-nrel:
Here are some reproduction results.  In all cases I am using conda environments originally set up with `conda create -n env_name python=3.7`.  

| environment | requirements | sensor Rd CI upper bound |
|-------------|--------------|----------|
| fresh       | #166         | -0.33    |
| fresh       | #164         | -0.33    |
| old         | ??         | -0.36    |

The "old" environment is my base conda environment with whatever random library combinations I happen to have installed for scratch work.  I didn't run into an installation error in either of the fresh environments.  I don't have an explanation for how this PR got -0.36.

mdeceglie:
We tracked down the changes in the upper limit of the confidence interval to stochastic differences in the boo-strapping across systems and package versions.  Using `np.random.seed()` yields repeatable results.
Screen Shot 2020-04-21 at 11 36 25 AM
We tracked the behavior in the upper limit of the

mdeceglie:
This fix ended up in #164. Closing this

Issue #143, opened 2020-02-06
compatibility with pvlib 0.7.x


kanderso-nrel:
pvlib 0.7 introduced two API changes that break RdTools, specifically `normalization.sapm_dc_power`.  The easiest way forward would just be to bump the pvlib requirement to 0.7.1, but assuming we want to support both 0.6.x and 0.7.x, here are some possible workarounds:

#### pvlib.pvsystem.sapm_effective_irradiance
pvlib 0.6.x divides `effective_irradiance` by `reference_irradiance` (default 1000).  RdTools passes in `reference_irradiance=1` to keep units W/m2, but pvlib 0.7 dropped the `reference_irradiance` argument so this line now raises TypeError: https://github.com/NREL/rdtools/blob/development/rdtools/normalization.py#L182

- make the call with `reference_irradiance=1` and if TypeError is raised, call it again without `reference_irradiance`
- check the pvlib version at runtime and include the argument if it is appropriate (example: https://stackoverflow.com/a/11887885/1641381). 
- drop the parameter and check if all return values are < ~1.5.  If so, multiply result by 1000.

#### pvlib.pvsystem.sapm_celltemp
The class method `PVSystem.sapm_celltemp` is deprecated in 0.7.0 and also had the argument names and order change.  The recommended replacement is `pvlib.temperature.sapm_cell` (new in 0.7.0).  

- Because `sapm_celltemp` arguments changed in 0.7.0, the code in `development` raises TypeError.  Similar to above, one option is to try the code as-is and fall back to `temperature.sapm_cell` if it fails.  
- Check pvlib version at runtime, dispatch appropriately. 

I think I prefer the version-checking method for both cases because catching `TypeError` is an overly broad test.  It does put a dependency on the `packaging` package, but it [comes with](https://github.com/pypa/setuptools/tree/master/pkg_resources/_vendor/packaging) `setuptools` so in theory should already exist anywhere RdTools is installed (I think?).  Any thoughts?

kanderso-nrel:
ping @mikofski since he usually has good ideas about this sort of thing

cwhanse:
Happy to help to fix this compatibility if you need it.

> assuming we want to support both 0.6.x 

What is driving this assumption?

kanderso-nrel:
Thanks @cwhanse!

>>assuming we want to support both 0.6.x
>
>What is driving this assumption?

[...]

PR #170, opened 2020-04-22
migrate to pvlib>=0.7.0

kanderso-nrel:
See #143.  This PR updates how we use the `sapm_effective_irradiance` and `sapm_cell` functions to match the [changes introduced in pvlib 0.7.0](https://pvlib-python.readthedocs.io/en/latest/whatsnew.html#v0-7-0-december-18-2019).  Note that since the notebook calculates cell temperature, it also needed updates (no change in results).

Side note -- I would be interested to learn why I get warnings when running the notebook with the notebook environment but apparently other people do not.  Need to get rid of those warnings before merging this.


kanderso-nrel:
@cwhanse - I think I've done these updates correctly, but if you are able to take a look to verify, that would be great.

rdtools/normalization.py
```diff
@@ -184,17 +184,16 @@ def sapm_dc_power(pvlib_pvsystem, met_data):
...
                                    aoi=aoi,
-                                   module=pvlib_pvsystem.module,
-                                   reference_irradiance=1)
+                                   module=pvlib_pvsystem.module)
 
     temp_cell = pvlib_pvsystem\
-        .sapm_celltemp(irrad=total_irradiance['poa_global'],
-                       wind=met_data['Wind Speed'],
-                       temp=met_data['Temperature'])
+        .sapm_celltemp(poa_global=total_irradiance['poa_global'],
```
cwhanse, line 190: This assumes that `pvlib_pvsystem.temperature_model_parameters` are set, or can be inferred from `pvlib_pvsystem.racking_model` and `pvlib_pvsystem.module_type`. I see that `racking_model` is set, but I didn't see where `module_type` is set (values can be `'glass_polymer'` and `'glass_glass'`, default `'glass_polymer'`). The code should work as is, but to be explicit about the temperature model, I would suggest setting `pvlib_pvsystem.module_type` or setting `pvlib_pvsystem.temperature_model_parameters` if that's not already being done. No promises if the current default behavior will continue after pvlib v0.8.0.

That's the only thing I see.

mdeceglie:
@kanderso-nrel When I remember to, I delete the warnings from the raw notebook JSON before finalizing an update to the notebook. It give the example a cleaner appearance, but perhaps it is confusing and we should leave them?

rdtools/normalization.py
kanderso-nrel: Thanks!  I've updated the docstrings and tests to be explicit about the temperature model.  I appreciate the review.

rdtools/normalization.py
```diff
@@ -143,7 +143,9 @@ def sapm_dc_power(pvlib_pvsystem, met_data):
     ----------
     pvlib_pvsystem : pvlib-python LocalizedPVSystem object
         Object contains orientation, geographic coordinates, equipment
-        constants (including DC rated power in watts).
+        constants (including DC rated power in watts).  The object must also
+        specify either the `temperature_model_parameters` attribute or both
+        `racking_model` and `module_type` to infer the model parameters.
```
cwhanse, line 148:
```suggestion
        `racking_model` and `module_type` attributes to infer the temperature model parameters.
```

rdtools/normalization.py
wholmgren: Note these are positional args, not keyword args, so I don't promise to deprecate the names if we decide to change them. In particular, I think `temp_air` is going to change at some point in the near future.

kanderso-nrel: Thanks @wholmgren.  `setup.py` does specify `pvlib<0.8.0`, but might as well try to be forward-compatible if possible.

kanderso-nrel:
> I delete the warnings from the raw notebook JSON before finalizing an update to the notebook. It gives the example a cleaner appearance, but perhaps it is confusing and we should leave them?

To the extent that the notebook acts as a reproducibility test for the user, I guess warnings are a part of the expected output, so I think I'd be in favor of leaving them in the notebook.  But maybe it would be better to just fix whatever is raising the warnings?  I get two warnings with the current notebook:
- pandas, `register_matplotlib_converters()`
- pandas, `Converting timezone-aware DatetimeArray to timezone-naive ndarray with 'datetime64[ns]' dtype.` from the `df.reindex` in the clearsky temp function

Using the most recent pandas version gets rid of both warnings.  And using an updated statsmodels gets rid of the third warning that the new pandas raises. 😉  Thoughts on bumping those requirements in a new PR?

mdeceglie:
I'm in favor of bumping requirements to get rid of warnings

rdtools/normalization.py
mdeceglie: If things change in future versions of pvlib, is this a case where we can use conditionals to continue to support 0.7?

mdeceglie: On a related note: If in the future we change to requiring pvlib>0.8, but that doesn't change the RdTools calculations or API, it seems we don't need a major version. (See [Semantic Versioning FAQ](https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api))

PR #171, opened 2020-04-23
Add example notebook using the NREL PVDAQ #4 system

kanderso-nrel:
Remaining To-Do items for this notebook:
- [x] Need to publicly host the dataset somewhere and add the link to this notebook
- [x] Update Step 0 description ("nothing in this first section utlizes the rdtools library")
- [x] Figure out what weather columns to use
- [x] Any changes necessary for using this higher-resolution datasets?  `max_timedelta`, e.g.
- [x] Step 2 description update
- [x] Get better data for the soiling analysis
- [x] Rename normalized filter vars (#139)
- [x] Update pvlib calls for 0.7.0 (#170)
- [x] Update commentary in Clear Sky 4


kanderso-nrel:
Test failures are because I've bumped the pvlib version for this PR so that the notebook can use the new cell temp interface, but the tests still use the old interfaces.  Should be fixed by merging #170.

docs/degradation_and_soiling_example_pvdaq_4.ipynb
```diff
@@ -360,25 +362,39 @@
    "execution_count": 10,
    "metadata": {},
    "outputs": [],
+   "source": [
+    "# Apply artificial soiling signal for example\n",
+    "# be sure to remove this for applicaitons on real data,\n",
```
kanderso-nrel, line 367:
```suggestion
    "# be sure to remove this for applications on real data,\n",
```

docs/degradation_and_soiling_example_pvdaq_4.ipynb
```diff
@@ -615,6 +631,8 @@
     "\n",
     "In this example, note that we have omitted wind data in the cell temperature calculations for illustrative purposes. Wind data can also be included when the data source is trusted for improved results\n",
     "\n",
+    "We generally reccomend that the clear sky workflow be used as a check on the sensor workflow. It tends to be more sensitive than the sensor workflow, and thus we don't reccomend it as a stand-alone analysis.\n",
```
kanderso-nrel, line 634:
```suggestion
    "We generally recommend that the clear sky workflow be used as a check on the sensor workflow. It tends to be more sensitive than the sensor workflow, and thus we don't recommend it as a stand-alone analysis.\n",
```

mdeceglie:
I think this is good to go once we get the data hosted somewhere. Will we be eliminating the old example notebook? Or maybe we should carry them both around for a while, at least until we are ready to release 2.0?

kanderso-nrel:
+1 to keeping the old notebook around for a while since it's a good indicator of changes in library behavior.  At some point the new notebook will fill that role but I think it's worth keeping the old one around for now.

cdeline:
Yeah, I agree we should keep the old notebook around as a legacy example.  Maybe even plan to keep it around until it breaks from some API change.

mdeceglie:
@kanderso-nrel  and @cdeline Let me know what you think of this notebook modification. It leaves modeling to pvlib, which I think is good for the open source PV ecosystem, and avoids converting PV to energy, which may be a confusing step for simple applications.

I suggest leaving the old notebook as is for the sake of diversity of examples.

kanderso-nrel:
Looks OK to me.  One other thing I notice is that the notebook renders with scrollbars on the code blocks when a line gets too long, which is a little annoying: https://rdtools.readthedocs.io/en/latest/example.html

Maybe we should be applying a line length limit to the notebook, too.  A good bit of the clearsky normalization line is going to be off-screen.

mdeceglie:
Good idea, I limited the line length to 90 characters, I think that will get rid of the scroll bars for a reasonable window size.

kanderso-nrel:
LGTM, one small typo in "The example automatially downloads the dataset used in this example."

It took a bit of waiting to download the CSV from the data hub, maybe we should store a local copy and only fetch it from the url if it doesn't already exist on disk?  Just a convenience for when we're running the notebook ourselves over time.  Not a big deal to me either way.

mdeceglie:
I considered a local cache, so since you bring it up too it seems like a good idea. I'll add that and fix the typo. Thanks!

kanderso-nrel:
I can't approve a PR I opened so here's a thumbs up instead 👍

Issue #141, opened 2020-02-04
energy_from_power/normalize_with_X regression in 2.0.0-alpha.0


kanderso-nrel:
`normalization.energy_from_power` has unintuitive (to me) output when the power timeseries is already at the target frequency:

```
In [3]: idx = pd.date_range('2019-01-01', periods=5, freq='h')
   ...: power = pd.Series([1,2,3,2,1], index=idx)
   ...: energy = rdtools.normalization.energy_from_power(power, 'h')
   ...: print(energy)
2019-01-01 01:00:00    1.5
2019-01-01 02:00:00    2.5
2019-01-01 03:00:00    2.5
2019-01-01 04:00:00    1.5
Freq: H, Name: energy_Wh, dtype: float64
```

Since the `normalize_with_` functions call `energy_from_power`, their behavior changed.  

Here's a comparison of normalizing between versions.  In all cases, input power is generated using the normalization method, so I would expect the normalized power to be exactly 1.0 at all timestamps.  That's the case for 1.2.2 but not 2.0.0-alpha.0:

RdTools 1.2.2:

![image](https://user-images.githubusercontent.com/57452607/73771403-8e0bc500-473b-11ea-8cdb-f95bf8e677ec.png)

RdTools 2.0.0-alpha.0:

![image](https://user-images.githubusercontent.com/57452607/73771541-cca17f80-473b-11ea-8844-6f16713636ce.png)

Code to generate the above plots:
<details>
  <summary>Click to expand</summary>

```
import pandas as pd
import pvlib
from pvlib.pvsystem import PVSystem, LocalizedPVSystem
import rdtools
import matplotlib.pyplot as plt


def make_sapm_inputs():
    lat, lon, altitude, timezone = (30, -110, 700, 'Etc/GMT+7')
    sandia_modules = pvlib.pvsystem.retrieve_sam('SandiaMod')
    sapm_inverters = pvlib.pvsystem.retrieve_sam('cecinverter')

    module = sandia_modules['Canadian_Solar_CS5P_220M___2009_']
    module['pdc0'] = module['Impo'] * module['Vmpo']
    module['gamma_pdc'] = -0.0043
    inverter = sapm_inverters['ABB__MICRO_0_25_I_OUTD_US_208_208V__CEC_2014_']

    times = pd.date_range(start='2015', end='2016', freq='h', tz=timezone)

    system = PVSystem(module=module,
                      module_parameters=module,
                      inverter=inverter,
                      surface_tilt=20,
[...]

PR #172, opened 2020-04-24
changelog for energy_from_power, normalize_with_x

kanderso-nrel:
See #141

### DIFF
--- a/docs/sphinx/source/changelog/v2.0.0.rst
+++ b/docs/sphinx/source/changelog/v2.0.0.rst
@@ -4,6 +4,10 @@ v2.0.0 (MONTH DAY, YEAR)
 
 API Changes
 -----------
+* The behavior of :py:func:`~rdtools.normalization.normalize_with_pvwatts`
+  and :py:func:`~rdtools.normalization.normalize_with_sapm` have changed to
+  expect input energy in the moving average form calculated by
+  :py:func:`~rdtools.normalization.energy_from_power` (:pull:`105`, :pull:`108`).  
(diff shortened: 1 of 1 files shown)


kanderso-nrel:
> Suggested a change, what do you think?

I don't see any suggestions :(

docs/sphinx/source/changelog/v2.0.0.rst
```diff
@@ -4,14 +4,18 @@ v2.0.0 (MONTH DAY, YEAR)
 
 API Changes
 -----------
+* The behavior of :py:func:`~rdtools.normalization.normalize_with_pvwatts`
+  and :py:func:`~rdtools.normalization.normalize_with_sapm` have changed to
+  expect input energy in the moving average form calculated by
+  :py:func:`~rdtools.normalization.energy_from_power` (:pull:`105`, :pull:`108`).
```
mdeceglie, line 10: I would characterize it as a trapezoidal rule integration rather than a moving average, but that may be moot because I suggest revising to:

>The calculations internal to normalize_with_pvwatts and normalize with sapm have changed. Generally, when working with raw power data it should be converted to right-labeled energy with energy_from_power before being used with the normalization functions.

The PR links will allow folks to dig into the details if they need to.

mdeceglie:
Sorry about that, should be there now.

docs/sphinx/source/changelog/v2.0.0.rst
kanderso-nrel: The moving average bit was quoting this: https://github.com/NREL/rdtools/blob/development/rdtools/normalization.py#L502

But agreed that your version is better anyway.

PR #173, opened 2020-05-05
Generalized model normalizaiton

mdeceglie:
Adds a function which enables normalization with any modeled power time series (see #102). We should also consider using this function in the new example of #171.

### DIFF
--- a/rdtools/normalization.py
+++ b/rdtools/normalization.py
@@ -30,7 +30,7 @@ def normalize_with_expected_power(pv, expected_power, irradiance, pv_input='powe
     Returns
     -------
     normalized_energy : pd.Series
-        Energy normalized based on `expected_power
+        Energy normalized based on `expected_power`
     insolation : pd.Series
         Insolation associated with each normalized point
 
@@ -44,7 +44,7 @@ def normalize_with_expected_power(pv, expected_power, irradiance, pv_input='powe
         energy = pv.copy()
         energy.name = 'energy_Wh'
     else:
-        raise ValueError("Unexpected value for pv_input. pv_input should be 'power' of 'energy'.")
+        raise ValueError("Unexpected value for pv_input. pv_input should be 'power' or 'energy'.")
 
     model_tds, mean_model_td = delta_index(expected_power)
     measure_tds, mean_measure_td = delta_index(energy)
--- a/rdtools/normalization.py
+++ b/rdtools/normalization.py
@@ -11,6 +11,61 @@ class ConvergenceError(Exception):
     '''Rescale optimization did not converge'''
     pass
 
+def normalize_with_expected_power(pv, expected_power, irradiance, pv_input='power'):
+    '''
+    Normalize pv output based on expected PV power.
+
+    Parameters
+    ----------
+    pv : pd.Series
+        Right-labeled time series PV energy or power. If energy, should *not*
+        be cumulative, but only for preceding time step.
+    expected_power : pd.Series
+        Right-labeled time series of expected PV power.
+    irradiance : pd.Series
+        Right labeled time series of plane-of-array irradiance associated with `expected_power`
+    pv_input : str
+        'power' or 'energy' to specify type of input used for pv parameter
+
+    Returns
+    -------
+    normalized_energy : pd.Series
+        Energy normalized based on `expected_power
+    insolation : pd.Series
+        Insolation associated with each normalized point
+
+    '''
+    
+    freq = check_series_frequency(pv, 'pv')
+
+    if pv_input == 'power':
+        energy = energy_from_power(pv, freq)
+    elif pv_input == 'energy':
+        energy = pv
+    else:
+        raise ValueError("Unexpected value for pv_input. pv_input should be 'power' of 'energy'.")
+
+    model_tds, mean_model_td = delta_index(expected_power)
+    measure_tds, mean_measure_td = delta_index(energy)
+
+    # Case in which the model is as or more ruent than the measurements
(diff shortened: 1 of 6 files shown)


rdtools/normalization.py
```diff
@@ -11,6 +11,62 @@ class ConvergenceError(Exception):
...
+        Right-labeled time series of expected PV power.
+    irradiance : pd.Series
+        Right labeled time series of plane-of-array irradiance associated with `expected_power`
+    pv_input : str
+        'power' or 'energy' to specify type of input used for pv parameter
+
+    Returns
+    -------
+    normalized_energy : pd.Series
+        Energy normalized based on `expected_power
```
kanderso-nrel, line 33:
```suggestion
        Energy normalized based on `expected_power`
```

rdtools/normalization.py
```diff
@@ -11,6 +11,62 @@ class ConvergenceError(Exception):
...
+    
+    freq = check_series_frequency(pv, 'pv')
+
+    if pv_input == 'power':
+        energy = energy_from_power(pv, freq)
+    elif pv_input == 'energy':
+        energy = pv.copy()
+        energy.name = 'energy_Wh'
+    else:
+        raise ValueError("Unexpected value for pv_input. pv_input should be 'power' of 'energy'.")
```
kanderso-nrel, line 47:
```suggestion
        raise ValueError("Unexpected value for pv_input. pv_input should be 'power' or 'energy'.")
```

rdtools/normalization.py
```diff
@@ -11,6 +11,62 @@ class ConvergenceError(Exception):
...
+        insolation = energy_from_power(irradiance, freq)
+
+    # Case in which the model less frequent than the measurements
+    elif mean_model_td > mean_measure_td:
+
+        expected_power = interpolate(expected_power, pv.index)
+        expected_energy = energy_from_power(expected_power, freq)
+
+        irradiance = interpolate(irradiance, pv.index)
+        insolation = energy_from_power(irradiance, freq)
```
kanderso-nrel, line 65: Both branches of this if/elif block call `insolation = energy_from_power(irradiance, freq)` and `expected_energy = energy_from_power(expected_power, freq)`, I think could change to:
```python
    if mean_model_td > mean_measure_td:
        expected_power = interpolate(expected_power, pv.index)
        irradiance = interpolate(irradiance, pv.index)

    expected_energy = energy_from_power(expected_power, freq)
    insolation = energy_from_power(irradiance, freq)
```
OK with me to leave as-is though

rdtools/test/test_normalize_with_expected_power.py
```diff
@@ -0,0 +1,109 @@
...
+	return pd.Series([1.2, 2.8, 2.0], index=times_30)
+
+@pytest.fixture()
+def irradiance_30(times_30):
+	return pd.Series([1000.0, 950.0, 890.0], index=times_30)
+
+def test_normalize_with_expected_power_uniform_frequency(pv_15, expected_15, irradiance_15):
+	norm, insol = normalize_with_expected_power(pv_15, expected_15, irradiance_15)
+	expected_norm = pd.Series({Timestamp('2020-01-01 12:15:00', freq='15T'): 1.0,
+							   Timestamp('2020-01-01 12:30:00', freq='15T'): 1.0784313725490198,
```
kanderso-nrel, line 42: Indentation, also I have never seen this style of creating a Series before -- interesting

rdtools/test/test_normalize_with_expected_power.py
```diff
@@ -0,0 +1,109 @@
...
+                               Timestamp('2020-01-01 12:15:00', freq='15T'): 5.714285714285714,
+                               Timestamp('2020-01-01 12:30:00', freq='15T'): 4.705882352941177,
+                               Timestamp('2020-01-01 12:45:00', freq='15T'): 3.5918367346938775,
+                               Timestamp('2020-01-01 13:00:00', freq='15T'): 4.097560975609756})
+	expected_norm.name = 'energy_Wh'
+
+	expected_insol = pd.Series({Timestamp('2020-01-01 12:15:00', freq='15T'): 231.25,
+ 								Timestamp('2020-01-01 12:30:00', freq='15T'): 225.0,
+ 								Timestamp('2020-01-01 12:45:00', freq='15T'): 240.625,
+ 								Timestamp('2020-01-01 13:00:00', freq='15T'): 233.125})
```
kanderso-nrel, line 68: From a user's perspective, it is a little strange that the returned normalized energy and insolation series have different lengths when `pv_input='energy'`.  Should we remove the leading nan on the normalized energy, or add one on the insolation?

Also, indentation for `expected_insol`

rdtools/test/test_normalize_with_expected_power.py
mdeceglie: Thanks for catching the formatting. Cleaning that up. I got that from `pd.Series.to_dict()` it seemed to be the most portable from the notebook I used to prototype the tests...

rdtools/normalization.py
mdeceglie: Agreed, good catch

rdtools/test/test_normalize_with_expected_power.py
mdeceglie: That's a good point. I can imaging a bunch of edge cases here resulting in different behavior, I think the most conservative thing to do is to reindex both series to the union of their indexes.

mdeceglie:
Thanks for the review @kanderso-nrel I think I've incorporated all your suggestions.

PR #174, opened 2020-05-08
add the logo to documentation home page

kanderso-nrel:
Comments:
- This uses the horizontal version
- Generally it's good to reduce the image dimensions to save bandwidth if it's not going to be displayed at full res, but the full image is only 330KB so I didn't do that here.
- Sphinx allows a logo at the top of the sidebar too ([example](https://docs.blender.org/manual/en/latest/)), so we could put the logo there in addition to/instead of where it currently is on the home page.

Looks like this when I build locally:

image


mdeceglie:
I think that looks great. Good to leave it higher resolution, that way if people grab it from here for presentations etc. better chance it will look good.

I could be convinced either way on the sidebar.

mdeceglie:
LGTM, shall we go ahead and merge @kanderso-nrel ?

kanderso-nrel:
Fine with me to merge as-is 👍

mikofski:
Kevin, why not add the new logo to the README as well? BTW: super cool! It looks great!

mdeceglie:
Good idea! Opened #178

Issue #177, opened 2020-05-13
data availability 


yonsa22:
Hello,
could you please refer me to where the data is deposited
thank you for your work

kanderso-nrel:
Hi @yonsa22, if you are asking about the Desert Knowledge Australia Solar Centre data used in the example notebook here: https://github.com/NREL/rdtools/blob/master/docs/degradation_example.ipynb
It looks like the DKA website has changed its structure and the URL in the notebook doesn't work anymore.  The data is still available though, see Site 12 on this page: http://dkasolarcentre.com.au/download?location=alice-springs

Let's keep this issue open until we fix the URL in the example notebook.

yonsa22:
Dear Kevin,
Thank you for your response, I was looking for a run to failure data
(reliability) of PV inverters
is there any publicly available data of that kind?

thank you for your help

Yonatan Saadon

kanderso-nrel:
RdTools does not currently have any functions to evaluate inverter reliability, although we do plan to add such functions in the future.  I'm not aware of any publicly-available large-scale datasets intended for inverter reliability analysis, but maybe there are some.  Perhaps this paper is of interest to you:

> Jordan, DC, Marion, B, Deline, C, Barnes, T, Bolinger, M. PV field reliability status—Analysis of 100 000 solar systems. Prog Photovolt Res Appl. 2020; 1– 16. https://doi.org/10.1002/pip.3262

kanderso-nrel:
DKASC links updated in #179 and #180

PR #179, opened 2020-05-29
update DKASC data link in example notebook on master

kanderso-nrel:
#177 brought up that the link to the desert knowledge dataset is broken.  I checked out the new data file and everything is the same, just the url to get it changed.  This PR is to master, should update development too. 

Do we want to add a note in the notebook saying "by the way the URL changed" or anything?


mdeceglie:
I think for this update, not worth cutting a release since it doesn't touch anything distributed through pypi.

Issue #181, opened 2020-06-23
Chage default behavior of interpolate


mdeceglie:
`rdtools.interpolate()` default behavior is to use a `max_timedelta` based on the median timedelta of the input. This can lead to a high fraction of NaNs in the result for some irregular time series. Changing changing the default behavior to 2x the median timedelta would prevent this. It may also help to add a warning when a substantial fraction of returned points are NaN.

PR #182, opened 2020-06-24
Change interpolate max_timedelta to 2x median, add warning

kanderso-nrel:
Addresses #181.  The changes to the tests were more involved than I'd like because the test dataset is so short that excluding any points exceeds the 10% threshold, so I had to increase the threshold for most tests to avoid cluttering the pytest report with warnings.  

~I think this doesn't need a changelog entry because the interpolate functions are new in 2.0, right?  Though the notebook(s) still need to be re-run.~


mdeceglie:
Thanks @kanderso-nrel  I have been erring on the side of putting changes to functions that are new in the change log, because people may have been using functions in their pre-release form and better safe than sorry. I'd suggest adding a note here.

kanderso-nrel:
> @kanderso-nrel can you confirm that you re-ran `degradation_and_soiling_example.ipynb` and it didn't change. (I wouldn't expect it to change because we pass `max_timedelta` in that example.)

Correct, no change in the DKA notebook.

PR #186, opened 2020-07-07
Overview update

mdeceglie:
I suggest slightly rephrasing the overview.


kanderso-nrel:
Fine by me.  That same wording is in README.md, may want to update that as well.

PR #196, opened 2020-09-03
Modelchain pytests

cdeline:
- [x] Updated changelog

Update to modelChain workflow - rename class to RdAnalysis.  Update notebooks to more closely match current best practice.  Start some pytests (admittedly I don't know how to use fixtures, so these can maybe be improved).


cdeline:
Tests appear to be failing due to a pvlib environment issue - Tables not found error in pvlib.lookup_linke_turbidity

>        try:
 >           import tables
 >       except ImportError:
>           raise ImportError('The Linke turbidity lookup table requires tables. '
                              'You can still use clearsky.ineichen if you '
                              'supply your own turbidities.')
>    E           ImportError: The Linke turbidity lookup table requires tables. You can still use clearsky.ineichen if you supply your own > turbidities.
>     ../../../virtualenv/python3.6.7/lib/python3.6/site-packages/pvlib/clearsky.py:190: ImportError
>     =============================== warnings summary ===============================
>     /home/travis/virtualenv/python3.6.7/lib/python3.6/distutils/__init__.py:4

cdeline:
Test coverage for analysis.py is 72%

kanderso-nrel:
We could get `tables` by installing `pvlib[optional]`, but that would also bring in a bunch of other stuff (cython, pvfactors, pysam...).  If it's necessary for the clear-sky workflow, should it be added to our setup.py as a requirement?

Possibly relevant: https://github.com/NREL/rdtools/issues/194

rdtools/analysis.py
```diff
@@ -207,26 +203,38 @@ def calc_cell_temperature(self, poa, windspeed, ambient_temperature):
         numeric
             calculated cell temperature
         '''
-        if self.temperature_model is None:
+        from packaging import version
+        
+        if version.parse(pvlib.__version__) < version.parse('0.7'):
```
kanderso-nrel, line 208: #170 bumped our pvlib requirement to >=0.7.0, so I think this tapdancing around for the old pvlib API isn't necessary.

cdeline: Thanks.  I have it in here for the time being because pvfleets_qa is still on pvlib 0.6.3, but I agree that this isn't strictly necessary based on our requirements.

cdeline:
OK, I'm up to 95% test coverage which includes all functionality except for some error messages. Still need to update Sphinx docs and release notes if we're happy with the way this is looking.

cdeline:
Notebook comparison is close but not exactly the same.
RdAnalysis p50 rd: -0.428  ci: [-0.5395 -0.3267];   Original notebook: -0.4335  ci: [-0.523 -0.331]
Clearsky RdAnalysis: -0.1039 ci: [-0.289  0.0788];  Original notebook: -0.0933. ci: [-0.245  0.0601]
SRR RdAnalysis: 0.9757  ci: [0.9699  0.980];           Original notebook: 0.973.     ci: [0.965  0.979]

cdeline:
There's some difference in filtering from the old notebook to the new modelchains. Both start with 895104 values in the dataset. 
RdAnalysis sensor_filter.__len__(): 291180   clearsky_filter.__len__(): 244907
Original notebook filtered.__len__291671   clearsky filtered.__len__ 244900

cdeline:
OK, looking at the PVDAQ4 example, results are closer for sensor analysis (-0.488 now vs -0.490 previously).  however Clearsky and SRR results are a little different:
New RdAnalysis  clearsky: -0.652  ci: [-0.777 -0.506]
Old example clearsky:      -0.569   ci: [-0.792 -0.410]
New SRR:  0.923   ci:   [0.846   0.965  ]
Old SRR:  0.945    ci:   [0.940    0.951  ]

cdeline:
Looks like the SRR differences are because of the synthetic soiling signal used for PVDAQ4.  I pushed a new example using that, and results are pretty close.  The only major difference is the Clearsky results for the Sanyo example: -0.652 vs -0.569

cdeline:
Here's a final comment on the difference in median values between the RdAnalysis and Notebook examples for Sanyo (pvdaq 4): 
- For Sensor-based analysis, there's a 0.0005% difference in the normalize_with_PVWatts output, due in part to a slight difference in initial _interpolation_ between the two methods.  In one case, the entire dataframe is interpolated as a whole. In the other, individual series are interpolated.
- This leads to a 0.006% difference in filtering as well.
- Overall sensor-based difference is 0.0032%/yr in Rd calculation

For clearsky analysis, after I updated the way in which clearsky_poa is calculated in the example notebook, there is still:
- 0.04% difference in calculation of clearsky_poa, and a
- 3% difference in calculating clearsky_filter
- This leads to a 0.05%/yr difference in Rd calculation, split roughly equally between the two factors.

For confidence interval calculation, the CI values are within the 0.02 %/yr natural variation that you get from changing the np.random.seed.

cdeline:
OK, got sensor-based difference down to zero. It turns out that there was a missing `max_timedelta='15T'`  on the PVDAQ4 example notebook.  I included this now to make it consistent with the DKCAS and RdAnalysis example.

cdeline:
These new pytests, sphinx docs and examples are ready to merge into ModelChains.

rdtools/test/analysis_test.py
```diff
@@ -0,0 +1,246 @@
...
+                        temperature_coefficient=0 )
+    # get clearsky expected
+    temp.clearsky_preprocess()  # get expected clear-sky values in clearsky_poa.
+    cs = temp.clearsky_poa
+    
+    rdCS = analysis.RdAnalysis(power*cs, cs, temperature_coefficient=meta['tempco'],
+            pvlib_location=loc, pv_tilt=meta['tilt'], pv_azimuth=meta['azimuth'], 
+            temperature_model = {'a': -3.47, 'b': -0.0594, 'deltaT': 3}
+            )
+    return rdCS
```
kanderso-nrel, line 121: What is the intent for this fixture?  If I understand it correctly, it puts together an RdAnalysis object where the measured POA is simulated clear-sky POA.  Why is that useful?

rdtools/test/analysis_test.py
```diff
@@ -0,0 +1,246 @@
...
+            clearsky_poa=cspoa, clearsky_ambient_temperature=csamb, pv_input='energy', 
+            pvlib_location=loc, pv_tilt=pd.Series(data=meta['tilt'], index=pv.index), 
+            pv_azimuth=pd.Series(data=meta['azimuth'], index=pv.index),
+            temperature_model = {'a': -3.47, 'b': -0.0594, 'deltaT': 3}, interp_freq='1H'
+            )
+    rdCS2.pv_power = getRdCS.pv_power
+    rdCS2.clearsky_analysis()
+    cs_yoy_results2 = rdCS2.results['clearsky']['yoy_degradation']
+    assert -5.1278 == pytest.approx(cs_yoy_results2['p50_rd'], abs=1e-3)
+    assert [-5.1285, -5.1269] == pytest.approx(cs_yoy_results2['rd_confidence_interval'], abs=1e-3)
```
kanderso-nrel, line 171: Also confused about this one.  What does this recycling approach do?  (also why is the returned Rd so crazy)

kanderso-nrel (commented):
Some comments/questions, only one or two of which are actually important. 

Another note: the ad-hoc filter section in the two new notebooks have two typos: `... consecuatively, which can be associated with faulty measurments...`.

rdtools/analysis.py
cdeline: I removed this and replaced with exception handling for AttributeError (no python module named 'temperature') which is raised for pvlib < 0.7

rdtools/analysis.py
```diff
@@ -207,26 +204,38 @@ def calc_cell_temperature(self, poa, windspeed, ambient_temperature):
         numeric
             calculated cell temperature
         '''
-        if self.temperature_model is None:
+        from packaging import version
```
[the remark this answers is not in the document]
cdeline: I removed this and replaced with exception handling for AttributeError (no python module named 'temperature') which is raised for pvlib < 0.7

docs/sphinx/source/changelog/v2.1.0.rst
```diff
@@ -0,0 +1,52 @@
...
+Requirements
+------------
+* tables=3.6.1 now added to requirements.txt and setup.py (:pull:`196`).
+
+Example Updates
+---------------
+* New example notebook based on PVDAQ system #4 for the new :py:class:`~rdtools.analysis.RdAnalysis`
+  analysis workflow (:pull:`196`).
+* Update the standard PVDAQ system #4 workbook to match best practice, including 
+  `pvlib.get_total_irradiance()`` and `max_timedelta = '15T'` in rdtools.interpolate (:pull:`196`).
```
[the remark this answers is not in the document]
cdeline: Fixed.

docs/sphinx/source/changelog/v2.1.0.rst
```diff
@@ -0,0 +1,52 @@
...
+v2.1.0 (October XX, 2020)
+**************************
+
+API Changes
+-----------
+* Add :py:class:`~rdtools.analysis.RdAnalysis` class for single-line analysis. (:pull:`117`).
+* Add :py:class:`~rdtools.analysis.RdAnalysis` class run functions  called :py:func:`~rdtools.analysis.sensor_analysis`
+  and :py:func:`~rdtools.analysis.RdAnalysis.clearsky_analysis`. (:pull:`117`).
+* Run SRR analysis by including 'srr_soiling' option when calling 
+  :py:func:`~rdtools.analysis.RdAnalysis.sensor_analysis` (:pull:`117`).
```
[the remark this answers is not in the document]
cdeline: I've cleaned this up. Feel free to double check.

rdtools/analysis.py
```diff
@@ -5,18 +5,15 @@
...
+from rdtools import soiling, clearsky_temperature, plotting
 
 
-class SystemAnalysis():
+
+class RdAnalysis():
     '''
-    Class for end-to-end analysis
+    Class for end-to-end degradation and soiling analysis using :py:class:`~rdtools.RdAnalysis.sensor_analysis`
+    or :py:class:`~rdtools.RdAnalysis.clearsky_analysis`
```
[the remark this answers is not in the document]
cdeline: Done.

docs/sphinx/source/api.rst
```diff
@@ -22,6 +23,16 @@ analysis workflow.
...
 
+Analysis
+===========
+
+Conduct end-to-end Degradation and Soiling analysis with sensor_analysis or clearsky_analysis
+
+.. autosummary::
+   :toctree: generated/
+
+   analysis.RdAnalysis
```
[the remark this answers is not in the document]
cdeline: OK, I've included this.  Thanks for your expertise on autodocs, I don't know how any of this works.

rdtools/test/analysis_test.py
```diff
@@ -0,0 +1,246 @@
...
+meta = {"latitude": -23.762028,
+    "longitude": 133.874886,
+    "timezone": 'Australia/North',
+    "tempco": -0.005,
+    "azimuth": 0,
+    "tilt": 20}
+
+"""
+#########
+##  THE FOLLOWING ARE OLD VERSIONS OF TESTS NOT USING FIXTURES. CAN BE REMOVED
```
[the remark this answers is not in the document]
cdeline: Yes, I've removed these old sections now.

rdtools/test/analysis_test.py
cdeline: OK, hear me out.  when you import the get_energy degradation test fixture, it's just a single downhill slope over 3 years. The clearsky filtering won't work if there's no daily irradiance.    Once you have the clearsky_poa by running through clearsky_preprocess() once then you can multiply cs_poa by the degradation signal to get something close enough to a real daily power and irradiance change that can pass through the clearsky analysis without getting rejected.

rdtools/test/analysis_test.py
cdeline: Maybe there's a better way to do this, but lines 147 and 148 return the clearsky degradation fixture which is based on passing in power and poa.  That only exercises a couple of the inputs into the clearsky analysis - you can also pass in energy, cs_poa, and cs_temperature.  Lines 154-169 are just increasing test coverage by calculating those values from the first run, and passing them back in.  As for why the calculated value (-5.13) differs from the first run (-4.75), based on ostensibly consistent input values, that's probably something worth looking into.

cdeline: OK, I checked this out, and the two assertion values are the same if you use the same temperature coefficient.  I added that in and changed the test value to match, just to avoid confusion.  Testing a different temperature coefficient could be done in a separate test.

kanderso-nrel: Ok thanks for the explanations, I get what's going on now.  The pattern of passing around the object you're testing (rather than arguments to the object) is a new idea to me.  Do you think it makes more sense here than how the [soiling tests](https://github.com/NREL/rdtools/blob/development/rdtools/test/soiling_test.py) work, where all the fixtures are parameters to the function/object?  If it works it works, I'm just thinking about how easy it will be to figure out again when we haven't looked at it for a few months.

kanderso-nrel: @cdeline I played around with rewriting these tests in the style I was talking about above.  I'm not convinced I like it better than what you're doing here, but am curious what you think: https://github.com/NREL/rdtools/blob/modelchains_pytests_anderson/rdtools/test/analysis_test.py

cdeline: OK, thanks.  I like your method, and think that it hews more closely to the intent of pytest fixtures, so let's go with that.  I'll bring it into my branch.

docs/sphinx/source/changelog/v2.1.0.rst
```diff
@@ -0,0 +1,54 @@
...
+-----------
+* Add :py:class:`~rdtools.analysis.RdAnalysis` class for single-line analysis. (:pull:`117`).
+
+
+Enhancements
+------------
+* Add new :py:mod:`~rdtools.analysis` module to focus on single-line analysis workflow
+  (:pull:`117`).
+* Add :py:class:`~rdtools.analysis.RdAnalysis` class for single-line analysis. 
+  Degradataion analysis is run by :py:meth:`~rdtools.analysis.RdAnalysis.sensor_analysis`
```
kanderso-nrel, line 15:
```suggestion
  Degradation analysis is run by :py:meth:`~rdtools.analysis.RdAnalysis.sensor_analysis` 
```

I make this typo all the time, I guess because our fingers are used to typing "data".

kanderso-nrel (approved):
Looks like I didn't rewrite the tests perfectly; the `test_clearsky_analysis_optional` test doesn't cover the interpolation lines in the RdAnalysis constructor because the parameters don't include `interp_freq`.

Do we care about testing every single error?  If not, looks good to me, one small typo below.

cdeline:
Great, thanks.  we're still at 91% coverage on analysis.py which is good enough for me.  I'm going to hit go - we can still make changes in PR #117.

Issue #194, opened 2020-08-25
Travis config not fully consistent with setup.py


kanderso-nrel:
Here are two proposed improvements to our [.travis.yml](https://github.com/NREL/rdtools/blob/development/.travis.yml).

---

First, Travis doesn't run `pip install rdtools[test]` during the job setup.  It's not currently a problem, I guess because Travis happens to supply the current requirements (like pytest) out of the box, so the tests pass.  But if we add a new requirement to `[test]` (for example, `tables` for #131), the requirements aren't met and the tests fail.  

I think changing these lines to `REQ_ENV='<...> .[test]'` will have the desired effect:
https://github.com/NREL/rdtools/blob/8f00cd4b75ba5f69cd4057496a21aeeb51e8607b/.travis.yml#L14-L16

---

Second, the minimum dependency versions specified in our setup.py aren't tested in the CI which makes it more likely for us to unintentionally break compatibility with the minimum supported versions.  I propose either: 1) changing requirements.txt to match our minimum supported versions, or 2) creating a `requirements-min.txt` with the minimum versions specified in setup.py and adding a third Travis `env` entry to use it for tests.  I'd prefer option 2 because `requirements.txt` is user-facing and I don't really think we should be encouraging people to use the oldest package versions possible.

cdeline:
OK, I agree with this approach.  I'll create a feature branch to test these and merge into development to check.  I can merge back into modelchain pytests to see if it fixes things.

cdeline:
Actually, on second thought I'm out of my depth here.  I'll leave it to you @kanderso-nrel, but I agree with this approach.

mdeceglie:
Closed with #214

PR #197, opened 2020-09-04
Update Travis config for test requirements and minimum requirements; bump numpy and matplotlib requirements

kanderso-nrel:
- [x] Updated changelog

See #194


kanderso-nrel:
I think this addresses the two points in #194.  I bumped the minimum matplotlib version to 3.0.0 (released Sep 18, 2018) from 2.2.2 (Mar 17, 2018) because 2.2.2 was causing the plotting tests to fail in the CI environment -- it was trying to spawn tkinter windows without any display available to host them.  Is that acceptable?

cdeline (approved):
Tests are passing, LooksGoodToMe.

mdeceglie:
The only downside I see is that we could accidentally put something in the `test` requirements in `setup.py` that is actually needed for the library itself, and we wouldn't notice. I guess I'm not that worried about it, because as soon as someone notices such a bug, it is a quick fix. But just wanted to add this to the discussion.

mdeceglie:
Now that I'm scrutinizing `requirements.txt` I see it has some surprising entries like `sphinx-rtd-theme`. Here is my result of `pip freeze`  after running `pip install .` in a clean python 3.7 environment:
```
certifi==2020.6.20
chardet==3.0.4
cycler==0.10.0
h5py==2.10.0
idna==2.10
kiwisolver==1.2.0
matplotlib==3.3.2
numpy==1.19.2
pandas==1.1.2
patsy==0.5.1
Pillow==7.2.0
pvlib==0.7.2
pyparsing==2.4.7
python-dateutil==2.8.1
pytz==2020.1
requests==2.24.0
scipy==1.5.2
six==1.15.0
statsmodels==0.12.0
urllib3==1.25.10
```

kanderso-nrel:
> we could accidentally put something in the test requirements in setup.py that is actually needed for the library itself, and we wouldn't notice

That's a good point -- this change would only give us confidence that `rdtools[test]` passes the tests, but really we want to know about `rdtools`.  Right now the difference between `rdtools` and `rdtools[test]` is only pytest and coverage, which is fine, but if we start adding other things like `tables` like I previously wanted to do for #131, it could get tricky.  OK with me if we just agree to be careful about dependencies, but here are a couple other ideas:

1) Instead of using external packages to generate data at test runtime, we could use them once to generate static datasets (CSV files and such) and read them in for the tests.  A lot of the files in [pvlib/data](https://github.com/pvlib/pvlib-python/tree/master/pvlib/data) are for the test suite.  
2) The notebook(s) could be incorporated into the CI testing.  They are designed to work with just `rdtools` and while they maybe aren't as comprehensive as the pytest suite, they still cover the majority of the package functionality.  Something like [nbval](https://github.com/computationalmodelling/nbval) could be useful for re-running notebooks and verifying that all the new output is equivalent to the output already in the notebook.

---

For the requirements, the sphinx dependencies were introduced in the pv-terms PR: https://github.com/NREL/rdtools/pull/185/commits/d7df0edffb5661044e09236a3cb08268257f12d2
Looks like most of the new requirements in your list (certifi, chardet, idna, urllib3) are from pvlib's relatively new requests requirement.  I think Pillow is needed by something in the sphinx world, but not sure why it's in your list.  Anyway do you think we shouldn't have the docs requirements listed in requirements.txt?

cdeline:
sphinx-rtd-theme is needed for `make html` to generate the sphinx documentation.  I don't think it's mandatory for pv-terms per se, just necessary to install in a bare environment if you want to check that sphinx docs are passing.
What's the downside of having it in requirements.txt?

kanderso-nrel:
I would guess that it's a small fraction of RdTools users that are interested in building the docs locally, so the sphinx-related entries in requirements.txt are unnecessary environment bloat for most people. Not a big downside, I could go either way on whether they should be included.  Just guessing, seems like they ought to be pretty light requirements compared with scipy, numpy, pandas etc.

mdeceglie:
My thinking is that folks who want to build docs would be using `pip install .[doc]` option. And if it's trivial to pull a few things out we should to prevent bloat. My thinking is to keep `requirements.txt` as simple as possible to document the minimum working environment. But it isn't a big deal. If it helps, I repeated my test but made sure to install the same versions of rdtools direct requirements as are currently specified in requirements.txt. This was the result (Pillow disappeared):

```
certifi==2020.6.20
chardet==3.0.4
cycler==0.10.0
h5py==2.10.0
idna==2.10
kiwisolver==1.2.0
matplotlib==3.1.2
numpy==1.17.3
pandas==1.0.3
patsy==0.5.1
pvlib==0.7.1
pyparsing==2.4.7
python-dateutil==2.8.1
pytz==2020.1
requests==2.24.0
scipy==1.3.2
six==1.15.0
statsmodels==0.11.1
urllib3==1.25.10
```

kanderso-nrel:
Test failure is because we use `np.quantile`, added in numpy 1.15 but our minimum is 1.12: https://travis-ci.org/github/NREL/rdtools/jobs/735446880#L714

Per [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table), we could actually bump up to 1.16 and still be following the guidelines, but I think we generally feel similarly to https://github.com/pvlib/pvlib-python/issues/828#issuecomment-561721447, so there's no need to drop versions just to follow NEP-29.  With that in mind, I'll bump the minimum to 1.15 here.

mdeceglie (approved):
LGTM

Issue #189, opened 2020-07-30
This line needs to be removed as it causes soiling intervals to be incorrectly determined when precip events are also considered.


matt14muller:
https://github.com/NREL/rdtools/blob/5b18cd7d3694911d201f930a8b21e2abdebabd12/rdtools/soiling.py#L162

kanderso-nrel:
I think this reproduces the problem:

```python
import pandas as pd
from rdtools import soiling
import matplotlib.pyplot as plt

N = 30
idx = pd.date_range('2019-01-01', freq='d', periods=N)
data = pd.DataFrame({
    'energy_normalized_daily': 1.0,
    'insolation_daily': 1.0,
    'precipitation_daily': 0.0,
}, index=idx)
data.loc[idx[5], 'precipitation_daily'] = 0.5
data.loc[idx[10:20], 'precipitation_daily'] = 0.5

srr = soiling.SRRAnalysis(**data)
srr._calc_daily_df(clean_criterion='precip_or_shift')
srr._calc_result_df()

fig, ax = plt.subplots(3, 1, sharex=True)
data['precipitation_daily'].plot(drawstyle='steps', ax=ax[0])
srr.daily_df['clean_event'].astype(int).plot(drawstyle='steps', ax=ax[1])
srr.analyzed_daily_df['run'].plot(drawstyle='steps', ax=ax[2])
for a in ax:
    a.legend()
```

![image](https://user-images.githubusercontent.com/57452607/89787446-b2759f00-dada-11ea-90fe-26106c2e7e72.png)

A string of multiple rainy days in a row only triggers a clean event on the last day of the string, meaning all the previous days get included in the previous run.  But presumably the first rainy day would be sufficient for a cleaning, so the previous run should end on the first day of the rain string, not the last.  Should each day in the rain string be its own run, and then it would get filtered out later for being too short?

kanderso-nrel:
Here is the same plot, but running RdTools with that line 162 removed:

![image](https://user-images.githubusercontent.com/57452607/89812563-8a4b6780-dafd-11ea-968e-f413863cfe82.png)

mdeceglie:
[...]

Issue #113, opened 2019-04-16
Consider warning for non-daily inputs to srr_analysis.__init__()


mdeceglie:
@GregoryKimball raised the question in PR #112 whether `soiling.srr_analysis.__init__()` should raise warnings instead of the current behavior to raise a`ValueError` when called with Pandas series that don't have the frequency set. 

(At time of entry, this issue does not apply to `master`)

mdeceglie:
Let's use `infer_freq()` to check wether it is truly daily and allow the code to proceed

martin-springer:
Closing as implementation was verified by copilot. 

Issue https://github.com/NatLabRockies/rdtools/pull/112 requested SRRAnalysis.__init__() use pd.infer_freq() to validate truly daily data instead of rejecting series without explicit frequency attributes.

Investigation confirms this is already implemented. Current code at lines 69-80 of soiling.py:

if pd.infer_freq(self.pm.index) != 'D':
    raise ValueError('Daily performance metric series must have daily frequency')
pd.infer_freq() returns 'D' for daily data even when .freq=None, allowing implicit daily series to proceed while rejecting truly non-daily data.

Verified by:

Existing test test_soiling_srr_non_daily_inputs validates implicit daily series pass
All 35 soiling tests pass
Manual testing confirms correct behavior for explicit freq, implicit freq, and non-daily cases
No code changes required.

PR #199, opened 2020-09-09
Consecutive cleaning logic

mdeceglie:
Update the cleaning detection logic so that consecutive cleaning events are not automatically removed. For a full discussion see #189 

- [x] Updated changelog


rdtools/soiling.py
```diff
@@ -85,6 +87,11 @@ def _calc_daily_df(self, day_scale=14, clean_threshold='infer',
             The daily precipitation threshold for defining precipitation cleaning events.
             Units must be consistent with ``self.precipitation_daily``.
         '''
+        if (day_scale % 2 == 0) and ('shift' in clean_criterion):
+            warnings.warn('An even value of day_scale was passed. An odd value is '
+                          'recommended, otherwise, consecutive days may be erronsiously '
+                          'flagged as cleaning events. '
+                          'See https://github.com/NREL/rdtools/issues/189')
```
kanderso-nrel, line 94: Should we be specifying a warning type?  The default is UserWarning, which I think is appropriate, but wanted to bring it up for discussion.  See https://docs.python.org/3/library/warnings.html#warning-categories

mdeceglie: I agree the default `UserWarning` is appropriate. Do you think we should be explicit? Or OK to rely on default?

kanderso-nrel: Fine with me to use the default 👍 I was wondering if emitting the default warning is like raising the base `Exception` class instead of a more specific exception, but I think UserWarning is fine.

kanderso-nrel:
Not sure what that RTD build failure is about; this branch builds fine locally for me.  Hopefully merging development will fix it.

mdeceglie:
To do while we're touching the soiling module: address #113 with a more helpful error message. 

Also, some of these updates should probably got to `master`

docs/sphinx/source/changelog/pending.rst
```diff
@@ -1,6 +1,30 @@
...
-*******
+************************
+
+API Changes
+-----------
+* The calculations internal to the SRR algorithm have changed such that consecutive
+  cleaning events are no longer removed. (:pull:`199`, :issue:`189`)
+
+* The default ``day_scale`` parameter in soiling funtions and methods was changed
+  from 14 to 13. (:pull:`199`, :issue:`189`)
```
kanderso-nrel, line 11: New `outlier_factor` parameter as well

rdtools/soiling.py
```diff
@@ -343,6 +351,17 @@ def _calc_monte(self, monte, method='half_norm_clean'):
...
+        # Raise a warning if there is >20% invalid data
+        if (method == 'half_norm_clean') or (method == 'random_clean'):
+            valid_fraction = self.analyzed_daily_df['valid'].sum()/len(self.analyzed_daily_df)
+            if valid_fraction <= 0.8:
+                warnings.warn('20% or more of the daily data is assigned to invalid soiling '
+                              'intervals. This can be problematic with the "half_norm_clean" '
+                              'and "random_clean" cleaning assumptions. Consider more permissive '
+                              'validity criteria such as increasing "max_relative_slope_error" '
+                              'and/or "max_negative_step" and/or decreasing "min_interval_length".'
+                              ' Alternatively, consider using method="perfect_clean".'
```
kanderso-nrel, line 363: Is this in reference to #234?  Might be good to add that breadcrumb in case folks are interesting in what exactly is problematic.

Also, should the 20% threshold be exposed as an optional parameter?

rdtools/soiling.py
```diff
@@ -343,6 +351,17 @@ def _calc_monte(self, monte, method='half_norm_clean'):
               the interval.
         '''
 
+        # Raise a warning if there is >20% invalid data
+        if (method == 'half_norm_clean') or (method == 'random_clean'):
+            valid_fraction = self.analyzed_daily_df['valid'].sum()/len(self.analyzed_daily_df)
```
kanderso-nrel, line 356: `valid_fraction = self.analyzed_daily_df['valid'].mean()` may be easier to read, up to you

rdtools/test/soiling_test.py
```diff
@@ -57,12 +57,34 @@ def test_soiling_srr(soiling_normalized_daily, soiling_insolation, soiling_times
...
+    sr, sr_ci, soiling_info = soiling_srr(soiling_normalized_daily, soiling_insolation, reps=reps,
+                                          max_relative_slope_error=20.0, method='half_norm_clean')
+    assert 0.915093 == pytest.approx(sr, abs=1e-6),\
+        'Soiling ratio different from expected value for half_norm_clean with consecutive invalid intervals'  # noqa: E501
+
+    np.random.seed(1977)
+    sr, sr_ci, soiling_info = soiling_srr(soiling_normalized_daily, soiling_insolation, reps=reps,
+                                          max_relative_slope_error=20.0, method='perfect_clean')
+    assert 0.977116 == pytest.approx(sr, abs=1e-6),\
+        'Soiling ratio different from expected value for perfect_clean with consecutive invalid intervals'  # noqa: E501
```
kanderso-nrel, line 85: Maybe worth `parametrize`ing (though the same could be said of all the other tests too):
```python
@pytest.mark.parametrize('method,expected_sr',
                         [('random_clean', 0.936177),
                          ('half_norm_clean', 0.915093),
                          ('perfect_clean', 0.977116)])
def test_soiling_srr_consecutive_invalid(soiling_normalized_daily, soiling_insolation,
                                         soiling_times, method, expected_sr):
    reps = 10
    np.random.seed(1977)
    sr, sr_ci, soiling_info = soiling_srr(soiling_normalized_daily, soiling_insolation, reps=reps,
                                          max_relative_slope_error=20.0, method=method)
    assert expected_sr == pytest.approx(sr, abs=1e-6),\
        f'Soiling ratio different from expected value for {method} with consecutive invalid intervals'  # noqa: E501
```

rdtools/soiling.py
```diff
@@ -286,12 +295,11 @@ def _calc_result_df(self, trim=False, max_relative_slope_error=500.0,
...
 
-        # Don't consider data outside of first and last valid intervals
         if len(results[results.valid]) == 0:
             raise NoValidIntervalError('No valid soiling intervals were found')
-        new_start = results[results.valid].start.iloc[0]
-        new_end = results[results.valid].end.iloc[-1]
-        pm_frame_out = daily_df[new_start:new_end]
+        new_start = results.start.iloc[0]
+        new_end = results.end.iloc[-1]
+        pm_frame_out = daily_df[new_start:new_end].copy()
```
kanderso-nrel, line 302: `pm_frame_out.reset_index()` on the next line returns a copy, so making another one here is unnecessary

rdtools/soiling.py
```diff
@@ -286,12 +295,11 @@ def _calc_result_df(self, trim=False, max_relative_slope_error=500.0,
             results.next_inferred_start_loss - results.inferred_end_loss,
             0, 1)
 
-        # Don't consider data outside of first and last valid intervals
         if len(results[results.valid]) == 0:
             raise NoValidIntervalError('No valid soiling intervals were found')
-        new_start = results[results.valid].start.iloc[0]
-        new_end = results[results.valid].end.iloc[-1]
```
kanderso-nrel, line 293: Not clear to me how this change relates to the others.  If it changes the outputs, does it need a changelog entry?

docs/sphinx/source/changelog/pending.rst
```diff
@@ -1,6 +1,30 @@
...
+
+Enhancements
+------------
+
+
+Bug fixes
```
kanderso-nrel, line 17: Mention #234 if this PR addresses that too

kanderso-nrel (commented):
Partial review.  Just an observation: I had a look at the PVDAQ4 notebook and noticed that the P50 soiling rate change is significant compared with the CI width.

docs/sphinx/source/changelog/pending.rst
```diff
@@ -1,6 +1,30 @@
-*******
+************************
 Pending
-*******
+************************
+
+API Changes
+-----------
+* The calculations internal to the SRR algorithm have changed such that consecutive
+  cleaning events are no longer removed. (:pull:`199`, :issue:`189`)
```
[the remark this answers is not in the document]
mdeceglie: The original behavior wasn't a bug, but rather our thinking on the preferred behavior has evolved. I'd prefer to leave it here unless we want to coin a new section "calculation changes" or something?

rdtools/soiling.py
mdeceglie: Thanks, yes this is an important change. It retains invalid intervals at beginning and end of the dataset

rdtools/soiling.py
mdeceglie: It's more related to the retention of invalid intervals at the start and end. I will open an issue on that to leave a picture for folks.

mdeceglie: Since it's only a warning, I don't think we need to expose another parameter. If someone is in the weeds enough to want to change that, they'd easily be able to detect whatever fraction they are worried about. This is more for folks who wouldn't have it on their radar.

rdtools/test/soiling_test.py
mdeceglie: Good suggestion, I've done this to a few of the tests.

mdeceglie:
Alright, I think this is ready to go. Any outstanding items to address?

kanderso-nrel (approved):

PR #200, opened 2020-09-11
Added new geometric clipping filter + unit test.

kperrynrel:
- [X] Code changes are covered by tests
- [x] Updated changelog


kperrynrel:
@kanderso-nrel @cdeline @matt14muller @mdeceglie PR for clipping filter.

cdeline:
I'm trying this functionality out currently.  One problem I'm having is that NaN values appear to be removed from the index.  This is an issue if you try to mask by the output of the geometric clipping filter.  For instance: I pass in a 120k length df with the first 20k values being NaN.  The filter boolean mask coming out has a length of 100k.  This appears to happen after your comment '#Outlier removal statement'.  It would be better if your boolean output is guaranteed to have the same length as the input power_ac

kperrynrel:
@cdeline I can make that change here today, will include a default of False  for any of the filtered items on the mask.

kperrynrel:
@cdeline @mdeceglie  @kanderso I gave the matplotlib build a shot for interactive mode--this is what I came up with: 

image

image

It does save as an html and it is interactive (requires the mpld3 package to do this) but the zoom quality is not amazing. If anyone has suggestions on how to improve this, let me know. Otherwise, I've kept things with plotly as the graphics are significantly better, and just added plotly as a package for the time being to the requirements.txt file; the builds look good with this.

kanderso-nrel:
Just thinking about the plotting function, sounds like either way (plotly or mpld3) this functionality would require a new dependency.  Looks like plotly is about as big a dependency as numpy or pandas, so not as big as scipy but not small either. 

Taking a step back for a moment, is the motivation to use plotly (or mpld3) to be able to save the clipping figure to HTML for later viewing?  That functionality seems useful for our internal processes, but I wonder if it necessarily needs to live in RdTools instead of one of our internal packages.

If we do think it's worth keeping in RdTools, we should consider if the other plotting functions could be improved by using plotly instead of mpl.

kperrynrel:
@kanderso I agree that we may want to take a step back regarding how we want to handle plotting, and whether it should be matplotlib only or plotly only (or a mixture of both??). Obviously I do have a bias as I think the graphics on plotly look amazing compared to those in matplotlib, but I don't want that to be a deciding factor here. mpld3 saves matplotlib where it can be accessed as an interactive html for later use. I've done a little bit of playing around with matplotlib in interactive mode in jupyter notebook, but generally it does require dependencies (I tried it with the ipympl package add on, where you can declare matplotlib as a widget--see https://towardsdatascience.com/how-to-produce-interactive-matplotlib-plots-in-jupyter-environment-1e4329d71651). Generally I've found that no matter what, we will probably be adding some new package as a dependency to get interactive plotting. Jupyter is also only place interactive mode will show for matplotlib--it doesn't work in spyder.

cdeline:
My opinion- matplotlib is not terribly user-friendly, and doesn't do user-interactive behavior very well.  Plotly sounds like it's a good solution for interactive plots and I'd like to get to know how to use it better.

rdtools/test/plotting_test.py
```diff
@@ -162,3 +165,51 @@ def test_soiling_rate_histogram_kwargs(soiling_info):
...
+    Return results of clipping filter applied to a degradation signal.
+
+    Returns
+    -------
+    signal_filtered: Pandas series, filtered degradation power signal
+    clipping_mask_series: Pandas series, boolean mask time series for 
+    clipping, with True indicating a clipping period and False 
+    representing a non-clipping period
+    '''
+    signal_filtered, clipping_mask_series = geometric_clip_filter(clipping_power_degradation_signal)
```
kanderso-nrel, line 191: @kperry I think the travis error is from `clipping_power_degradation_signal` vs `degradation_power_signal`

rdtools/test/plotting_test.py
```diff
@@ -162,3 +165,51 @@ def test_soiling_rate_histogram_kwargs(soiling_info):
...
+
+
+def test_clipping_filter_plots_kwargs(clipping_info):
+    signal_filtered, clipping_mask_series = clipping_info
+
+    # test kwargs
+    kwargs = dict(
+        display_web_browser = True
+    )
+    result = tune_clip_filter_plot(clipping_power_degradation_signal,
```
kanderso-nrel, line 212: Probably need to add `clipping_power_degradation_signal` as a parameter of this test function too

kperrynrel: Ah, will give that a shot. Thanks for the tip @kanderso-nrel I've been trying to debug this for the past twenty minutes trying to figure out why the Travis CI build was failing

docs/sphinx/source/api.rst
```diff
@@ -97,6 +97,7 @@ Filtering
    :toctree: generated/
     
    clip_filter
+   geometric_clip_filter
```
kanderso-nrel, line 100: Update with the other functions too

kanderso-nrel (commented):
I didn't step through the code in detail to understand every step yet, but here's a quick first pass.  At first glance it seems like there's a fair bit of similar code in the two new `filtering` functions -- would it make sense to extract the common code into a private function?

Test coverage is pretty good but there are some untested lines -- see here for how we use the `coverage` package to see which lines are untested: https://rdtools.readthedocs.io/en/stable/developer_notes.html#running-the-test-suite

Also were the changes to `soiling.py` made to satisfy the 80-char limit for pep8?  If so, sorry that I didn't mention this before, but we decided to relax it to 99-chars for RdTools (see [here](https://github.com/NREL/rdtools/blob/master/.flake8#L5-L6)).  I think it should be easy to just revert commit 6d7f794 for that.

docs/sphinx/source/changelog/pending.rst
```diff
@@ -2,6 +2,11 @@
 Pending
 *******
 
+Enhancements
+------------
+* Add new function :py:func:`~rdtools.filtering.geometric_clip_filter` (:pull:`200`).
```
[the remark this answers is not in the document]
kperrynrel: @kanderso-nrel I did end up updating the soiling script because it was failing pep8 protocol. I can pull back a commit for that. Thanks for having a look at this--I'll look at what I can do to clean up code redundancies + unit testing this afternoon.

docs/sphinx/source/api.rst
kperrynrel: Addressed in commit 390762799ae1e298fc56be5c778936c05426941d

docs/sphinx/source/api.rst
```diff
@@ -97,6 +97,8 @@ Filtering
    :toctree: generated/
     
    clip_filter
+   logic_clip_filter
+   xgboost_clip_filter
```
kanderso-nrel, line 101: Also add `tune_clip_filter_plot` to the `Plotting` section.

Side note, I wonder if `clip_filter` ought to be renamed to `quantile_clipping_filter` or something...

rdtools/filtering.py
```diff
@@ -124,3 +140,322 @@ def csi_filter(poa_global_measured, poa_global_clearsky, threshold=0.15):
 
     csi = poa_global_measured / poa_global_clearsky
     return (csi >= 1.0 - threshold) & (csi <= 1.0 + threshold)
+
+
+def format_clipping_time_series(power_ac, mounting_type):
```
kanderso-nrel, line 145: Do you think users will ever want to call `format_clipping_time_series` and `calculate_max_derivative` themselves?  If they're really just for the other functions to use, it's probably a good idea to make these ones private by prepending an underscore to the function name -- that way if we need to change them someday we won't have to cut a major release.

rdtools/filtering.py
```diff
@@ -124,3 +140,322 @@ def csi_filter(poa_global_measured, poa_global_clearsky, threshold=0.15):
...
+                      derivative_cutoff=0.2,
+                      roll_periods=None):
+    '''
+    This filter is a logic-based filter that is used to filter out
+    clipping periods in AC power and AC energy time series.
+    The AC power or energy time series is filtered based on the
+    maximum derivative over a rolling window, as compared to a user-set
+    derivate_cutoff (default set to 0.2).  The size of the
+    rolling window is increased when the system is a tracked system.
+    Parameters
```
kanderso-nrel, line 225:
```suggestion
    rolling window is increased when the system is a tracked system.

    Parameters
```

Sphinx is very particular about whitespace in docstrings -- without this blank line, the first parameter doesn't render correctly: https://rdtools--200.org.readthedocs.build/en/200/generated/rdtools.filtering.logic_clip_filter.html

rdtools/filtering.py
```diff
@@ -124,3 +140,322 @@ def csi_filter(poa_global_measured, poa_global_clearsky, threshold=0.15):
...
+        than once every 30 minutes.
+    Returns
+    -------
+    pd.Series
+        The passed time series, filtered to exclude clipping periods.
+    pd.Series
+        Boolean Series of whether to include the point because it is not
+        clipping.
+        True values delineate non-clipping periods, and False values delineate
+        clipping periods.
```
kanderso-nrel, line 255: Seems like the boolean series is really the primary output here, as it is used to filter more than just the input power series.  Should we remove the filtered time series return value?  It's easy for the user to calculate it themselves, and it would make this function's return value consistent with the existing filtering functions.

rdtools/filtering.py
```diff
@@ -124,3 +140,322 @@ def csi_filter(poa_global_measured, poa_global_clearsky, threshold=0.15):
...
+
+    Returns
+    -------
+    pd.Series
+        The passed time series, filtered to exclude clipping periods.
+    pd.Series
+        Boolean Series of whether to include the point because it is not
+        clipping.
+        True values delineate non-clipping periods, and False values delineate
+        clipping periods.
```
kanderso-nrel, line 393: Same comment as above re: return values and parameter defaults

rdtools/filtering.py
```diff
@@ -124,3 +140,322 @@ def csi_filter(poa_global_measured, poa_global_clearsky, threshold=0.15):
...
+    rolling window is increased when the system is a tracked system.
+    Parameters
+    ----------
+    power_ac : pd.Series
+        Pandas time series, representing PV system power or energy with
+        a pandas datetime index.
+    mounting_type : String
+        String representing the mounting configuration associated with the
+        AC power/energy time series. Can either be "Fixed" or "Tracking".
+        Default set to 'Fixed'.
```
kanderso-nrel, line 233: Not sure it really matters, but the numpy docstring convention is to have default values next to the parameter name and type rather than the description.  So for `mounting_type` it would be `mounting_type: string, default 'Fixed'`.

rdtools/plotting.py
```diff
@@ -256,14 +260,60 @@ def soiling_rate_histogram(soiling_info, bins=None):
...
+    df = pd.DataFrame(power_ac)
+    # Add the clipping mask as a column
+    df['clipping_mask'] = clipping_mask
+    df = df.reset_index()
+    fig = px.scatter(df, x=index_name, y=column_name, color='clipping_mask')
+    # If display_web_browser is set to True, the time series with clipping
+    # is rendered via the web browser.
+    if display_web_browser is True:
+        pio.renderers.default = "browser"
+        fig.show()
```
kanderso-nrel, line 313: I'm always hesitant to modify external defaults like this (mostly because I don't want other packages modifying my defaults!).  What do you think about changing the `display_web_browser` parameter to `renderer="browser"` or something that gets passed through to [fig.show()](https://plotly.com/python-api-reference/generated/plotly.graph_objects.Figure.html#plotly.graph_objects.Figure.show)?  Could also remove the `pio.renderers.default = "browser"` line at the top of this file then too. 

Another option is to do away with that parameter entirely and make a note in the docstring that it'll display in whatever they set `pio.renderers.default` to be.

rdtools/test/filtering_test.py
```diff
@@ -1,96 +1,123 @@
...
+    # Add datetime index to second series
+    time_range = pd.date_range('2016-12-02T11:00:00.000Z',
+                               '2017-06-06T07:00:00.000Z', freq='H')
+    power_datetime_index.index = pd.to_datetime(time_range[:100])
+    # Note: Power is expected to be Series object with a datetime index.
+    return power_no_datetime_index, power_datetime_index
+
+
+def test_logic_clip_filter(generate_power_time_series):
+    ''' Unit tests for geometric clipping filter.'''
```
kanderso-nrel, line 78: this comment and the one in the test below might be out of date

rdtools/test/filtering_test.py
```diff
@@ -1,96 +1,123 @@
...
+def test_xgboost_clip_filter(generate_power_time_series):
+    ''' Unit tests for geometric clipping filter.'''
+    power_no_datetime_index, power_datetime_index = \
+        generate_power_time_series
+    # Test that a Type Error is raised when a pandas series
+    # without a datetime index is used.
+    pytest.raises(TypeError,  xgboost_clip_filter,
+                  power_no_datetime_index)
+    # Expect none of the sequence to be clipped (as it's
+    # constantly increasing)
```
kanderso-nrel, line 100: In addition to testing that the functions don't detect clipping when it isn't present, it might be good to test that they do detect clipping when it is 😄

rdtools/plotting.py
```diff
@@ -256,14 +260,60 @@ def soiling_rate_histogram(soiling_info, bins=None):
...
     return fig
 
 
+def tune_clip_filter_plot(power_ac, clipping_mask, display_web_browser=True):
+    """
+    This function allows the user to visualize a clipping filter in
+    a matplotlib plot, after tweaking the function's different
+    hyperparameters. The plot can be zoomed in on, for an in-depth look at
+    clipping in the AC power time series.
+    Parameters
```
kanderso-nrel, line 277: Probably need an empty line before this too

kanderso-nrel (commented):
Some more quick notes.  Dumb question: is sklearn used only for its `preprocessing.MinMaxScaler`, or is it also needed to load the xgboost model or something?  If it's just for the `MinMaxScaler` then we might consider just DIYing that scaling to avoid a big dependency.

mdeceglie:
I haven't gone through in great detail yet, but two points for us to consider:

1. How about a single `clip_filter` with different methods available?
2. (Related) I think we should make the new filters available in `TrendAnalysis()`

kperrynrel:
@kanderso-nrel just resolved all of the issues that you mentioned above in commit c1913ab1d0ab9bd1e5a31acd2ceee933b7b30c84, including adding units testing for detecting clipped periods. Let me know what you think

kperrynrel:
@mdeceglie I have no problem incorporating this into a master filter with the desired filter passed as a parameter, but this may merit further discussion between the three of us on whether/how we'd like to format this

mdeceglie:
My thinking is the cleanest way to do a master function would be something like `clip_filter(power_ac, method=quantile, kwargs)` where method can be `quantile`, `logic`, or `xgboost`.  The problem with this though is that it would break the existing clip filter for users who use entirely positional arguments and I'm not sure we want a breaking change for this. So we could add a new function with a new name or do something sneaky where if `method` is a float, revert to the quantile method and use it at the `quantile` parameter. @kperrynrel and @kanderso-nrel  what do you think?

kanderso-nrel:
If there is a master clipping function, seems like it would be deserving of the name `clip_filter`, which means some kind of check for the quantile filter behavior during the deprecation period is necessary.  Might be worth thinking if we can rely on duck typing somehow instead of explicit type checks.  If we think we might run into this issue for other functions down the road too, we might consider proactively using [keyword-only arguments](https://www.python.org/dev/peps/pep-3102/).

A master function with `method=` is interesting.  pvlib has many of them (e.g. `irradiance.get_total_irradiance` and the `PVSystem.get_*` methods).  This situation is a little different because usually only one of the underlying pvlib models would get used in a given use case (if you've already called `get_total_irradiance(... method='haydavies')`, there's little reason to call it again with `method='perez'`), but with these clipping filters I think there are good reasons to combine multiple models.  Not sure if that distinction is significant though. 

On that note, how would `TrendAnalysis.filter_params` work with a master function?  I think our current `filter_params` approach of having function names as keys and kwargs as values is not flexible enough to support calling a master function multiple times to combine the different clipping filters, so we'd either have to change how `filter_params` works or make it clear that people need to use the lower-level functions instead of the master wrapper.

mdeceglie:
Those are good points I guess `TrendAnalysis.filter_params` should accept either the master function or the functions for the different methods (meaning we should not make them private) and we should introduce a new `quantile_clip_filter()` function that does the current thing. 

Agreed it can wait for a separate PR. What about integrating the new functions with `TrendAnalysis` though?

kperrynrel:
@mdeceglie @kanderso-nrel I've been looking over the functions the past 10 minutes or so thinking about how we format into a master function. How do we feel about keeping the each individual function so that it can be referenced as-needed by users, but then add one more function ("clip_filter" or something like that), and run through the desired filter logic based on the model selected? For example:


def master_clip_filter(power_ac, params, model = "clip_filter"):
    """
    Master wrapper for running one of the desired clipping filters.
    """
    if model == 'clip_filter':
        clip_mask = clip_filter(power_ac, params)
    elif model == 'xgboost_clip_filter':
        clip_mask = xgboost_clip_filter(power_ac, params)
    elif model == 'logic_clip_filter':
        clip_mask = logic_clip_filter(power_ac, params)
    return clip_mask

This obviously is not functional yet, but it's my first pass at adapting what you guys are thinking.

mdeceglie:
That's along the lines of what I was thinking. I have a few suggested modifications:

1. the master filter should be called 'clip_filter' and the thing that is today called 'clip_filter` should be called `quantile_clip_filter`
2. For backward compatibility, we should check for a special case when someone calls `clip_filter(power_ac, quantile)` using positional syntax. @kanderso-nrel  suggested duck typing (which I'm not very familiar with) another is to check whether the `model` is an instance of `numbers.Number` and if so interpret that as a call to `quantile_filter` using the existing syntax of `clip_filter()`.

cdeline:
>     1. the master filter should be called `clip_filter` and the thing that is today called `clip_filter` should be called `quantile_clip_filter`
> 
+1 vote from me.  I think we can move forward with this plan for the current pull request.  I was starting to think we might want to split into 2, but I think this is straightforward enough.  The TrendAnalysis interactions may be a different story though...


>     2. For backward compatibility, we should check for a special case when someone calls `clip_filter(power_ac, quantile)` using positional syntax. @kanderso-nrel  suggested duck typing (which I'm not very familiar with) another is to check whether the `model` is an instance of `numbers.Number` and if so interpret that as a call to `quantile_filter` using the existing syntax of `clip_filter()`.

I like just checking for the quantile keyword or if model is a number.  Our example notebook leaves the quantile blank, I'm guessing 90% of our users don't bother messing with it.  I'm fine with making this API change at the 2.1 release, semantic versioning is a suggestion not an iron law.

cdeline:
> On that note, how would `TrendAnalysis.filter_params` work with a master function? I think our current `filter_params` approach of having function names as keys and kwargs as values is not flexible enough to support calling a master function multiple times to combine the different clipping filters, so we'd either have to change how `filter_params` works or make it clear that people need to use the lower-level functions instead of the master wrapper.

Yes, but the lower level functions don't currently get exposed in the TrendAnalysis code.  One possibility is if the model kwarg was a list instead of a string, it could trigger master filter to loop through multiple times?  we would need to ensure the kwargs in the underlying functions are unique so there's no ambiguous assignment...

> 
> Edit: regardless I'd say adding a wrapper function for clipping filters probably should be done in a separate PR to keep this one manageable.

Yeah, you might be right - maybe saving the upper level wrapper change and TrendAnalysis would work best as a new PR??

rdtools/filtering.py
```diff
@@ -102,25 +185,325 @@ def clip_filter(power_ac, quantile=0.98):
...
+    Parameters
+    ----------
+    power_ac : pd.Series
+        Pandas time series, representing PV system power or energy with
+        a pandas datetime index.
+    mounting_type: string, default 'Fixed'
+        String representing the mounting configuration associated with the
+        AC power/energy time series. Can either be "Fixed" or "Tracking".
+        Default set to 'Fixed'.
+    derivative_cutoff : float, default 0.2
```
mdeceglie, line 284: I wonder if we can come up with a more accurate term than derivative. I think we're calculating something more like a "relative range" but I'm not sure if there is a better more technically correct term. "Relative variation" or something maybe? 

Also, I think it's a percentage, is that right? We should specify.

mdeceglie (commented):
I haven't gotten all the way through yet, but here are some comments for consideration

rdtools/filtering.py
kanderso-nrel: Some similar commentary here: https://github.com/NREL/rdtools/pull/200#discussion_r591865932

rdtools/filtering.py
```diff
@@ -102,25 +185,325 @@ def clip_filter(power_ac, quantile=0.98):
...
+        daily_clipping_min = daily_clipping_min.reindex(index=power_ac.index,
+                                                        method='ffill')
+        daily_clipping_max = daily_clipping_max.reindex(index=power_ac.index,
+                                                        method='ffill')
+    # Set all values to clipping that are between the maximum and minimum
+    # power levels where clipping was found on a daily basis.
+    final_clip = (daily_clipping_min <= power_ac) & \
+        (power_ac <= daily_clipping_max)
+    final_clip = final_clip.reindex(index=power_copy.index,
+                                    fill_value=False)
```
mdeceglie, line 415: I wonder if the final reindex will get us into trouble with irregular series as input. Would it work to reindex `daily_clipping_min` and `daily_clipping_max` in the conditionals above to `power_copy.index` and then here doing:
```
    final_clip = (daily_clipping_min <= power_copy) & \
        (power_copy <= daily_clipping_max)
```
work?

rdtools/filtering.py
```diff
@@ -102,25 +185,325 @@ def clip_filter(power_ac, quantile=0.98):
...
+    # changes for clipping evaluation
+    original_time_series_sampling_frequency = time_series_sampling_frequency
+    power_copy = power_ac.copy()
+    # Drop duplciate indices
+    power_ac = power_ac.reset_index().drop_duplicates(
+                        subset=power_ac.index.name,
+                        keep='first').set_index(power_ac.index.name)
+    freq_string = str(time_series_sampling_frequency) + 'T'
+    if time_series_sampling_frequency >= 10:
+        power_ac = power_ac.asfreq(freq_string)
```
mdeceglie, line 322: For conforming to a regular frequency, what do you think about interpolating instead of just using `asfreq()`?

kanderso-nrel:
@kperrynrel -- the model loading issue might be related to whether `scikit-learn` is installed.  If I remove it from my test environment I can reproduce the same error the CI is currently showing (`TypeError: NoneType.__new__(X): X is not a type object (NoneType)`), and installing `scikit-learn` resolves the issue for me locally.  Although then the tests fail for another reason, but I think that one is unrelated and probably easy to fix.  So maybe `scikit-learn` has to be a dependency after all, even if we don't use it explicitly.  I guess try adding it back into requirements.txt and see what the CI thinks?

kperrynrel:
@kanderso-nrel that is incredibly odd, as the build is XGBoost, which should be entirely independent of scikit-learn. Thanks for the tip off on this--I'm going to do some reading today to figure out why we're needing sklearn to run things.

rdtools/filtering.py
mdeceglie: I still think we need to eliminate "derivative" in favor of something like "rolling range" or similar since we're not strictly calculating a derivative

mdeceglie:
For both new filters let's use `single_axis_tracking` rather than `tracking` as the allowed parameter value

mdeceglie:
How well have we characterized/validated the xgboost method for irregular time series? If we're not sure, perhaps raise a warning for irregular time steps?

mdeceglie:
It looks like the test coverage drops from 100 to 84% on the filtering module. Can we improve that? (https://rdtools.readthedocs.io/en/stable/developer_notes.html#running-the-test-suite)

kperrynrel:
> For both new filters let's use `single_axis_tracking` rather than `tracking` as the allowed parameter value

rdtools/filtering.py
```diff
@@ -79,7 +96,80 @@ def tcell_filter(temperature_cell, temperature_cell_low=-50,
...
+    """
+    Master wrapper for running one of the desired clipping filters.
+    The default filter run is the quantile clipping filter.
+    
+    Parameters
+    ----------
+    power_ac : pd.Series
+        Pandas time series, representing PV system with
+        a pandas datetime index.
+    model : string, default 'quantile_clip_filter'
```
mdeceglie, line 133: I suggest changing the parameter options to just `quantile`, `xgboost`, and `logic`

rdtools/plotting.py
```diff
@@ -256,14 +259,61 @@ def soiling_rate_histogram(soiling_info, bins=None):
...
-            bins=bins)
+    ax.hist(100.0 * soiling_summary.loc[soiling_summary['valid'],
+                                        'soiling_rate'], bins=bins)
     ax.set_xlabel('Soiling rate (%/day)')
     ax.set_ylabel('Count')
 
     return fig
 
 
+def tune_clip_filter_plot(power_ac, clipping_mask, display_web_browser=True):
```
mdeceglie, line 270: I am ambivalent about the default value for `display_web_browser` if we think most users will use it in a notebook, I think `False` is preferable. But if people will use it in spyder etc then maybe `True` is better.

rdtools/plotting.py
```diff
@@ -256,14 +259,61 @@ def soiling_rate_histogram(soiling_info, bins=None):
...
+    """
+    This function allows the user to visualize a clipping filter in
+    a matplotlib plot, after tweaking the function's different
```
mdeceglie, line 273:
```suggestion
    a plotly plot, after tweaking the function's different
```

rdtools/plotting.py
```diff
@@ -256,14 +259,61 @@ def soiling_rate_histogram(soiling_info, bins=None):
...
+    df = pd.DataFrame(power_ac)
+    # Add the clipping mask as a column
+    df['clipping_mask'] = clipping_mask
+    df = df.reset_index()
+    fig = px.scatter(df, x=index_name, y=column_name, color='clipping_mask')
+    # If display_web_browser is set to True, the time series with clipping
+    # is rendered via the web browser.
+    if display_web_browser is True:
+        fig.show(renderer="browser")
+    return fig
```
mdeceglie, line 314: I'd suggest adjusting the legend. The current version can be confusing because clipping_filter=True actually means clipping not detected. Maybe the legend values can be changed True->not clipping and False->clipping?

rdtools/plotting.py
```diff
@@ -256,14 +259,61 @@ def soiling_rate_histogram(soiling_info, bins=None):
...
+
```
mdeceglie, line 315: We should consider making this more general for inspecting any filter. Could be useful for clear sky as well for example. I don't think we would need to change the code at all, just the name and docs. If we go with this suggestion we should rethink my above comment about the legend and perhaps use "included" and "excluded"

rdtools/plotting.py
```diff
@@ -256,14 +259,61 @@ def soiling_rate_histogram(soiling_info, bins=None):
...
+
```
mdeceglie, line 316: Trying to run this on large data sets, e.g. in degradation_and_soiling_example_pvdaq_4.ipynb:
`rdtools.tune_clip_filter_plot(df['power'], (clip_mask), display_web_browser=False)`
Seems to be too arduous for my computer. Is there anything we can do to help with large datasets?

mdeceglie:
Since we've identified some bugs and edge cases recently, let's have the new functions emit a experimental warning like the soiling module

rdtools/test/filtering_test.py
```diff
@@ -1,96 +1,298 @@
...
+    power = pd.Series(np.arange(1, 101))
+    # Note: Power is expected to be Series object because clip_filter makes
+    #       use of the Series.quantile() method.
+    filtered = quantile_clip_filter(power, quantile=0.98)
+    # Expect 99% of the 98th quantile to be filtered
+    expected_result = power < (98 * 0.99)
+    assert ((expected_result == filtered).all())
+
+
+def test_logic_clip_filter(generate_power_time_series_no_clipping,
```
kanderso-nrel, line 132: These three test functions are pretty similar except for which function is being run, would it make sense to have a single test function [parametrized](https://docs.pytest.org/en/6.2.x/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions) across the filter functions to be more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)?

rdtools/test/filtering_test.py
```diff
@@ -1,96 +1,298 @@
...
+        generate_power_time_series_clipping
+    # Expect 4 values in middle of sequence to be clipped (when x=50)
+    mask_c = logic_clip_filter(power_datetime_index_c)
+    filtered_c = power_datetime_index_c[mask_c]
+    assert bool(mask_nc.all(axis=None))
+    assert (len(filtered_c) == 96)
+    assert bool((mask_one_min.index.to_series().diff()[1:] ==
+                 np.timedelta64(60, 's')).all(axis=None))
+    assert bool((mask_irregular.index == power_datetime_index_irregular.index)
+                .all(axis=None))
```
kanderso-nrel, line 184: I think the calls to `bool` are unnecessary here, as well as `axis=None`.  FYI there's also [series.index.equals](https://pandas.pydata.org/docs/reference/api/pandas.Index.equals.html) for checking index equality

rdtools/filtering.py
```diff
@@ -102,25 +192,504 @@ def clip_filter(power_ac, quantile=0.98):
...
+        power_ac = power_ac.asfreq(freq_string)
+    # if time_series_sampling_frequency >= 10:
+    #     power_ac = power_ac.asfreq(freq_string)
+    # High frequency data (less than 10 minutes) has demonstrated
+    # potential to have more noise than low frequency  data.
+    # Therefore, the  data is resampled to a 15-minute median
+    # before running the filter.
+    if time_series_sampling_frequency < 10:
+        power_ac = power_ac.resample('15T').mean()
+        time_series_sampling_frequency = 15
```
mdeceglie, line 429: Should this be `elif` with the preceding `if`?

rdtools/plotting.py
```diff
@@ -268,14 +270,65 @@ def soiling_rate_histogram(soiling_info, bins=None):
...
+    This function allows the user to visualize filtered data in
+    a Plotly plot, after tweaking the function's different
+    hyperparameters. The plot can be zoomed in on, for an in-depth look at
+    the mask in the AC power time series.
+
+    Parameters
+    ----------
+    power_ac : pd.Series
+        AC power in Watts. Index of the Pandas series is a Pandas
+        datetime index.
```
mdeceglie, line 292: I suggest changing the name of this parameter to be more general. For example, the plotting function works well with POA or PV energy. Maybe call it `signal` and give pv power, energy, irradiance, or temperature as examples and note that as long as they share their index with the filter, it should work.

setup.py
```diff
@@ -39,12 +39,16 @@
     'matplotlib >= 3.0.0',
     'numpy >= 1.15',
     # exclude pandas==1.0.0 & 1.0.1 for GH142, and 0.24.0 for GH114
-    'pandas >= 0.23.0,!=0.24.0,!=1.0.0,!=1.0.1',
+    'pandas >= 0.23.0,!=0.24.0,!=1.0.0,!=1.0.1,<=1.1.3',
```
kanderso-nrel, line 42: Why exclude pandas versions >1.1.3?

kanderso-nrel (commented):
A "meta-review" focused more on docs and such than the code itself

rdtools/filtering.py
```diff
@@ -79,7 +97,80 @@ def tcell_filter(temperature_cell, temperature_cell_low=-50,
...
 
 
-def clip_filter(power_ac, quantile=0.98):
+def csi_filter(poa_global_measured, poa_global_clearsky, threshold=0.15):
+    '''
+    Filtering based on clear-sky index (csi)
+
+    Parameters
+    ----------
+    poa_global_measured : pd.Series
```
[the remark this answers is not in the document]
mdeceglie: I also noticed that some changes I made to the warning messages on this branch got lost along the way here

mdeceglie (approved):
Looks great. Let's go ahead and merge. Thanks everyone!

cdeline (commented):
Nice job, thanks for keeping this moving forward.

Issue #205, opened 2020-10-05
annual_soiling_ratios() doesn't perform insolation weighting


mdeceglie:
`soiling.annual_soiling_ratios()` stated output according to docs is insolation-weighted soiling ratio, however it does not perform insolation weighting. Insolation-weighted soiling ratio is the value returned by `soiling_srr()` and `SRRAnalysis.run()`. (Currently this function is not part of tagged release, this only affects development branch)

mdeceglie:
Closed with #214

PR #207, opened 2020-10-05
Fix/issue 205

mdeceglie:
Our previous version of `soiling.annual_soiling_ratios()` was not properly calculating the **irradiance-weighted** soiling ratio. (See #205)

- [x] Updated changelog


rdtools/soiling.py
```diff
@@ -791,7 +791,9 @@ def annual_soiling_ratios(stochastic_soiling_profiles, confidence_level=68.2):
         List of pd.Series representing profile realizations from the SRR monte carlo.
         Typically ``soiling_interval_summary['stochastic_soiling_profiles']`` obtained with
         :py:func:`rdtools.soiling.soiling_srr` or :py:meth:`rdtools.soiling.SRRAnalysis.run`
-
+    insolation_daily : pd.Series
+        Daily plane-of-array insolation corresponding to
+        `energy_normalized_daily`. Arbitrary units.
```
kanderso-nrel, line 796: I don't think `energy_normalized_daily` exists in this context

rdtools/soiling.py
```diff
@@ -820,12 +822,20 @@ def annual_soiling_ratios(stochastic_soiling_profiles, confidence_level=68.2):
...
     '''
 
-    all_profiles = pd.concat(stochastic_soiling_profiles)
-    annual_groups = all_profiles.groupby(all_profiles.index.year)
+    # Create a df with each realization as a column
+    all_profiles = pd.concat(stochastic_soiling_profiles, axis=1)
+    # Weight each day by insolation
+    all_profiles_weighted = all_profiles.multiply(insolation_daily, axis=0)
+
+    # Compute the insolation-weighted soiling ration (IWSR) for each realization
```
kanderso-nrel, line 830:
```suggestion
    # Compute the insolation-weighted soiling ratio (IWSR) for each realization
```

rdtools/soiling.py
```diff
@@ -820,12 +822,20 @@ def annual_soiling_ratios(stochastic_soiling_profiles, confidence_level=68.2):
...
+    annual_insolation = insolation_daily.groupby(insolation_daily.index.year).sum()
```
kanderso-nrel, line 831: I think there might be a bug here.  The date range of `all_profiles` doesn't necessarily match up with the date range of `insolation_daily`, which means `all_profiles_weighted` might have a bunch of nans that don't get handled appropriately in `annual_insolation`.  For example in the PVDAQ 4 notebook, `insolation_daily` starts at `2010-02-25` but the first date in `all_profiles` is `2010-03-12`.  The insolation from 2-25 to 3-12 gets counted in the annual insolation, but because there is no soiling profile data in that range, the annual sum of `all_profiles_weighted` misses the points (equivalenty, treating the range as if the system was 100% soiled) in that period.  A similar effect happens at the end of the date range because `insolation_daily` extends farther than `all_profiles`. I think this might explain why the first and last years in the notebook's annual soiling ratio results are so much lower than the other years.

So maybe reindex or mask `insolation_daily` to match `all_profiles`?  And on a related note, is there any way that nans could be present in `all_profiles`?  That would cause a similar mismatch.

docs/degradation_and_soiling_example_pvdaq_4.ipynb
```diff
@@ -616,37 +616,6 @@
...
-       "Name: start, Length: 64, dtype: datetime64[ns, Etc/GMT+7]"
-      ]
-     },
-     "execution_count": 17,
-     "metadata": {},
-     "output_type": "execute_result"
-    }
-   ],
-   "source": [
-    "soiling_summary['start']"
```
kanderso-nrel, line 643: Was it a conscious decision to remove this cell?  OK with me either way, just wanted to make sure it wasn't an accident.

docs/sphinx/source/changelog/pending.rst
```diff
@@ -7,6 +7,7 @@ API Changes
 -----------
 
 * Change the column names 'slope', 'slope_low', and 'slope_high' to 'soiling_rate', 'soiling_rate_low', and 'soiling_rate_high' in ``calc_info['soiling_interval_summary']`` returned from :py:meth:`~rdtools.soiling.SRRAnalysis.run()` and :py:func:`rdtools.soiling.soiling_srr()` (:pull:`193`).
+* Change the calculations in :py:func:`rdtools.soiling.annual_soiling_raatios()` to perform irradiance-weighting (:issue:`205` :pull:`207`)
```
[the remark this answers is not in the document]
mdeceglie: Yeah, perhaps better to keep it under one bullet and list both PRs

rdtools/soiling.py
mdeceglie: Good point. I think I'll add an `all_profiles.dropna()` to your suggestion to get rid of any nan rows. I don't think there should be any, but if there are, they should appear in every realization.

docs/degradation_and_soiling_example_pvdaq_4.ipynb
mdeceglie: Confirmed. I think that was left in there from some prior debugging.

rdtools/soiling.py
mdeceglie: I also added a warning and similar logic to the original Monte Carlo. Let me know what you think.

rdtools/test/soiling_test.py
```diff
@@ -234,37 +234,53 @@ def test_soiling_srr_with_nan_interval(normalized_daily, insolation, times):
...
+    expected_data = np.array([[2018, 4.5, 1.431, 7.569],
+        [2019, 14.5, 11.431, 17.569]])
     expected = pd.DataFrame(data=expected_data,
         columns=['year', 'soiling_ratio_median', 'soiling_ratio_low', 'soiling_ratio_high'])
     expected['year'] = expected['year'].astype(int)
     
-    result = annual_soiling_ratios(srr_profiles)
+    srr_profiles, insolation = multi_year_profiles
+    result = annual_soiling_ratios(srr_profiles, insolation)
+    print(result)
```
kanderso-nrel, line 259: leftover print, here and below

rdtools/test/soiling_test.py
```diff
@@ -234,37 +234,53 @@ def test_soiling_srr_with_nan_interval(normalized_daily, insolation, times):
...
+    srr_profiles, insolation = multi_year_profiles
+    result = annual_soiling_ratios(srr_profiles, insolation, confidence_level=95)
+    print(result)
 
     pd.testing.assert_frame_equal(result, expected, atol=1e-8)
 
+def test_annual_soiling_ratios_warning(multi_year_profiles):
+    srr_profiles, insolation = multi_year_profiles
+    insolation = insolation.iloc[:-200]
+    with pytest.warns(UserWarning):
```
kanderso-nrel, line 280: UserWarning is pretty common, could consider using the `match` parameter to increase the precision, e.g.: https://github.com/NREL/rdtools/blob/development/rdtools/test/availability_test.py#L472-L473

kanderso-nrel (approved):
LGTM!  A couple minor comments on the tests

PR #209, opened 2020-10-06
Patch/notebook requirements

mdeceglie:
- [x] Updated changelog


mdeceglie:
my problem was potentially related to using python 3.8. Opened #210  in lieu of pushing this through

Issue #206, opened 2020-10-05
normalization.energy_from_power does not use power.index.freq where available


kanderso-nrel:
**Is your feature request related to a problem? Please describe.**
`normalization.energy_from_power` calculates the difference between timestamps to determine the measurement period of the data.  I don't really have an issue with this when there is a more straightforward alternative of using the `power.index.freq` attribute (maybe it's a slight inefficiency) but in the case where the time series is a single value, the `np.diff` approach fails.  It would be nice to have a shortcut of using the stated period when it is available. 

**Full error message and traceback**
```python
In [1]: import rdtools
   ...: rdtools.__version__
   ...: 
Out[1]: '2.0.0-beta.0+147.g1cb2a08'

In [2]: import pandas as pd
   ...: power = pd.Series([100], index=pd.date_range('2019-01-01', freq='15T', periods=1))
   ...: power
   ...: 
Out[2]: 
2019-01-01    100
Freq: 15T, dtype: int64

In [3]: rdtools.energy_from_power(power)
   ...: 
C:\Users\KANDERSO\Software\Anaconda3\envs\fleets-dev\lib\site-packages\numpy\lib\nanfunctions.py:1112: RuntimeWarning: Mean of empty slice
  return np.nanmean(a, axis, out=out, keepdims=keepdims)
Traceback (most recent call last):

  File "<ipython-input-3-06a550e25220>", line 1, in <module>
    rdtools.energy_from_power(power)

  File "C:\Users\KANDERSO\Software\Anaconda3\envs\fleets-dev\lib\site-packages\rdtools\normalization.py", line 495, in energy_from_power
    target_frequency = str(int(median_step_ns)) + 'N'

ValueError: cannot convert float NaN to integer
```

**Describe the solution you'd like**
Have `normalization.energy_from_power` use `power.index.freq` if it exists, otherwise fall back to the current approach.

**Describe alternatives you've considered**
None

**Additional context**
I stumbled on this in an edge case when the new AvailabilityAnalysis class found outages that only last one timestamp: https://github.com/NREL/rdtools/blob/development/rdtools/availability.py#L492

mdeceglie:
I think this should be a pretty straightforward change. The median time step is also used for the `max_time_delta`, when `max_time_delta` is not supplied.

mdeceglie:
We will also have to figure out this error for the single-value case:
```python
In [1]: import rdtools
   ...: import pandas as pd
   ...: 
[...]

PR #212, opened 2020-10-07
Allow single-value timeseries in energy_from_power

kanderso-nrel:
- [x] Updated changelog

See #206.  The diff is a mess because I indented a big block of code and git isn't having it, but besides that the block wasn't changed.


rdtools/normalization.py
```diff
@@ -488,7 +488,11 @@ def energy_from_power(power, target_frequency=None, max_timedelta=None):
                          'DatetimeIndex')
 
     t_steps = t_step_nanoseconds(power)
-    median_step_ns = t_steps.median()
+    if power.index.freq is not None:
+        timedelta = pd.to_timedelta(power.index.freq)
+        median_step_ns = timedelta.total_seconds() * 1e9
```
kanderso-nrel, line 493: I did this before I saw the `pd.tseries.frequencies.to_offset(freq).nanos` approach below.  Is there reason to prefer one over the other?

rdtools/normalization.py
```diff
@@ -513,37 +517,44 @@ def energy_from_power(power, target_frequency=None, max_timedelta=None):
...
-        if energy.index.tz is None:
-            energy = energy.drop(energy.index[0])
-        else:
-            tz = str(energy.index.tz)
-            energy.index = energy.index.tz_convert('UTC')
-            energy = energy.drop(energy.index[0])
-            energy.index = energy.index.tz_convert(tz)
+    if len(power) <= 1:
+        # just one value, doesn't make sense to interpolate or trapz aggregate.
+        energy = power * freq_interval_size_ns / 1e9 / 3600  # ns to s to h
```
kanderso-nrel, line 522: Is `max_timedelta` meaningful for single-value inputs?  I don't think so, so it's not used in this branch.

kanderso-nrel: Also we get the timezone and frequency automatically from the input by doing this, so no need to do anything explicit like the main branch does.

kanderso-nrel (commented):
The case of single-value, no `freq` is not handled gracefully:

```python
In [11]: from rdtools.normalization import energy_from_power
    ...: import pandas as pd
    ...:
    ...: power = pd.Series([100], pd.date_range('2019-01-01', periods=1, freq='15T'))
    ...: power.index.freq = None
    ...: energy_from_power(power)
    ...:

ValueError                                Traceback (most recent call last)
<ipython-input-11-f4a9dbc05c6b> in <module>()
      4 power = pd.Series([100], pd.date_range('2019-01-01', periods=1, freq='15T'))
      5 power.index.freq = None
----> 6 energy_from_power(power)

C:\Users\KANDERSO\projects\rdtools\rdtools\normalization.py in energy_from_power(power, target_frequency, max_timedelta)
    497     if target_frequency is None:
    498         # 'N' is the Pandas offset alias for ns
--> 499         target_frequency = str(int(median_step_ns)) + 'N'
    500
    501     if max_timedelta is None:

ValueError: cannot convert float NaN to integer
```

rdtools/normalization.py
mdeceglie: I don't know of a reason to prefer one over the other. The only thing to consider would be consolidating the repeated logic in a function `_freq_to_nanoseconds()` or similar. Borderline case in my opinion.

rdtools/normalization.py
mdeceglie: I agree on both counts

mdeceglie:
I'm getting a numpy runtime warning with this example:
```python
In [8]: from rdtools.normalization import energy_from_power
   ...: import pandas as pd
   ...: power = pd.Series([100], pd.date_range('2019-01-01', periods=1, freq='15 T'))
   ...: energy_from_power(power)
   ...: 
/Users/mdecegli/opt/anaconda3/envs/rdtools_dev_mode/lib/python3.7/site-packages/numpy/lib/nanfunctions.py:1111: RuntimeWarning: Mean of empty slice
  return np.nanmean(a, axis, out=out, keepdims=keepdims)
Out[8]: 
2019-01-01    25.0
Freq: 15T, Name: energy_Wh, dtype: float64
```
Do you see the same warning @kanderso-nrel ?

kanderso-nrel:
Hmm I do now that the line `median_step_ns = t_steps.median()` is evaluated for all inputs.  Could switch up that line and the following conditional to check for `t_steps.notnull().any()`, or suppress the warning with the warnings module (looks like `np.errstate` doesn't work on this warning).  Any preference @mdeceglie?

mdeceglie:
There's another edge case that errors out in a not too helpful way, which is a single value series with a non-fixed frequency
```python
In [20]: from rdtools.normalization import energy_from_power
    ...: import pandas as pd
    ...: power = pd.Series([100], pd.date_range('2019-01-01', periods=1, freq='M'))
    ...: energy_from_power(power)
    ...: 
/Users/mdecegli/opt/anaconda3/envs/rdtools_dev_mode/lib/python3.7/site-packages/numpy/lib/nanfunctions.py:1111: RuntimeWarning: Mean of empty slice
  return np.nanmean(a, axis, out=out, keepdims=keepdims)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-20-197246666e5e> in <module>()
      2 import pandas as pd
      3 power = pd.Series([100], pd.date_range('2019-01-01', periods=1, freq='M'))
----> 4 energy_from_power(power)

/Users/mdecegli/Documents/GitHub/rdtools/rdtools/normalization.py in energy_from_power(power, target_frequency, max_timedelta)
    493     if np.isnan(median_step_ns):
    494         if power.index.freq is not None:
--> 495             timedelta = pd.to_timedelta(power.index.freq)
    496             median_step_ns = timedelta.total_seconds() * 1e9
    497         else:

/Users/mdecegli/opt/anaconda3/envs/rdtools_dev_mode/lib/python3.7/site-packages/pandas/core/tools/timedeltas.py in to_timedelta(arg, unit, errors)
    120 
    121     # ...so it must be a scalar value. Return scalar.
--> 122     return _coerce_scalar_to_timedelta_type(arg, unit=unit, errors=errors)
    123 
    124 

/Users/mdecegli/opt/anaconda3/envs/rdtools_dev_mode/lib/python3.7/site-packages/pandas/core/tools/timedeltas.py in _coerce_scalar_to_timedelta_type(r, unit, errors)
    126     """Convert string 'r' to a timedelta object."""
    127     try:
--> 128         result = Timedelta(r, unit)
    129     except ValueError:
    130         if errors == "raise":

pandas/_libs/tslibs/timedeltas.pyx in pandas._libs.tslibs.timedeltas.Timedelta.__new__()

ValueError: Value must be Timedelta, string, integer, float, timedelta or convertible, not MonthEnd
```

What do you think about adding a simple-as-possible `if len(power) <= 1:` near the top of the function that ends in `return energy`? It seems there is a bunch of code not relevant to the single-value case that still resides outside the length conditional.

mdeceglie (approved):
Looks good, thanks!

PR #213, opened 2020-10-08
Add import warnings for soiling and availability modules

kanderso-nrel:
- [x] Updated changelog

Behavior:

```
In [1]: import rdtools

In [2]: from rdtools.soiling import SRRAnalysis
C:\Users\KANDERSO\projects\rdtools\rdtools\soiling.py:9: UserWarning: The soiling analysis functions are currently experimental. The API and results may change in future releases as the code matures.
  'The soiling analysis functions are currently experimental. The API and '

In [3]: from rdtools.availability import AvailabilityAnalysis
C:\Users\KANDERSO\projects\rdtools\rdtools\availability.py:14: UserWarning: The availability analysis functions are currently experimental. The API and results may change in future releases as the code matures.
  'The availability analysis functions are currently experimental. The API '
```


rdtools/__init__.py
```diff
@@ -14,15 +14,15 @@
...
-from rdtools.soiling import monthly_soiling_rates
-from rdtools.soiling import annual_soiling_ratios
+# from rdtools.soiling import soiling_srr
+# from rdtools.soiling import monthly_soiling_rates
+# from rdtools.soiling import annual_soiling_ratios
 from rdtools.plotting import degradation_summary_plots
 from rdtools.plotting import soiling_monte_carlo_plot
 from rdtools.plotting import soiling_interval_plot
 from rdtools.plotting import soiling_rate_histogram
 from rdtools.plotting import availability_summary_plots
```
kanderso-nrel, line 24: Should we comment out the soiling/availability plotting function imports too?  Do they need a warning?

rdtools/availability.py
```diff
@@ -10,6 +10,11 @@
 from scipy.interpolate import interp1d
 import warnings
 
+warnings.warn(
+    'The availability analysis functions are currently experimental. The API '
+    'and results may change in future releases as the code matures.'
```
mdeceglie, line 15: What do you think about:
```
The availability module is currently experimental. The API, results,
and default behaviors may change in future releases (including MINOR
and PATCH releases) as the code matures.
```

rdtools/soiling.py
```diff
@@ -5,6 +5,10 @@
 import numpy as np
 from scipy.stats.mstats import theilslopes
 
+warnings.warn(
+    'The soiling analysis functions are currently experimental. The API and '
```
mdeceglie, line 9: (similar suggestion as in availability)

mdeceglie:
Thanks for getting this started @kanderso-nrel. I wonder if we should do something in RTD as well. Perhaps in the overview and or in the API section (maybe via module-level docstring). A more involved option would be to break out a separate part in the RTD sidebar for "API (Experimental)" or similar. We should seek a balance between making the experimental nature obvious to folks and giving ourselves too many things to remember to change back when these solidify.

kanderso-nrel:
Rather than hardcoding (slightly different) module descriptions in `api.rst`, the last commit uses the `automodule` directive to bring in the module docstrings.  Because of the way `automodule` affects downstream imports, it was easiest to change the function lists to be just the function name instead of module.function, e.g. they show as `clip_filter` instead of `filtering.clip_filter` in the function lists.  Example: https://rdtools--213.org.readthedocs.build/en/213/api.html#soiling

rdtools/__init__.py
mdeceglie: I'm leading toward commenting them out of `__init__.py`, adding a warning at the function level, and adding a note to the function docstrings.

kanderso-nrel: Done, example: https://rdtools--213.org.readthedocs.build/en/213/generated/rdtools.plotting.soiling_interval_plot.html

mdeceglie (approved):
Looks good, thanks @kanderso-nrel

Issue #211, opened 2020-10-06
Update notebook_requirements.txt


mdeceglie:
installing  notebook_requirements.txt currently works but raises the following: 

```
ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.

We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.

nbclient 0.5.0 requires jupyter-client>=6.1.5, but you'll have jupyter-client 5.2.3 which is incompatible.
nbclient 0.5.0 requires nbformat>=5.0, but you'll have nbformat 4.4.0 which is incompatible.
jupyterlab-pygments 0.1.2 requires pygments<3,>=2.4.1, but you'll have pygments 2.2.0 which is incompatible.
```

mdeceglie:
Closed with #214

Issue #210, opened 2020-10-06
Document the python version used for example environment


mdeceglie:

mdeceglie:
Closed with #214

Issue #198, opened 2020-09-08
Deprecations in pvlib 0.8.0


kanderso-nrel:
pvlib 0.8.0 (released 2020-09-08) deprecated the LocalizedPVSystem class, which is used by our SAPM-based normalization functions (e.g. [sapm_dc_power](https://rdtools.readthedocs.io/en/latest/generated/rdtools.normalization.sapm_dc_power.html)).  Note that we don't currently support the pvlib 0.8.x series, so technically this isn't an issue yet.  But it would be nice to solve it sooner rather than later.  Further context: https://github.com/pvlib/pvlib-python/pull/1034#issuecomment-684032830

Some ideas:
- Deprecate the SAPM functions in RdTools 2.0 and remove them in an eventual 3.0.
    - I have the notion that the SAPM-based normalization is not commonly-used.  People that still want to use it could have that function in their own code and use our `normalize_with_expected_power` function.
- Change the `pvlib_pvsystem` parameter to optionally be an (unlocalized) PVSystem object and add an optional `location` parameter.
    - I haven't prototyped this, but it seems like it would work, the function would just have to do a little bit of argument checking to figure out where it can find what it needs.
- Switch the parameter to be a pvlib ModelChain?
    - If we want to keep to the pattern of passing in a pvlib object to do an expected energy calculation, I suppose that's the way to do it.  But I'm not sure how much it helps the user to have us call `mc.run_model(...)` instead of them doing it and passing the result to `normalize_with_expected_power`.
- Something else?

I think the LocalizedPVSystem is the only deprecation that matters for us.  I ran the test suite on development with pvlib 0.8.0 installed and got no failures, only the deprecation warnings.

mdeceglie:
Closed with #214

Issue #195, opened 2020-08-31
Labeling convention inconsistency


mdeceglie:
The `normalize_with_expected_power()`  docstring states that it takes a right labeled time series of energy or power and a right labeled time series of expected power. However it feeds those values to `energy_from_power()` which is designed to take instantaneous values of power.

mdeceglie:
Closed with #214

PR #214, opened 2020-10-12
2.0.0 Release

mdeceglie:
This PR highlights all the changes proposed for 2.0.0 relative to 1.2.3. Target for release of 2.0.0 is Oct. 15, 2020. All the code currently in `development` has been reviewed when it was merged to `development`. The purpose of this PR is to give the community a chance to comment prior to release and to make sure everything works together at a high level.

Slightly modified PR checklist:
- [x] Test coverage is adequate (total coverage: 97%)
- [x] `__init__.py` configured appropriately
- [x] API.rst is up to date
- [x] Example notebooks are run in the specified environment. Results scrutinized for differences.
- [x] Changelog is up to date (combine entry for 2.0.0b and pending)
- [x] Changelog 2.0.0 release date correct
- [x] Check the issue tracker to make sure we haven't forgotten anything that should be included
- [x] Consider deprecations

Issues to fix:
- [x] #210 
- [x] #211
- [x] #198
- [x] #195
- [x] #194
- [x] #191
- [x] #118 (leave open, not available in pvlib yet)
- [x] #114


setup.py
```diff
@@ -34,24 +36,45 @@
...
-    'pandas >= 0.23.0, <1.0.0',
+    'matplotlib >= 3.0.0',
+    'numpy >= 1.15',
+    # exclude pandas==1.0.0 & 1.0.1 for GH142, and 0.24.0 for GH114
+    'pandas >= 0.23.0,!=0.24.0,!=1.0.0,!=1.0.1',
     'statsmodels >= 0.8.0',
     'scipy >= 0.19.1',
     'h5py >= 2.7.1',
-    'pvlib >= 0.5.0, <0.6.0',
+    'pvlib >= 0.7.0, <0.8.0',
```
kanderso-nrel, line 46: We should consider allowing pvlib 0.8.x; I think the only issue is the deprecated `LocalizedPVSystem` class, which we've already deprecated for the rdtools 2.x series.

setup.py
```diff
@@ -34,24 +36,45 @@
 ]
 
 INSTALL_REQUIRES = [
-    'numpy >= 1.12',
-    'pandas >= 0.23.0, <1.0.0',
+    'matplotlib >= 3.0.0',
+    'numpy >= 1.15',
+    # exclude pandas==1.0.0 & 1.0.1 for GH142, and 0.24.0 for GH114
+    'pandas >= 0.23.0,!=0.24.0,!=1.0.0,!=1.0.1',
```
kanderso-nrel, line 42: Consider adding upper limits to these requirements (e.g. add `,<2.0.0` to pandas)

docs/sphinx/source/changelog/v2.0.0b0.rst
```diff
@@ -0,0 +1,93 @@
...
+
+Requirements
+------------
+* Drop support for Python 2.7, minimum supported version is now 3.6 (:pull:`135`).
+* Increase minimum pvlib version to 0.7.0.
+* Update requirements.txt and notebook_requirements.txt to avoid conflicting specifications. Taken together,
+  they represent the complete environment for the notebook example (:pull:`164`).
+
+Example Updates
+---------------
```
kanderso-nrel, line 78: Adding soiling to the example notebook isn't mentioned

docs/sphinx/source/changelog/v2.0.0b0.rst
```diff
@@ -0,0 +1,93 @@
...
+  * :py:func:`~rdtools.normalization.normalize_with_pvwatts` pvwatts_kws dictionary keys have been renamed.
+  * :py:func:`~rdtools.normalization.pvwatts_dc_power` input kwargs are now ``poa_global``, ``power_dc_rated``, ``temperature_cell``, ``poa_global_ref``, ``temperature_cell_ref``, ``gamma_pdc``.
+
+Enhancements
+------------
+* Add new :py:mod:`~rdtools.soiling` module to implement the stochastic rate and
+  recovery method (:pull:`112`).
+* Add new function :py:func:`~rdtools.normalization.normalize_with_expected_power` (:pull:`173`).
+* Add new functions :py:func:`~rdtools.normalization.energy_from_power` and
+  :py:func:`~rdtools.normalization.interpolate` (:pull:`105`, :pull:`108`).
```
kanderso-nrel, line 44: Should `interpolate_series`, `trapz_aggregate`, and `t_step_nanoseconds`  be mentioned here as well?

docs/sphinx/source/changelog/v2.0.0b0.rst
```diff
@@ -0,0 +1,93 @@
...
+    :py:func:`~rdtools.degradation.degradation_ols` and 
+    :py:func:`~rdtools.degradation.degradation_classical_decomposition`
+    first kwarg is now ``energy_normalized``.
+  * :py:func:`~rdtools.filtering.normalized_filter` input kwargs are now ``energy_normalized``, ``energy_normalized_low`` and ``energy_normalized_high``.
+  * :py:func:`~rdtools.filtering.poa_filter` input kwargs are now ``poa_global``, ``poa_global_low`` and ``poa_global_high``.
+  * :py:func:`~rdtools.filtering.tcell_filter` input kwargs are now ``temperature_cell``, ``temperature_cell_low`` and ``temperature_cell_high``.
+  * :py:func:`~rdtools.filtering.clip_filter` input kwargs are now ``power_ac`` and ``quantile``.
+  * :py:func:`~rdtools.filtering.csi_filter` first two kwargs are now ``poa_global_measured``, ``poa_global_clearsky``.
+  * :py:func:`~rdtools.normalization.normalize_with_pvwatts` pvwatts_kws dictionary keys have been renamed.
+  * :py:func:`~rdtools.normalization.pvwatts_dc_power` input kwargs are now ``poa_global``, ``power_dc_rated``, ``temperature_cell``, ``poa_global_ref``, ``temperature_cell_ref``, ``gamma_pdc``.
```
kanderso-nrel, line 36: `irradiance_rescale` had a parameter name change too, `irrad_sim` from `modeled_irrad`

docs/sphinx/source/changelog/v2.0.0b0.rst
```diff
@@ -0,0 +1,93 @@
...
+* Create sphinx documentation and set up ReadTheDocs (:pull:`125`).
+* Add guides on running tests and building sphinx docs (:pull:`136`).
+* Improve module-level docstrings (:pull:`137`).
+
+Requirements
+------------
+* Drop support for Python 2.7, minimum supported version is now 3.6 (:pull:`135`).
+* Increase minimum pvlib version to 0.7.0.
+* Update requirements.txt and notebook_requirements.txt to avoid conflicting specifications. Taken together,
+  they represent the complete environment for the notebook example (:pull:`164`).
```
kanderso-nrel, line 75: Missing the new matplotlib requirement

docs/sphinx/source/changelog.rst
```diff
@@ -0,0 +1,6 @@
+RdTools Change Log
+==================
+
+.. include:: changelog/pending.rst
+.. include:: changelog/v2.0.0b0.rst
```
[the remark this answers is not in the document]
mdeceglie: Agreed. I updated the punch list at the top of the PR

.coveragerc
```diff
@@ -0,0 +1,6 @@
+[run]
+omit = 
+    # _version.py doesn't count
+    rdtools/_version.py
```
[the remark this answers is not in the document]
mdeceglie: Agreed

docs/sphinx/source/index.rst
```diff
@@ -0,0 +1,304 @@
...
+
+For more detailed instructions, see the :ref:`developer_notes` page.
+
+RdTools currently is tested on Python 3.6+.
+
+Usage and examples
+------------------
+
+Full workflow examples are found in the notebooks in `example notebook`_.
+The examples are designed to work with python 3.6. For a consistent
```
mdeceglie, line 148:
```suggestion
The examples are designed to work with python 3.7. For a consistent
```

docs/sphinx/source/changelog/v2.0.0b0.rst
mdeceglie: Perhaps they ought to be made private with leading `_` ?

setup.py
mdeceglie: Can we safely leave off these upper limits until Travis finds a problem with `upgrade-strategy=eager`?

setup.py
mdeceglie: I think that makes sense

docs/sphinx/source/index.rst
```diff
@@ -0,0 +1,304 @@
...
+
+.. code:: python
+
+   import rdtools
+
+The most frequently used functions are:
+
+.. code:: python
+
+   normalization.normalize_with_pvwatts(energy, pvwatts_kws)
```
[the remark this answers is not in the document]
mdeceglie: Good catch!

setup.py
kanderso-nrel: Eager upgrades for testing helps, but the problem is that we can't retroactively impose upper limits.  For example we release rdtools 2.0 today, pandas releases 2.0 tomorrow and breaks something, anyone who installs rdtools 2.0 will have a broken installation because we can only restrict the pandas version going forward.

docs/sphinx/source/changelog/v2.0.0b0.rst
kanderso-nrel: Up to you.  Those three are listed in `api.rst`, so that'll need to be changed if they are made private.

docs/degradation_example.ipynb
```diff
@@ -1,630 +0,0 @@
-{
```
GregoryKimball, line 1: I see that this file was modified and renamed as `docs/degradation_and_soiling_example.ipynb`. Were there any changes to the sections other than the soiling model addition (new Section 5)? 

There is an opportunity to use the soiling model as part of data normalization. This could entail correcting performance index for soiling losses as an "expected clean" performance index. For this release that idea is out of scope but is it on your roadmap?

docs/system_availability_example.ipynb
```diff
@@ -0,0 +1,664 @@
+{
```
GregoryKimball, line 1: This is a nice addition. Bringing in analytics around inverter outages versus communication outages is useful for energy loss accounting but not so useful for degradation analysis. Do you envision `rdtools` scope moving towards energy loss accounting?

I'm glad to see more use of energy tags and a flexible 'expected power' tag.

How does the algorithm handle partial day comms or inverter outages? How about overlapping outages?

rdtools/__init__.py
```diff
@@ -1,6 +1,9 @@
 from rdtools.normalization import normalize_with_sapm
 from rdtools.normalization import normalize_with_pvwatts
 from rdtools.normalization import irradiance_rescale
```
GregoryKimball, line 3: I didn't realize `rdtools` was explicitly importing so many submodules. Generally I prefer leaving the python namespace clear and using local aliasing if needed.

rdtools/__init__.py
```diff
@@ -10,6 +13,17 @@
 from rdtools.filtering import poa_filter
 from rdtools.filtering import tcell_filter
 from rdtools.filtering import clip_filter
+from rdtools.filtering import normalized_filter
+# from rdtools.soiling import soiling_srr
+# from rdtools.soiling import monthly_soiling_rates
+# from rdtools.soiling import annual_soiling_ratios
```
GregoryKimball, line 19: Were these import comments intentionally left?

rdtools/availability.py
```diff
@@ -0,0 +1,620 @@
...
+import pandas as pd
+import numpy as np
+from scipy.interpolate import interp1d
+import warnings
+
+warnings.warn(
+    'The availability module is currently experimental. The API, results, '
+    'and default behaviors may change in future releases (including MINOR '
+    'and PATCH releases) as the code matures.'
+)
```
GregoryKimball, line 21: Wow this is a loud warning!

rdtools/degradation.py
```diff
@@ -314,7 +339,7 @@ def _mk_test(x, alpha=0.05):
     if s > 0:
         z = (s - 1) / np.sqrt(var_s)
     elif s == 0:
-            z = 0
+        z = 0
```
GregoryKimball, line 342: good catch...

rdtools/normalization.py
```diff
@@ -1,63 +1,125 @@
...
+
+    index_union = energy_normalized.index.union(insolation.index)
+    energy_normalized = energy_normalized.reindex(index_union)
+    insolation = insolation.reindex(index_union)
+
+    return energy_normalized, insolation
+
+
+@deprecated(since='2.0.0', removal='3.0.0',
+            alternative='normalize_with_expected_power')
```
GregoryKimball, line 72: Do we have a function to generate `expected_power` when this function is removed? I would like to encourage the team to include the same range of expected power functions even as the normalization module simplifies to only include `normalize_with_expected_power`. I understand that the models are simple or documented elsewhere but having some examples of how to use them will support the goal of standardized analysis.

I would recommend migrating them to an `expected_power` module or to an example notebook focused on expected power modeling.

rdtools/normalization.py
```diff
@@ -209,175 +258,524 @@ def normalize_with_sapm(energy, sapm_kws):
...
+        # a meaningful interval for the first element of the timeseries
+        # Length of each interval calculated by using 'int64' to convert to
+        # nanoseconds.
+        deltas = (series.index - series.index.shift(-1)).astype('int64') / \
+                    (10.0**9 * 3600.0)
     return deltas, np.mean(deltas.dropna())
 
 
-def irradiance_rescale(irrad, modeled_irrad, max_iterations=100, method=None):
+def irradiance_rescale(irrad, irrad_sim, max_iterations=100,
```
GregoryKimball, line 334: This is a cool idea for a function. Is there a citation for this method? This seems like another great candidate for an example notebook.

GregoryKimball (approved):
Great work guys. The additions around soiling and availability are welcome and support the goal of offering standard tools for common problems.

I would like to see the example notebooks continue to grow with the new features, and please try to keep the expected power modeling support as you move through version 2.X development.

rdtools/__init__.py
mdeceglie: Agreed. But putting them in `__init__.py` only makes them available as, e.g. `rdtools.normalize_with_sapm()` it prevents the user from needing to use `rdtools.normalization.normalize_with_sapm()`. They don't end up polluting the python namespace. Did I understand your comment correctly?

```python
In [1]: import rdtools

In [2]: rdtools.normalize_with_sapm
Out[2]: <function rdtools.normalization.normalize_with_sapm>

In [3]: normalize_with_sapm
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-3-1103d42db088> in <module>()
----> 1 normalize_with_sapm

NameError: name 'normalize_with_sapm' is not defined
```

rdtools/__init__.py
mdeceglie: Yes it was intentional. Our intent here was that not importing those by default would help signal their experimental nature.

rdtools/availability.py
mdeceglie: The semantic versioning docs are pretty loud themselves. Must be an echo.  ;-)
https://semver.org/

rdtools/normalization.py
mdeceglie: This is a good question. We are thinking of moving in the direction of keeping modeling functionality out of RdTools. Viewing that more as in the wheelhouse of pvlib and other modeling frameworks. The idea is to keep the open source PV space as clean and organized as possible. But I'm hearing that you see value in some modeling capability within RdTools. I'm happy to keep that discussion open with the broader community.

rdtools/normalization.py
mdeceglie: I don't believe we have a citation for this function. It's always good to add useful examples. I'm not sure we'll get to it before release, but I'll open an issue so we keep this on the radar. Thanks!

rdtools/__init__.py
GregoryKimball: OK, I understand now.  Looks good.

rdtools/normalization.py
cwhanse: This is the same idea as is used in `pvlib.clearsky.detect_clearsky`, to scale the modeled clearsky irradiance to maximize the number of detected clearsky points.  The piece of code that finds `fact` (`alpha` in the pvlib function) will be separated into its own function in [this PR](https://github.com/pvlib/pvlib-python/pull/1074/). It's pretty short code, though, maybe not worth the effort of re-using it.

docs/degradation_example.ipynb
mdeceglie: The addition of soiling was the main change. Another substantial change was moving plotting stuff into a plotting module.  I made this diff to show all the changes if anyone interested: https://app.box.com/s/czynwmqimduibeduavgw75w4p5fe30xc

I agree that it is interesting to consider using soiling as part of normalization. I attempted this in https://ieeexplore.ieee.org/document/8981183 but the algorithm didn't generalize well to other cases and isn't solved in a self-consistent closed loop.

@asmunds came up with a very interesting and fundamentally different approach described here: https://ieeexplore.ieee.org/document/9186286 we'd like to get it into RdTools as well (#150)

mdeceglie:
Thank you for your input @GregoryKimball !

rdtools/normalization.py
```diff
@@ -209,175 +258,524 @@ def normalize_with_sapm(energy, sapm_kws):
...
+    # enforce max_timedelta
+    t_steps = t_steps.reindex(energy.index, method='backfill')
+    energy.loc[t_steps > max_interval_nanoseconds] = np.nan
+
+    energy.name = 'energy_Wh'
+
+    return energy
+
+
+def trapz_aggregate(time_series, target_frequency, max_timedelta=None):
```
cwhanse, line 578: seems like an opportunity to use `time_series.rolling(window).apply(np.trapz)` if `window = some_function(target_frequence)` can be done.

rdtools/availability.py
```diff
@@ -0,0 +1,620 @@
...
+            only applied to the subsystem loss calculation.
+        """
+        power_subsystem = self.power_subsystem
+        power_system = self.power_system
+        power_subsystem = power_subsystem.fillna(0)
+        power_system = power_system.clip(lower=0)
+
+        # Part A
+        if low_threshold is None:
+            low_threshold = power_subsystem.quantile(0.99) / 1000
```
cwhanse, line 251: `/ 1000` I initially assumed is a unit conversion but after `looks_online` I take it that `/ 1000` is a way to set a lower threshold, presumably `.quantile(0.01)` or similar is unreliable (could be 0 if there are outages)? Maybe `power_subsystem[power_subsystem.nonzero].quantile(0.01)` just a thought.

Maybe worth a comment.

rdtools/availability.py
```diff
@@ -0,0 +1,620 @@
...
+        | Column Name          | Description                                  |
+        +======================+==============================================+
+        | 'lost_production'    | Production loss from outages. Units match the|
+        |                      | input power units (e.g. if power is given in |
+        |                      | kW, 'lost_production' will be in kWh).       |
+        +----------------------+----------------------------------------------+
+        | 'actual_production'  | System energy production. Same units as      |
+        |                      | 'lost_production'.                           |
+        +----------------------+----------------------------------------------+
+        | 'availability'       | Energy-weighted system availability as a     |
```
cwhanse, line 84: This is just an FYI - NERC is developing reporting requirements for solar generators > 20MW capacity. In that context, "availability" and "outage" have very specific meanings. Down the road there will be a need for tools that compute the NERC-required metrics.

rdtools/availability.py
```diff
@@ -0,0 +1,620 @@
...
+import warnings
+
+warnings.warn(
+    'The availability module is currently experimental. The API, results, '
+    'and default behaviors may change in future releases (including MINOR '
+    'and PATCH releases) as the code matures.'
+)
+
+
+class AvailabilityAnalysis:
```
cwhanse, line 24: Cool to see this, and I understand the API.

rdtools/degradation.py
```diff
@@ -1,40 +1,40 @@
...
-    confidence_level: the size of the confidence interval to return, in percent
+    confidence_level: float, default 68.2
+        The size of the confidence interval to return, in percent.
 
     Returns
     -------
-    (degradation rate, confidence interval, calc_info)
-        calc_info is a dict that contains slope, intercept,
+    Rd_pct : float
+        Estimated degradation rate in units percent/year.
```
cwhanse, line 24: What is the base for the % here?

rdtools/soiling.py
```diff
@@ -0,0 +1,987 @@
...
+# Custom exception
+class NoValidIntervalError(Exception):
+    '''raised when no valid rows appear in the result dataframe'''
+    pass
+
+
+class SRRAnalysis():
+    '''
+    Class for running the stochastic rate and recovery (SRR) photovoltaic
+    soiling loss analysis presented in Deceglie et al. JPV 8(2) p547 2018
```
cwhanse, line 29: worth including the full reference somewhere. It's easier to search with a title.

rdtools/soiling.py
```diff
@@ -0,0 +1,987 @@
...
+        method : str, default 'half_norm_clean'
+            How to treat the recovery of each cleaning event:
+
+            * 'random_clean' - a random recovery between 0-100%
+            * 'perfect_clean' - each cleaning event returns the performance
+              metric to 1
+            * 'half_norm_clean' (default) - The three-sigma lower bound of
+              recovery is inferred from the fit of the following interval, the
+              upper bound is 1 with the magnitude drawn from a half normal
+              centered at 1
```
cwhanse, line 500: "half normal centered at 1" is this a truncated normal with upper bound at 1 and lower bound at the three-sigma value? If so, suggest "a random recovery sampled from a normal distributed with mean 1 and sigma = lower  bound / 3 (could use a name for this), truncated between 3*sigma and 1"

rdtools/soiling.py
```diff
@@ -0,0 +1,987 @@
...
+    'and PATCH releases) as the code matures.'
+)
+
+# Custom exception
+class NoValidIntervalError(Exception):
+    '''raised when no valid rows appear in the result dataframe'''
+    pass
+
+
+class SRRAnalysis():
```
cwhanse, line 26: I haven't worked fully through the API for this module, but again it's cool to see published algorithms in code.

cwhanse (commented):
Hope some of this is helpful. Nice to see the broadening of scope.

rdtools/normalization.py
```diff
@@ -1,63 +1,125 @@
...
+    '''
+    Normalize pv output based on expected PV power.
+
+    Parameters
+    ----------
+    pv : pd.Series
+        Right-labeled time series PV energy or power. If energy, should *not*
+        be cumulative, but only for preceding time step.
+    power_expected : pd.Series
+        Right-labeled time series of expected PV power.
```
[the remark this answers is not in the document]
mdeceglie: Agreed there was a disconnect here in the labeling convention/type. We've just pushed some updates that should help (#217). In particular `energy_from_power` is updated to properly handle either instantaneous or right labeled average power.

rdtools/normalization.py
```diff
@@ -1,63 +1,125 @@
...
+from rdtools._deprecation import deprecated
 
 class ConvergenceError(Exception):
+    '''Rescale optimization did not converge'''
     pass
 
+def normalize_with_expected_power(pv, power_expected, poa_global,
+                                  pv_input='power'):
+    '''
+    Normalize pv output based on expected PV power.
```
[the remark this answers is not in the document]
mdeceglie: I think that #217 should help with this to some extent, but I see that there is still room for improvement in the docs here. I will take a swing at that.

Regarding your suggestion to get rid of `pv_input`. That is an interesting suggestion and would simplify things for users. Thinking this through, we want to support the following use cases:

1. Power and expected power on the same or similar time scale (I think this is very common)
2. Energy and expected energy on the same or similar time scale.
3. Energy at low time resolution (e.g. weekly) but expected power at high time resolution (e.g 30 minute from a model using satellite data)

Perhaps we should make some wrappers for cases 1 and 2 (and maybe 3?) that use the present function in the correct way. We'd have to make some modifications to allow `power_expected` to be an energy though.

The insolation is needed for some subsequent calculations (especially soiling), and there is some value in it going through the exact same aggregation as `power_expected`. I think it would be good to make `poa_global` optional in the future (and not provide insolation in that case).

rdtools/normalization.py
```diff
@@ -143,23 +186,28 @@ def sapm_dc_power(pvlib_pvsystem, met_data):
 
     Parameters
     ----------
-    pvlib_pvsystem: pvlib-python LocalizedPVSystem object
+    pvlib_pvsystem : pvlib-python LocalizedPVSystem object
```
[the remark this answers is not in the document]
mdeceglie: Thanks for the heads up. We've decided to go with the first option suggested by @kanderso-nrel  in #198.

rdtools/normalization.py
mdeceglie: Thank you for the suggestion. Ultimately, this function got more complicated to handle either a a trapezoidal or rectangular sum in #217

docs/system_availability_example.ipynb
kanderso-nrel: RE the scope, the RdTools tag line has been updated to "RdTools is an open-source library to support reproducible technical analysis of time series data from photovoltaic energy systems."

>How does the algorithm handle partial day comms or inverter outages? How about overlapping outages?

The new availability analysis uses two methods.  The first goes timestamp-by-timestamp to compare inverter power data with system meter power data to distinguish inverter communication/data outages from real inverter power outages.  Because it goes timestamp-by-timestamp, it will detect and classify outages down to the order of the time resolution of the data you give it.  That method can handle multiple inverters offline at a time; it only requires one reporting inverter and the system meter.  The second method is used when the entire system looks to be offline (no inverters communicating production) and works on each outage in its entirety (in contrast to the first method that examines each timestamp in isolation).

rdtools/availability.py
kanderso-nrel: > I take it that / 1000 is a way to set a lower threshold

Correct -- not a unit conversion, just 0.1% of the 99th percentile.  It's done that way to defend against various low-power issues:  outages like you suggest, nighttime values, dawn/dusk shading, snow coverage... If there was a way to reliably filter out abnormal behavior then I'd be fine with `filtered_subset.quantile(0.01)` like you suggest, but that filter seems difficult to construct.  For example some inverters report small nonzero numbers (positive or negative) at night, so just checking `.nonzero` wouldn't work for them.   In contrast, it's much more likely for the high-power data to reflect normal behavior, so that seemed like the most reliable way to calculate this value.  The 99th percentile is just to avoid the rare non-physical data spike you sometimes see in field datasets. 

I agree it warrants a comment.  Thanks!

rdtools/availability.py
cwhanse: Let me see if I can find and share the draft spec

rdtools/soiling.py
mdeceglie: It isn't technically truncated on the lower edge. Perhaps it should be truncated based on the end of the last period, but for now I'll try to be more precise: 

> The starting point of each interval is taken randomly from a half normal distribution with its mode at 1 and its three-sigma point at the intercept of the fit to the interval.

rdtools/normalization.py
mdeceglie: That's something to keep an eye on. It looks to me like both still use their respective definition of clear sky detection internally, so reusing at this stage would be tricky. But if either gets more generalized we could consider reuse for this.

cwhanse: It's only [this short block of code](https://github.com/pvlib/pvlib-python/blob/3e25627e34bfd5aadea041da85a30626322b3a99/pvlib/clearsky.py#L763-L773) that will be pulled out into a helper function

rdtools/soiling.py
cwhanse: I understand now. Probably should drop the phrase "lower bound" in other places where 'half_norm_clean' is described. Perhaps in a future PR the sampling could be truncated to avoid the very rare event that soiling increases after the cleaning event, due to an extreme sample resulting in a negative recovery value. 

"mode at 1" is unambiguous but (splitting hairs a bit) here sigma is a distribution parameter rather than the standard deviation of the folded normal. "three-sigma point" might be misunderstood as three standard deviations of the folded normal. Maybe more clear to describe the distribution in terms of it's parameters, mu = 1 and sigma equal to 1/3 of the intercept of the fit to the interval.

mdeceglie: Thanks, that's a good suggestion. It's actually 1/3 * (1-intercept), I'll update each mention.

rdtools/normalization.py
mdeceglie: Ah I see, thank you for clarifying. Given the timelines involved, let's keep them separate for now, but keep this in mind for the future.

rdtools/normalization.py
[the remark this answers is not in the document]
mdeceglie: I've opened an issue about this so we keep it on the radar.

mdeceglie:
Thank you @cwhanse  and @GregoryKimball for your input! Those suggestions we didn't address in this release I've captured as issues.

Once #222 is merged, I think this is just about ready for a once over and a final run of the examples.

This PR was originally directed at release/2.0.0 to facilitate review of all changes relative to 1.2.3. Once we're satisfied, we can change the base for the pr to `master` and merge.

Thanks again everyone!

Issue #191, opened 2020-08-12
Minor notebook to rst conversion issue on RTD


kanderso-nrel:
A superscript footnote number loses the formatting when converting the source notebook to RST during the sphinx build:

https://rdtools.readthedocs.io/en/latest/example.html#5:-Soiling-calculations
![image](https://user-images.githubusercontent.com/57452607/90068959-c7a32700-dcae-11ea-9087-364ea2999198.png)

kanderso-nrel:
One way forward is using restructured text references, which survive the notebook->html conversion process and end up rendering nicely.  This specific example renders a little strangely because the leading `M.` and `G.` are interpreted as nested list elements instead of part of the citation text. 

If we do use this feature, we should consider using it for all references across the documentation because they are global.

https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#citations

mdeceglie:
Closed with #214

Issue #114, opened 2019-05-29
aggregation returns error with 1D frequency. '24h' frequency is ok.


cdeline:
This is weird.  When I'm running one of the Task 13 systems, I get a `ValueError: Values falls after last bin` after a call to rdtools.aggregation_insol.

The problem is fixed if I replace the `frequency = '1D'` with `frequency = '24h'`

Here's the traceback.  Recommend we change the default and notebook examples from 1D to 24h

> daily = rdtools.aggregation_insol(filtered.normalized, filtered.insolation, frequency = '1D')
> Traceback (most recent call last):
> 
>   File "<ipython-input-56-aadfc5a5167a>", line 1, in <module>
>     daily = rdtools.aggregation_insol(filtered.normalized, filtered.insolation, frequency = '1D')
> 
>   File "c:\users\cdeline\documents\python scripts\kwh_rdtools\rdtools\aggregation.py", line 23, in aggregation_insol
>     aggregated = (insolation * normalized_energy).resample(frequency).sum() / insolation.resample(frequency).sum()
> 
>   File "C:\Users\cdeline\anaconda3\lib\site-packages\pandas\core\resample.py", line 855, in f
>     return self._downsample(_method, min_count=min_count)
> 
>   File "C:\Users\cdeline\anaconda3\lib\site-packages\pandas\core\resample.py", line 999, in _downsample
>     self._set_binner()
> 
>   File "C:\Users\cdeline\anaconda3\lib\site-packages\pandas\core\resample.py", line 173, in _set_binner
>     self.binner, self.grouper = self._get_binner()
> 
>   File "C:\Users\cdeline\anaconda3\lib\site-packages\pandas\core\resample.py", line 181, in _get_binner
>     binner, bins, binlabels = self._get_binner_for_time()
> 
>   File "C:\Users\cdeline\anaconda3\lib\site-packages\pandas\core\resample.py", line 988, in _get_binner_for_time
>     return self.groupby._get_time_bins(self.ax)
> 
>   File "C:\Users\cdeline\anaconda3\lib\site-packages\pandas\core\resample.py", line 1421, in _get_time_bins
>     ax_values, bin_edges, self.closed, hasnans=ax.hasnans)
> 
>   File "pandas\_libs\lib.pyx", line 691, in pandas._libs.lib.generate_bins_dt64
> 
> ValueError: Values falls after last bin

cwhanse:
[...]

PR #215, opened 2020-10-13
Issue cleanup for 2.0.0 release, part 1

kanderso-nrel:
- ~New functions added to `__init__.py`~  no new functions
- [x] Updated changelog

Old issue cleanup for 2.0.0.  This addresses #191, #114, and #198. 

For #114, I tested each pandas release from 0.23.2 to 1.0.0 and the only one with the issue was 0.24.0.  Rather than bump the minimum version, I just excluded that specific version.

The complexity of the deprecation machinery is worth it IMHO -- it automatically modifies docstrings with a deprecation message, which then shows up in IDEs as well as propagating to the sphinx docs: 

image

---

image

---

image


mdeceglie (approved):
LGTM!

PR #217, opened 2020-10-14
195 timestamp labeling

mdeceglie:
- [x] Updated changelog


rdtools/normalization.py
```diff
@@ -567,7 +551,7 @@ def energy_from_power(power, target_frequency=None, max_timedelta=None):
     return energy
 
 
-def trapz_aggregate(time_series, target_frequency, max_timedelta=None):
+def _aggregate(time_series, target_frequency, max_timedelta=None, series_type='right_labeled'):
     '''
     Returns a right-labeled series with frequency target_frequency generated by
     aggregating `time_series` with the trapezoidal rule (in units of hours).
```
kanderso-nrel, line 557: Docstring updates: update this line so it says the method depends on `series_type`, also add `series_type` to the parameter description list

rdtools/normalization.py
```diff
@@ -602,13 +586,16 @@ def trapz_aggregate(time_series, target_frequency, max_timedelta=None):
         max_interval_nanoseconds = max_timedelta.total_seconds() * 10.0**9
```
kanderso-nrel, line 586: Now that it's a private function, do we need to allow `max_timedelta` to be None?  I think `energy_from_power` will always provide the appropriate value.

rdtools/normalization.py
```diff
@@ -602,13 +586,16 @@ def trapz_aggregate(time_series, target_frequency, max_timedelta=None):
...
 
     # in x*hours
-    trap_sum = (values[1:] + values[:-1]) / 2 * t_diffs / 10**9 / 3600.0
+    if series_type is 'instantaneous':
+        series_sum = (values[1:] + values[:-1]) / 2 * t_diffs / 10**9 / 3600.0
+    elif series_type is 'right_labeled':
+        series_sum = values[1:] * t_diffs / 10**9 / 3600.0
```
kanderso-nrel, line 592:
```suggestion
        series_sum = values[1:] * t_diffs / 10**9 / 3600.0
    else:
        raise ValueError("series_type must be either 'instantaneous' or 'right_labeled', "
                         "not '{}'".format(series_type))
```

rdtools/normalization.py
```diff
@@ -602,13 +586,16 @@ def trapz_aggregate(time_series, target_frequency, max_timedelta=None):
         max_interval_nanoseconds = max_timedelta.total_seconds() * 10.0**9
 
     # in x*hours
-    trap_sum = (values[1:] + values[:-1]) / 2 * t_diffs / 10**9 / 3600.0
+    if series_type is 'instantaneous':
+        series_sum = (values[1:] + values[:-1]) / 2 * t_diffs / 10**9 / 3600.0
+    elif series_type is 'right_labeled':
```
kanderso-nrel, line 591: Using `is` for string equality would probably work in the vast majority of cases here, but it does have some weird edge cases.  Would be safer to use `==`.  Example:

```python
In [13]: "instantaneous_"[:-1] is "instantaneous"
Out[13]: False

In [14]: "instantaneous_"[:-1] == "instantaneous"
Out[14]: True
```

rdtools/normalization.py
```diff
@@ -602,13 +586,16 @@ def trapz_aggregate(time_series, target_frequency, max_timedelta=None):
         max_interval_nanoseconds = max_timedelta.total_seconds() * 10.0**9
 
     # in x*hours
-    trap_sum = (values[1:] + values[:-1]) / 2 * t_diffs / 10**9 / 3600.0
+    if series_type is 'instantaneous':
+        series_sum = (values[1:] + values[:-1]) / 2 * t_diffs / 10**9 / 3600.0
```
kanderso-nrel, line 590: Nit, `t_diffs * 10**9 / 3600.0` is used twice, so might as well factor it out into `t_diffs_hours` or something

rdtools/normalization.py
kanderso-nrel: Also need to drop `trapz_aggregate` from `api.rst`

rdtools/normalization.py
mdeceglie: Not quite. When I made the change it blew up a lot of tests.`max_timedelata` get's passed straight in, even if it is `None`. However this caused me to notice that `_aggregate()` and `interpolate()` have slightly different default behavior when they get `max_timedelta=None` and that is undesirable for this application. Better to be specific. So I modified `energy_from_power` to calculate and pass `max_timedelta`.

mdeceglie:
Given the change log consolidation coming down the pike in in #218 I don't think we need to add anything here.

mdeceglie:
The results in the example notebooks do change slightly. A summary of the changes can be viewed by downloading the html diffs here: https://app.box.com/s/k8kgxvbu1c26hup8b1f1oxyesbuhs438

cdeline:
Thanks for sharing those Mike - it looks like RdTools confidence intervals tighten up somewhat, which is nice.  I think that this is a good change / capability to distinguish center-labeled from right-labeled data.  It's a somewhat minor point when it comes to the overall process, but good to have it.

kanderso-nrel (commented):
It took a minute to wrap my head around what `_aggregate` is doing, but I think it makes sense to me now.  For the availability notebook, I'm ok with the differences, but it looks like an import warning got lost because the cells were re-run without restarting the kernel.  If we plan on doing a "final run" of the notebooks before cutting the release then NBD. 

Couple points below, otherwise LGTM

rdtools/normalization.py
```diff
@@ -486,6 +488,8 @@ def energy_from_power(power, target_frequency=None, max_timedelta=None):
         returned for that interval. If omitted, `max_timedelta` is set
         internally to the median time delta in `power`. Ignored when `power`
         has fewer than two elements.
+    power_type : {'right_labeled', 'instantaneous'}
```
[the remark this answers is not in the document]
mdeceglie: Oh, good point. Either raise or point out that power_type is ignored for single-value input. Preference?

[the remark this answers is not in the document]
kanderso-nrel: I guess preference for raising.  Otherwise you get inconsistent results if you pass in a 2-element input and then only pass in the last element, both with "instantaneous" -- seems like you should get the same last value either way.

kanderso-nrel (approved):
LGTM!

cdeline:
Yay!

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #218, opened 2020-10-14
Issue cleanup for 2.0.0 release, part 3

kanderso-nrel:
- [x] Updated changelog

Addressing comments raised in review of #214.  FYI the description for older releases in the version history came from here: https://github.com/NREL/rdtools/releases


docs/sphinx/source/conf.py
```diff
@@ -53,7 +53,8 @@
 # directories to ignore when looking for source files.
 # This pattern also affects html_static_path and html_extra_path.
 
-exclude_patterns = []
+exclude_patterns = ['changelog/*']
```
kanderso-nrel, line 56: This just prevents sphinx from emitting `v2.0.0.rst: WARNING: document isn't included in any toctree`

docs/sphinx/source/changelog/v2.0.0.rst
```diff
@@ -1,6 +1,6 @@
-************************
-v2.0.0b0 (July 31, 2020)
-************************
+*************************
+v2.0.0 (October 15, 2020)
```
mdeceglie, line 2:
```suggestion
v2.0.0 (TBD, 2020)
```

setup.py
```diff
@@ -36,14 +36,14 @@
...
-    'pandas >= 0.23.0,!=0.24.0,!=1.0.0,!=1.0.1',
-    'statsmodels >= 0.8.0',
-    'scipy >= 0.19.1',
-    'h5py >= 2.7.1',
-    'pvlib >= 0.7.0, <0.8.0',
+    'pandas >= 0.23.0,!=0.24.0,!=1.0.0,!=1.0.1,<2.0.0',
+    'statsmodels >= 0.8.0,<1.0.0',
+    'scipy >= 0.19.1,<2.0.0',
+    'h5py >= 2.7.1,<3.0.0',
+    'pvlib >= 0.7.0, <0.9.0',
```
mdeceglie, line 46: I'm not sure we need to be restrictive for packages without known issues (that are beyond version 1). Travis will catch a breaking change in a new version of a dependency and let us know. Pandas for example doesn't seem to be restrictive: https://github.com/pandas-dev/pandas/blob/master/setup.py 

The goal of setup.py is to put some guardrails up but leave pip and conda (for example) as much leeway as possible to solve environments.

kanderso-nrel: Well I checked astropy and metpy and nobody uses upper limits like this, so maybe you're right.  I'm imagining a scenario where one of our dependencies releases a new major version after we cut this release and when someone goes to install rdtools in a fresh environment, they get strange errors.  And since we already cut the release, there's nothing we can do about it except cut a fresh one with a fix, but that doesn't help people wanting to use the older release.

I guess our docs make a point of recommending that users use `requirements.txt` to make sure everything works.  I'll revert these upper bounds.

mdeceglie (approved):
Looks good pending the setup req changes. Thanks!

cdeline:
🎉

PR #221, opened 2020-10-19
Release updates md2

mdeceglie:
- [x] Updated changelog

### DIFF
--- a/rdtools/normalization.py
+++ b/rdtools/normalization.py
@@ -456,7 +456,7 @@ def check_series_frequency(series, series_description):
     return freq
 
 
-def t_step_nanoseconds(time_series):
+def _t_step_nanoseconds(time_series):
     '''
     return a series of right labeled differences in the index of time_series
     in nanoseconds
@@ -522,7 +522,7 @@ def energy_from_power(power, target_frequency=None, max_timedelta=None, power_ty
         energy.name = 'energy_Wh'
         return energy
 
-    t_steps = t_step_nanoseconds(power)
+    t_steps = _t_step_nanoseconds(power)
     median_step_ns = t_steps.median()
 
     if target_frequency is None:
@@ -544,7 +544,7 @@ def energy_from_power(power, target_frequency=None, max_timedelta=None, power_ty
                                      power.index[-1],
                                      freq=target_frequency)
             temp_series = pd.Series(data=1, index=temp_ind)
-            temp_diffs = t_step_nanoseconds(temp_series)
+            temp_diffs = _t_step_nanoseconds(temp_series)
             freq_interval_size_ns = temp_diffs.median()
         else:
             raise
@@ -645,7 +645,7 @@ def _aggregate(time_series, target_frequency, max_timedelta, series_type):
     return aggregated
 
 
-def interpolate_series(time_series, target_index, max_timedelta=None,
+def _interpolate_series(time_series, target_index, max_timedelta=None,
                        warning_threshold=0.1):
     '''
     Returns an interpolation of time_series onto target_index, NaN is returned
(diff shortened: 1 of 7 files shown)


rdtools/normalization.py
```diff
@@ -14,20 +14,22 @@ class ConvergenceError(Exception):
...
-        Right-labeled time series of expected PV power.
+        Right-labeled time series of expected PV power. (Note: Expected energy
+        is not supported.)
     poa_global : pd.Series
         Right-labeled time series of plane-of-array irradiance associated with
         `expected_power`
-    pv_input : str
-        'power' or 'energy' to specify type of input used for pv parameter
+    pv_input : {'power' or 'energy'}
+        Specifies the type of input used for pv parameter. Default: 'power'
```
kanderso-nrel, line 32:
```suggestion
    pv_input : {'power' or 'energy'}, default 'power'
        Specifies the type of input used for `pv` parameter.
```

rdtools/normalization.py
```diff
@@ -14,20 +14,22 @@ class ConvergenceError(Exception):
...
-    Normalize pv output based on expected PV power.
+    Normalize PV power or energy based on expected PV power.
 
     Parameters
     ----------
     pv : pd.Series
         Right-labeled time series PV energy or power. If energy, should *not*
-        be cumulative, but only for preceding time step.
+        be cumulative, but only for preceding time step. Type (energy or power)
+        must be specified in the pv_input parameter.
```
kanderso-nrel, line 24:
```suggestion
        must be specified in the `pv_input` parameter.
```

kanderso-nrel (commented):
Took me a second to understand how to read "Estimated degradation rate in units percent/year relative to energy_normalized=1." -- would something like this be better?

> Estimated annual degradation rate as a fraction of year 0 system capacity [%/yr]

Changelog updates look good to me, thanks for merging those stragglers.

rdtools/normalization.py
mdeceglie: https://numpydoc.readthedocs.io/en/latest/format.html Section 4 seems to indicate that the default should be in the description. But maybe I'm misinterpreting?

kanderso-nrel: Hmm I thought `paramname : type, default X` was fairly standard but I'm not finding many examples to support that in the big packages.  Only one I found so far is the `axis` parameter here, but it's not really apples to apples: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.shift.html

Okay with me to follow the numpydoc guidelines.  I do still think `pv` needs the backticks.

mdeceglie:
Regarding the degradation rate,
> Estimated annual degradation rate as a fraction of year 0 system capacity [%/yr]

will work for ols and classical decomposition. But not for YOY with `recenter=False` If recenter is off, the result depends is strictly relative to normalized_energy=1. Is there a more clear way to describe that case?
```
In [1]: import pandas as pd
   ...: import rdtools
   ...: 
   ...: x = pd.Series([1, 0.9, 0.8, 0.7], index=pd.date_range('01/01/01', periods=4, freq="AS"))
   ...: 

In [2]: rdtools.degradation_year_on_year(x)
Out[2]: 
(-9.999999999999998, array([-10., -10.]), {'YoY_values': dt
  2002-01-01   -10.0
  2003-01-01   -10.0
  2004-01-01   -10.0
  Name: yoy, dtype: float64,
  'renormalizing_factor': 1.0,
  'exceedance_level': -10.000000000000009})

In [3]: rdtools.degradation_year_on_year(10*x)
Out[3]: 
(-9.999999999999998, array([-10., -10.]), {'YoY_values': dt
  2002-01-01   -10.0
  2003-01-01   -10.0
  2004-01-01   -10.0
  Name: yoy, dtype: float64,
  'renormalizing_factor': 10.0,
  'exceedance_level': -10.000000000000009})

In [4]: rdtools.degradation_year_on_year(10*x, recenter=False)
Out[4]: 
(-100.0, array([-100., -100.]), {'YoY_values': dt
  2002-01-01   -100.0
  2003-01-01   -100.0
  2004-01-01   -100.0
  Name: yoy, dtype: float64,
  'renormalizing_factor': 1.0,
  'exceedance_level': -100.0})
```

mdeceglie:
Let's go with @kanderso-nrel 's suggestion and describe the `recenter=False` case in more detail in the description for that parameter.

kanderso-nrel (approved):
LGTM

PR #222, opened 2020-10-19
make a few more functions private

mdeceglie:
- [ ] Updated changelog

### DIFF
--- a/docs/sphinx/source/changelog/v2.0.0.rst
+++ b/docs/sphinx/source/changelog/v2.0.0.rst
@@ -10,8 +10,6 @@ API Changes
   right-labeled energy with :py:func:`~rdtools.normalization.energy_from_power`
   before being used with these normalization functions (:pull:`105`, :pull:`108`).
 * Remove ``low_power_cutoff`` parameter in :py:func:`~rdtools.filtering.clip_filter` (:issue:`84`).
-* :py:func:`~rdtools.normalization.delta_index` and :py:func:`~rdtools.normalization.check_series_frequency`
-  made private functions. (:pull:`222`)
 * Many kwargs have changed name (but not input order) to bring nomenclature into 
   closer alignment with the `DuraMAT pv-terms project <https://duramat.github.io/pv-terms>`_: (:pull:`185`)
 
@@ -37,6 +35,7 @@ Deprecations
   :py:func:`~rdtools.normalization.normalize_with_sapm` have been deprecated
   in favor of :py:func:`~rdtools.normalization.normalize_with_expected_power`.
   (:pull:`215`)
+* :py:func:`~rdtools.normalization.delta_index` and :py:func:`~rdtools.normalization.check_series_frequency` (:pull:`222`)
 
 Enhancements
 ------------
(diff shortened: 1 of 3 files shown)


docs/sphinx/source/changelog/v2.0.0.rst
```diff
@@ -10,6 +10,8 @@ API Changes
   right-labeled energy with :py:func:`~rdtools.normalization.energy_from_power`
   before being used with these normalization functions (:pull:`105`, :pull:`108`).
 * Remove ``low_power_cutoff`` parameter in :py:func:`~rdtools.filtering.clip_filter` (:issue:`84`).
+* :py:func:`~rdtools.normalization.delta_index` and :py:func:`~rdtools.normalization.check_series_frequency`
+  made private functions. (:pull:`222`)
```
kanderso-nrel, line 14: Oh, didn't realize these functions weren't new in 2.0.  I think the Right Thing To Do would be to deprecate them and remove in 3.0, but maybe we don't think anybody would care if they get removed without deprecation?

mdeceglie:
@kanderso-nrel  Do you think we should include the underscores in the change log?

kanderso-nrel:
> @kanderso-nrel Do you think we should include the underscores in the change log?

I don't think we should advertise private functions in the documentation, if that's what you mean?  I think `Function X is now a private function` and `Function X is removed from the public API` are fine, but not `Function X was renamed to _X`.

docs/sphinx/source/changelog/v2.0.0.rst
mdeceglie: Ah good point. Yes, let's do The Right Thing.

mdeceglie: Actually this causes the functions that rely on them to emit deprecation warnings. If there isn't an easy way around that, I don't think it's a big deal to make them private. I'd be surprised if anyone is using them, they were already omitted from `__init__.py`

mdeceglie: We could catch them where they are called (https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings) But I'm not sure it's worth the maintenance. I could go either way, preference?

kanderso-nrel: I think the `@deprecated` decorator can handle that.  Make the function private, switch any callers to using the new private name, and make a deprecated public version that just wraps the private one (calling the decorator directly instead of using it as a decorator, it's a little weird but it works):

```python
from rdtools._deprecation import deprecated

def _delta_index(...):
    ....

delta_index = deprecated('2.0.0')(_delta_index)
```
So users calling `delta_index` will get the deprecated warning, but our functions that call `_delta_index` won't. I didn't test it but in theory it should work...

kanderso-nrel (approved):
LGTM, I don't see any warnings (from this) in the travis logs and sphinx shows the deprecation warnings  :+1:

cdeline:
Looks good guys. 👍

cdeline:
🎉

PR #225, opened 2020-10-21
Fix travis PyPI deploy condition

kanderso-nrel:
-  ~Code changes are covered by tests~
-  ~API.rst is up to date, along with other sphinx docs pages~
-  ~Example notebooks are rerun and differences in results scrutinized~
-  ~Updated changelog~

Fixes an unintentional change I made in #197 that prevented travis from deploying tagged commits to PyPI.

@mdeceglie I think it makes sense to merge this straight to master instead of development, do you agree?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ python:
 env:
   - REQ_ENV='-r requirements.txt .[test]'
   - REQ_ENV='-r requirements-min.txt .[test]'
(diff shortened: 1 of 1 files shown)


cdeline:
I agree.  We shouldn't be shy about deploying bug fixes like this to Master

mdeceglie (approved):
LGTM no objection to merging straight to master

PR #226, opened 2020-10-21
Fix test suite dtype error on Windows

kanderso-nrel:
- ~Updated changelog~

Apparently the default long int width is 32 bits on 64-bit Windows but 64 bits on the other main OSs.  That causes a couple tests to fail when I run the test suite locally on Windows (the package returns int64 but the expected values are int32):

```python
>       pd.testing.assert_frame_equal(result, expected)
E       AssertionError: Attributes of DataFrame.iloc[:, 0] (column name="year") are different
E
E       Attribute "dtype" are different
E       [left]:  int64
E       [right]: int32
rdtools\test\soiling_test.py:288: AssertionError
```

This PR updates the dtypes in the failing tests to specify `np.int64` instead of the platform-dependent default `int`.  Reference: https://stackoverflow.com/a/36279549

Seems too minor to be worth mentioning in the changelog, but I'll add it if requested.

### DIFF
--- a/rdtools/test/soiling_test.py
+++ b/rdtools/test/soiling_test.py
@@ -280,12 +280,12 @@ def test_annual_soiling_ratios(multi_year_profiles):
         [2019, 14.5, 11.431, 17.569]])
     expected = pd.DataFrame(data=expected_data,
         columns=['year', 'soiling_ratio_median', 'soiling_ratio_low', 'soiling_ratio_high'])
-    expected['year'] = expected['year'].astype(np.int64)
+    expected['year'] = expected['year'].astype(int)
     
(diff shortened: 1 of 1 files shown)


cdeline:
Kevin - I fixed this in a different way. on line 288 and 301 of soiling_test.py you can add `, check_dtype=False`  which was present on some of the other test cases, but not this one.  I found it to work then.

mdeceglie (approved):
LGTM. Tests pass on osx.

mdeceglie:
I think it's fine without the change log entry, I will proceed to merge. Thanks @kanderso-nrel and @cdeline

PR #230, opened 2020-10-30
GitHub actions

mdeceglie:
- ~Updated changelog~


mdeceglie:
We won't know whether deploy works until a release is published, but it follows the example almost exactly. Otherwise I think this is good to go.

mdeceglie:
I didn't update the changelog, but can if we think it will be useful.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,35 @@
+name: test
+
+on: [pull_request]
```
kanderso-nrel, line 3: We should think about including `push` in this list.  My understanding is that using the `pull_request` event will try to merge this PR into the base branch during the checkout ([example](https://github.com/NREL/rdtools/runs/1333729136?check_suite_focus=true#step:2:204)), which is usually what we want, but I assume it'll fail/not run when there's a merge conflict.  Adding `push` would mean that a second set of tests run just on the PR branch without trying to merge.  Travis does this ("PR" vs "push").

I think `push` would mean we also get an independent CI run for each commit pushed to master, which I think would be nice.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,35 @@
+name: test
```
kanderso-nrel, line 1: Maybe a more descriptive name here, something with "suite" or "pytest" or something?  Here's what the checks interface currently looks like:

image

.github/workflows/test.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+on: [pull_request]
+
+jobs:
+  build:
```
kanderso-nrel, line 6: Consider a better name here too, maybe even just "pytest" since that's the main point of this job?

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,28 @@
...
+  deploy:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
```
kanderso-nrel, line 17: Consider specifying an explicit version here -- I can't find the exact rules for what `3.x` means, but if it means "the most recent version", it might end up using a version we don't explicitly support yet.  For instance they already added python 3.9 to GH actions but we aren't formally testing on it yet.

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,28 @@
...
+      run: |
+        python -m pip install --upgrade pip
+        pip install setuptools wheel twine
+    - name: Build and publish
+      env:
+        TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
+        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+      run: |
+        python setup.py sdist bdist_wheel
+        twine upload dist/*
```
kanderso-nrel, line 28: No opinion from me on whether it's better or worse than using twine manually, but PyPA provides an action for PyPI deploys: https://github.com/pypa/gh-action-pypi-publish

kanderso-nrel: Could also consider adding another step for saving the dist files as job artifacts -- might help with debugging if something weird happens with the PyPI deploy.   Something like: 

```
      - uses: actions/upload-artifact@v2
        with:
          path: dist/*
```

kanderso-nrel (commented):
Looks pretty good!  Some suggestion below

.github/workflows/test.yaml
mdeceglie: I changed the filename and top level name, but adding pytest here feels redundant

.github/workflows/test.yaml
mdeceglie: Interesting. We could modify the behavior like this as well: https://github.com/actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit but I think doing both is ultimately what we want like you say.

.github/workflows/deploy.yaml
mdeceglie: I noticed that pypi action as well. Ultimately I thought the explicit twine was just cleaner. I followed this example: https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-python#publishing-to-package-registries 

Good idea on the artifacts. I have added that.

kanderso-nrel (approved):
LGTM, it'll be nice to have fast tests again!

PR #231, opened 2020-11-02
flake8 compliance; flake8 GH Actions workflow

kanderso-nrel:
- [x] Updated changelog

Most of the diff is self-explanatory, but shifting fixtures around deserves an explanation:  if a fixture is needed in multiple files, it is recommended to [house it in conftest.py](https://docs.pytest.org/en/stable/fixture.html#conftest-py-sharing-fixture-functions), which automatically makes it available to all test files.  I made that change here because flake8 took significant offense at us directly importing across test files, but we probably should have been doing it anyway. 

FYI: I set the max line length to 99 characters as a starting point but that's up for discussion if anyone prefers a different value.


mdeceglie:
Since we do have a testing section of the change log, I think it would make sense to add something now that we have new testing scope. Also some additional breadcrumbs about the shared fixtures in the changelog might be good. Other than that, looks good to me.

mdeceglie (approved):

kanderso-nrel:
From internal discussion with @mdeceglie we decided to limit the lint check to PR diffs (instead of the entire codebase) so that we can choose to merge non-compliant code (e.g. test files) without causing irrelevant failures in future PRs.  

The new workflow configuration is a little more complex, but it seems to work: https://github.com/NREL/rdtools/pull/231/checks?check_run_id=1381245386#step:6:8

cdeline (approved):
Yep, this looks like a good addition to have a linter check pull requests.  I'll probably need help to modify my analysis_test.py to match the new organization of test fixtures.

kanderso-nrel:
@mdeceglie I rebased this on `release/2.0.5`.  I guess we can cherrypick (or just recreate) the analysis_chains fixes in a separate PR to development.

PR #232, opened 2020-11-03
Model chains exp energy

cdeline:
- [x] Updated changelog

### DIFF
--- a/docs/RdAnalysis_example.ipynb
+++ b/docs/RdAnalysis_example.ipynb
@@ -727,7 +727,7 @@
    "metadata": {},
    "source": [
     "## Substituting a different modeled power\n",
-    "Incidentally, you can optionally run the analysis while normalizing with your own calculated expected power.  By default, `RdAnalysis` normalizes by the included `pvwatts_dc_power` calculation.  To override with a different value, manually set `RdAnalysis.power_expected`. You can substitute any modeled power here, like from SAM or PVSyst or PVLib.  To return to default values, set `RdAnalysis.power_expected = None`"
+    "Incidentally, you can optionally run the analysis while normalizing with your own calculated expected power.  By default, `RdAnalysis` normalizes by the included `pvwatts_dc_power` calculation.  To override with a different value, pass in `RdAnalysis.power_expected`. You can substitute any modeled power here, like from SAM or PVSyst or PVLib.  "
    ]
   },
   {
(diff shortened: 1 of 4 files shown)


cdeline:
@kanderso-nrel This now addresses your < 2 yr data post-filter concern in #117

rdtools/analysis.py
```diff
@@ -455,23 +470,31 @@ def srr_soiling(self, aggregated, aggregated_insolation, **kwargs):
...
+        Perform sensor-based normalization, filtering, and aggregation.
+        If self.power_expected is set manually prior to this process, 
+        normalize_with_expected_power will be used instead of pvwatts.
         '''
         if self.poa is None:
             raise ValueError('poa must be available to perform sensor_preprocess')
         if self.cell_temperature is None and self.ambient_temperature is None:
             raise ValueError('either cell or ambient temperature must be available to perform sensor_preprocess')
         if self.cell_temperature is None:
             self.cell_temperature = self.calc_cell_temperature(self.poa, self.windspeed, self.ambient_temperature)
```
kanderso-nrel, line 482: Are the thermal model inputs needed for anything if the user provides expected power?  If not then maybe these should be moved into the `if self.power_expected is None` block.

cdeline:
oops, looks like the 2-yr check would be better off in the YoY degradation section - soiling SRR doesn't require 2 years of data, so it's better to move the error checking closer to where it's required.

docs/RdAnalysis_example.ipynb
```diff
@@ -721,6 +721,58 @@
...
+   "outputs": [],
+   "source": [
+    "power_expected = rdtools.normalization.pvwatts_dc_power(poa_global=rd.poa,\n",
+    "                                                           power_dc_rated=meta['power_dc_rated'], \n",
+    "                                                           temperature_cell=rd.cell_temperature,\n",
+    "                                                           gamma_pdc=meta['gamma_pdc'])\n",
+    "\n",
+    "rd_with_expected_power = rdtools.RdAnalysis(pv=df['power'], poa=df['poa'],                              \n",
+    "                             power_expected = power_expected,\n",
+    "                             cell_temperature=rd.cell_temperature,\n",
```
kanderso-nrel, line 746: I think cell temp could be removed here, or is it left in to keep the filtering identical?

cdeline: Yeah, prior to commit 6101b2b it would throw an error, but it can be removed now.  Including cell_temperature would ensure the filtering was identical if there were any unphysical Tcell values (which wasn't the case with this dataset).

kanderso-nrel (approved):
LGTM -- the new `filter_check` method isn't really part of the primary user interface so no need to add it to api.rst IMO.  I could go either way on whether this needs a changelog entry, but would lean towards just rolling it into the one for #117

cdeline:
Merging this.

PR #233, opened 2020-11-09
Model chains set clearsky

cdeline:
- [x] Updated changelog


cdeline:
After conversations with Mike and Kevin, it sounds like removing some input parameters from the _init_ of RdAnalysis and making a separate set_clearsky function would streamline the basic RdAnalysis usage. At a minimium, pvlib_loc and tilt and azimuth must be passed to set_clearsky.  Optional input parameters include externally calculated clearsky_poa and clearsky temperatures, but these will be modeled if not explicitly passed.

cdeline:
P.S. running `set_clearsky` is MANDATORY prior to running `clearsky_analysis`.  Functionally, it doesn't really change the workflow at all - you just set the clearsky related variables in this function rather than all at once in the _init_.

rdtools/test/analysis_test.py
```diff
@@ -82,40 +92,43 @@ def test_sensor_analysis_exp_power(sensor_analysis_exp_power):
...
+def clearsky_optional(cs_input, clearsky_analysis):
     # optional parameters to exercise other branches
     times = clearsky_analysis.poa.index
     extras = dict(
         clearsky_poa=clearsky_analysis.clearsky_poa,
         clearsky_cell_temperature=clearsky_analysis.clearsky_cell_temperature,
         clearsky_ambient_temperature=clearsky_analysis.clearsky_ambient_temperature,
-        pv=clearsky_analysis.pv_energy,
-
+        #pv=clearsky_analysis.pv_energy,  # todo: set up a test for pv_energy
```
kanderso-nrel, line 124: Is this still a to-do?

rdtools/test/analysis_test.py
```diff
@@ -82,40 +92,43 @@ def test_sensor_analysis_exp_power(sensor_analysis_exp_power):
...
+        
         # series orientation instead of scalars to exercise interpolation
-        pv_tilt=pd.Series(clearsky_parameters['pv_tilt'], index=times),
-        pv_azimuth=pd.Series(clearsky_parameters['pv_azimuth'], index=times)
+        pv_tilt=pd.Series(cs_input['pv_tilt'], index=times),
+        pv_azimuth=pd.Series(cs_input['pv_azimuth'], index=times)
     )
     # merge dicts, favoring new params over the ones in clearsky_parameters
-    return {**clearsky_parameters, **extras}
+    return {**extras}
```
kanderso-nrel, line 131: Suggest `return extras` here if there is no need to merge dicts

rdtools/analysis.py
```diff
@@ -155,6 +124,55 @@ def __init__(self, pv, poa=None, cell_temperature=None, ambient_temperature=None
         if power_expected is not None and cell_temperature is None:
             del self.filter_params['tcell_filter']
 
+    def set_clearsky(self, pvlib_location=None, pv_azimuth=None, pv_tilt=None,
+                 clearsky_poa=None, clearsky_cell_temperature=None, 
+                 clearsky_ambient_temperature=None, albedo=0.25 ):
```
kanderso-nrel, line 129: Nit, continuation indentation to get a head start on #231

rdtools/analysis.py
```diff
@@ -34,33 +34,19 @@ class RdAnalysis():
...
         Right-labeled time series of expected PV power. (Note: Expected energy
         is not supported.)
     temperature_model : str
         Model parameter pvlib.pvsystem.sapm_celltemp() used in calculating cell
         temperature from ambient
-    pv_azimuth : numeric
-        Azimuth of PV array in degrees from north
-    pv_tilt : numeric
-        Tilt of PV array in degrees from horizontal
+
```
kanderso-nrel, line 49: Having blank lines here but not elsewhere makes the docstring look a little wonky

kanderso-nrel (commented):
I think I like this idea; the `__init__` is certainly less intimidating.  For anyone interested, I've seen this approach of having a sort of staged object construction with "add-on" parameters called the "fluent interface pattern", e.g. https://martinfowler.com/bliki/FluentInterface.html.  Probably there are other names for it too. 

@cdeline what do you think about raising an error if the user calls `clearsky_analysis` without having called `set_clearsky` first?  Right now it emits `AttributeError: 'RdAnalysis' object has no attribute 'clearsky_poa'`.  

Not related to this PR, just something I noticed: `Rd_analysis.ipynb` has a python 2.7 check that I think we can remove.  cell 3, the try/except when renaming dataframe columns

mdeceglie:
Haven't done a detailed review yet, but I like the idea of raising an informative error as @kanderso-nrel  suggests.

rdtools/test/analysis_test.py
cdeline: Test coverage dropped from 91% to 90%, but it doesn't appear that this section was actually increasing coverage any (?)

rdtools/analysis.py
```diff
@@ -155,6 +121,55 @@ def __init__(self, pv, poa=None, cell_temperature=None, ambient_temperature=None
...
+                     clearsky_poa=None, clearsky_cell_temperature=None, 
+                     clearsky_ambient_temperature=None, albedo=0.25 ):
+        '''
+        Initialize values for a clearsky analysis which requires configuration 
+        of location and orientation details. If optional parameters `clearsky_poa`,
+        `clearsky_ambient_temperature` are not passed, they will be modeled
+        based on location and orientation.
+        
+        
+        pvlib_location : pvlib.location.Location
```
kanderso-nrel, line 134:
```suggestion
        Parameters
        ----------
        pvlib_location : pvlib.location.Location
```

rdtools/test/analysis_test.py
kanderso-nrel: I think it would hit the block at line 94 (`elif pv_input == 'energy':`), but not a big deal.

rdtools/analysis.py
```diff
@@ -177,9 +192,9 @@ def calc_clearsky_poa(self, times=None, rescale=True, **kwargs):
         if times is None:
             times = self.poa.index
         if self.pvlib_location is None:
-            raise ValueError('pvlib location must be provided')
+            raise Exception('pvlib location must be provided using set_clearsky()')
         if self.pv_tilt is None or self.pv_azimuth is None:
-            raise ValueError('pv_tilt and pv_azimuth must be provided')
+            raise Exception('pv_tilt and pv_azimuth must be provided using set_clearsky()')
```
kanderso-nrel, line 197: I'm not sure what exception type makes the most sense to raise here (and in `calc_clearsky_tamb` and `clearsky_preprocess`), but ideally we wouldn't be raising bare `Exception`s.  `ValueError` is a little awkward because it's raised in a different function from where the relevant argument is passed.  Maybe `AttributeError`?  Not sure.

docs/RdAnalysis_example.ipynb
```diff
@@ -242,15 +250,15 @@
...
    "metadata": {},
    "outputs": [
     {
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "0.9762135050486365\n",
-      "[0.97231317 0.97940615]\n"
+      "0.9761360319007901\n",
+      "[0.97242596 0.97941629]\n"
```
kanderso-nrel, line 261: Just checking, do we expect these (small) differences?

kanderso-nrel (commented):
Edit: GH saved my previous comment text somehow.  Real comments below:

rdtools/analysis.py
cdeline: Maybe ValueError would be ok.  I'm looking through and not finding any great suggestions in the list of errors.  Here's description of ValueError, which I guess applies here :  
>exception ValueError
Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as IndexError.

cdeline: NameError would also work here - we're expecting a named parameter pv_tilt and pv_azimuth that don't exist yet..

docs/RdAnalysis_example.ipynb
cdeline: I think this was actually related to a small change made to the soiling module back in Master that is finally propagating through to the example notebook.  it shouldn't have been caused by anything in RdAnalysis which just calls the functions.

cdeline:
OK, I think I've addressed all of Kevin's comments at this point.

PR #237, opened 2020-11-16
Remove max_timedelta parameters from step 0

mdeceglie:
Minor update to the old DKSC example, removing an optional parameter in favor of default behavior.

- [x] Updated changelog


mdeceglie:
Only small differences in the output, here are some of the highlights
Screen Shot 2020-11-16 at 2 35 16 PM
Screen Shot 2020-11-16 at 2 35 23 PM
Screen Shot 2020-11-16 at 2 35 28 PM

kanderso-nrel (approved):
Looks fine to me.  If this is going into a `2.0.2` sometime soon, might be good to think about anything else we want to include there too.  I don't have anything in mind but maybe @cdeline does

cdeline (approved):
Could this target master instead of development? Seems like a small change that can be rolled out immediately.  I'm on board with rapid roll-out and frequent minor patch releases.  Development is getting loaded with larger changes that will block these patches from seeing the light of day for a while.

mdeceglie:
@cdeline  I agree that straightforward patches like this one should get included in master and a patch release quickly (then update development from master). At the moment master and development are aligned, so I think it can go either way on this one (as long as we follow up with an update to master). 

Agreed that once we merge something like #117 to development, we don't want to have that block straightforward patches. In those cases, we can plan to PR straight to master as you suggest.

I'll merge this to development, then to master and release 2.0.2. Thanks @cdeline @kanderso-nrel !

PR #240, opened 2020-11-19
Display example notebooks in a gallery on RTD

kanderso-nrel:
- [x] Updated changelog

@mdeceglie and I briefly discussed offline how we should scale the documentation as we add new example notebooks.  We already have two, and #117 will add another.  Future new features will probably add more.  

I originally tried to set up [sphinx-nbexamples](https://sphinx-nbexamples.readthedocs.io/en/latest/index.html) but had significant trouble getting everything working with the rdtools repo layout.  Because of that, the fact that `sphinx-nbexamples` doesn't seem to get much maintenance attention, and my personal preference for the rendered HTML produced by `nbsphinx` over what `sphinx-nbexamples` produced, I gave up on it.

This PR uses a newer feature of `nbsphinx` that can generate galleries similar to the ones from `sphinx-gallery` but using jupyter notebooks as the source instead of python scripts.  The way I've used it here is to have a simple top-level `Examples` page that shows a thumbnail and links to each example (only two for now).  Each notebook gets its own page, but is listed under the `Examples` entry in the sidebar to keep it tidy.  https://rdtools--240.org.readthedocs.build/en/240/examples.html


mdeceglie (approved):
Looks good to me. @kanderso-nrel  Please update the wiki with instructions for adding new notebooks. What is written in example.rst is a great start, but depends on our future selves finding it, the wiki is a safer bet.

kanderso-nrel:
Dear readers of the future, here is the wiki page: https://github.com/NREL/rdtools/wiki/Example-gallery-notes

PR #248, opened 2020-12-11
Add project urls to setup.py for pypi page

kanderso-nrel:
- ~Updated changelog~

PyPI can be configured to display various URLs through `setup.py` ([docs](https://packaging.python.org/guides/distributing-packages-using-setuptools/#project-urls)).  Here's an [example](https://pypi.org/project/pandas/):

image

I'm not sure where the logic for matching icons to text (the dict keys) is documented, if anywhere, so I'm just copying pandas's descriptions.  One thing to note:  apparently the display order on pypi.org is reversed from the dictionary order here (see https://github.com/pypa/warehouse/issues/3097).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,6 +88,12 @@ KEYWORDS = [
     'PV'
 ]
 
+PROJECT_URLS = {
+    "Bug Tracker": "https://github.com/NREL/rdtools/issues",
+    "Documentation": "https://rdtools.readthedocs.io/",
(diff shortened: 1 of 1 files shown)


kanderso-nrel:
Could try it out on `test.pypi.org` I suppose.  I don't think I have credentials set up for that but probably @mdeceglie does?

mdeceglie (approved):
LGTM. Let's just go for it on our next release and patch things if it breaks.

Issue #241, opened 2020-11-19
Add intersphinx links for pvlib docs where appropriate 


mdeceglie:
Here's one example from #117 

```suggestion
        Model parameters for pvlib.temperature.sapm_cell(). Used in calculating cell
        temperature from ambient. If string, must be a valid entry
        for sapm model in pvlib.temperature.TEMPERATURE_MODEL_PARAMETERS. If dict, must
        have keys 'a', 'b', 'deltaT'. See pvlib.temperature.sapm_cell() documentation
```

Might be good to set up an [intersphinx link](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) to pvlib's docs so that things like this can automatically link to pvlib's API reference.  Future PR.

_Originally posted by @kanderso-nrel in https://github.com/NREL/rdtools/pull/117#discussion_r526980946_

PR #258, opened 2021-01-31
add intersphinx so that type links work

mikofski:
[Sphinx intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) is not added by default, it must be enabled and configured. Added links for:
* python
* numpy
* pandas
* scipy

this will create links for types like `float`, `dict`, _etc_. and `pandas.DataFrame`, `numpy.ndarray`, _etc_.

-  ~Code changes are covered by tests~
-  ~API.rst is up to date, along with other sphinx docs pages~
- [x] Updated changelog


kanderso-nrel:
Thanks @mikofski!  Should we add pvlib as well?  #241

docs/sphinx/source/conf.py
```diff
@@ -86,6 +87,14 @@
 
 master_doc = 'index'
 
+# Example configuration for intersphinx: refer to the Python standard library.
+intersphinx_mapping = {
+    'python': ('https://docs.python.org/3.8/', None),
+    'pandas': ('https://pandas.pydata.org/pandas-docs/stable/', None),
+    'numpy': ('https://numpy.org/doc/stable/', None),
+    'scipy': ('https://docs.scipy.org/doc/scipy/reference/', None'),
```
mdeceglie, line 95:
```suggestion
    'scipy': ('https://docs.scipy.org/doc/scipy/reference/', None),
```
This should fix some of the failing tests

kanderso-nrel:
Looks like maybe intersphinx doesn't work with nbsphinx example galleries :(  I'm not seeing any links in the example notebooks, e.g.: https://rdtools--258.org.readthedocs.build/en/258/examples/degradation_and_soiling_example_pvdaq_4.html

docs/sphinx/source/conf.py
```diff
@@ -86,6 +87,15 @@
 
 master_doc = 'index'
 
+# Example configuration for intersphinx: refer to the Python standard library.
+intersphinx_mapping = {
+    'python': ('https://docs.python.org/3.8/', None),
+    'pandas': ('https://pandas.pydata.org/pandas-docs/stable/', None),
+    'numpy': ('https://numpy.org/doc/stable/', None),
+    'scipy': ('https://docs.scipy.org/doc/scipy/reference/', None),
+    'pvlib': ('https://pvlib-python.readthedocs.io/en/stable/', None),
```
kanderso-nrel, line 96: I think it's worth pausing and thinking about what versions make the most sense to link against -- `stable` seems like the right choice to me for packages, but what about python itself?  Link against the version we expect people to be using (I'm guessing 3.7 or 3.8?), or maybe the most recent stable version, so 3.9?

mikofski: Sorry it should just say "3" eg: https://docs.python.org/3/ not 3.8. Python 3 is stable and always points to the most recent version. But you're right, that's 3.9. Hmm, up to you all I guess

kanderso-nrel:
>But now we need to go through and change some references to full namespaces, like `pandas.DataFrame`, not `pd.DataFrame` which isn't in the object inventory for Pandas docs. Ditto for numpy, _etc_. Perhaps a separate PR?

I'm happy to go through and update docstrings so things link correctly, we won't make you do it :)   I'll update this PR with those edits.

kanderso-nrel:
Ok, I think I'm happy with this.  I'm aware of two issues I think we'll just have to live with:
1. Like I mentioned above, the example notebooks don't get links. I looked through the nbsphinx docs and source repo, but didn't find a way to get it working.
2. Return values don't get type links when the function returns more than one value:
- single return value, works: https://rdtools--258.org.readthedocs.build/en/258/generated/rdtools.filtering.clip_filter.html
- multiple return value, doesn't work: https://rdtools--258.org.readthedocs.build/en/258/generated/rdtools.degradation.degradation_classical_decomposition.html

I wondered if maybe it was because we're still using an older version of sphinx (1.8.5), but I did a quick test build with the most recent sphinx (3.4.3) with the same results.  I also found this, but it doesn't seem to make a difference for me: https://github.com/sphinx-doc/sphinx/issues/8290

kanderso-nrel (approved):

mdeceglie:
`availability.py` line 591: the type link for tuple doesn't seem to be working

mdeceglie:
Since we're planning to include this in an upcoming release, I will change the base branch to development. We'll have to update the docstrings in new functions that weren't in master.

mdeceglie:
I think this is looking close to ready. We do have several spots where we can take a number or a series. We've been using the pvlib convention of calling these "numeric". It would be nice to set up a type link explaining that as proposed here: https://github.com/pvlib/pvlib-python/issues/1229 I believe @kanderso-nrel is looking into this.

docs/sphinx/source/index.rst
```diff
@@ -221,6 +221,20 @@ The most frequently used functions are:
...
+
+Some RdTools function parameters can take one of several types.  For example,
+the ``albedo`` parameter of :py:class:`TrendAnalysis` can be a static value like
+``0.2`` or a time-varying ``pandas.Series``.  To indicate that a parameter can
+take one of several types, we document them using the type alises listed below:
+
+.. glossary ::
+
+   ``numeric``
+      scalar, ``numpy.array``, ``pandas.Series``. Typically int or float dtype.
```
kanderso-nrel, line 235: 1) I copied this from pvlib's definition as a starting place.  Not sure we want `numpy.array` here.
2) It's a little silly to have a `glossary` with only one entry (for now, anyway), but this is the most convenient way I know of to get a linkable entity like `:term:` to use in the alias target.

setup.py
```diff
@@ -49,12 +49,12 @@
...
 EXTRAS_REQUIRE = {
     'doc': [
-        'sphinx==1.8.5',
+        'sphinx==3.2',
         'nbsphinx==0.8.5',
         'nbsphinx-link==1.3.0',
         'pandas==0.23.0',
         'pvlib==0.7.1',
-        'sphinx_rtd_theme==0.4.3',
+        'sphinx_rtd_theme==0.5.0',
```
kanderso-nrel, line 57: Bumping `sphinx_rtd_theme` was necessary to have nice rendering with the newer sphinx version

mdeceglie:
Looks good, I'll go ahead and merge this. Thanks @mikofski and @kanderso-nrel !

mikofski:
Thanks folks! 🚀

PR #261, opened 2021-02-02
Dependency updates

mdeceglie:
Updates bleach (see https://github.com/advisories/GHSA-vv2x-vrpj-qqpq ) and matplotlib (to fix a testing failure) versions

- [x] Updated changelog


kanderso-nrel (approved):
[Last time](https://rdtools.readthedocs.io/en/stable/changelog.html#requirements) we made it clear that it's only a change in recommended/requirements.txt versions, not a bump of minimum versions; I think that would be a good idea here too.

Issue #236, opened 2020-11-13
Within the object oriented soiling and degradation analysis, keep track of individual filters


mdeceglie:
It may be useful to keep track of individual filter results to allow post analysis inspection https://github.com/NREL/rdtools/pull/117#discussion_r345307662

kanderso-nrel:
I still think this is a good idea and suggest dumping all the filters in dataframes like `self.sensor_filter_components`.  I think the same motivation applies to keeping track of the normalized series.

martin-springer:
Closed with #263

PR #263, opened 2021-02-23
Store filter components on TrendAnalysis

kanderso-nrel:
- [x] Updated changelog

See #236


rdtools/analysis_chains.py
```diff
@@ -373,7 +373,9 @@ def _filter(self, energy_normalized, case):
         -------
         None
         '''
-        bool_filter = True
+        filter_components = pd.DataFrame({
+            'default': pd.Series(True, index=energy_normalized.index)
+        })
```
mdeceglie, line 378: Do we need this default column? Could we just do:
`filter_components = pd.DataFrame(index=energy_normalized.index)`
This would save us dropping the default column below. On the other hand, what about a case where someone turns off all the filtering? In that case maybe we want the `default` column and should retain it below.

kanderso-nrel: Yeah I think I was anticipating the case where all filtering is disabled, although now that I'm looking at it I don't think it's actually needed in the degenerate no-filter case?  `pd.DataFrame(index=[1,2,3]).all(axis=1)` returns all `True` for me on pandas `0.25.1` and `1.0.3`.  I'll change to your suggestion.

mdeceglie:
The new implementation has a slight difference in the index of `TrendAnalysis.sensor_filter` I believe this is due to initializing the dataframe with `index = energy_normalized.index` and that we can reproduce the old behavior by building the dataframe up with a concatenation of all the filter series. Though we ultimately filter `energy_normalized` I think using the union of all the filter indexes will be more robust to edge cases we haven't thought up yet.

rdtools/test/analysis_chains_test.py
```diff
@@ -131,6 +131,39 @@ def test_sensor_analysis_ad_hoc_filter(sensor_parameters):
...
+def test_filter_components_no_filters(sensor_parameters):
+    rd_analysis = TrendAnalysis(**sensor_parameters, power_dc_rated=1.0)
+    rd_analysis.filter_params = {}  # disable all filters
+    rd_analysis.sensor_analysis(analyses=['yoy_degradation'])
+    expected = pd.Series(True, index=rd_analysis.pv_energy.index)
+    pd.testing.assert_series_equal(rd_analysis.sensor_filter, expected)
+    assert rd_analysis.sensor_filter_components.empty
+
+
+def test_filter_ad_hoc_warnings(sensor_parameters):
```
mdeceglie, line 151: For both of the cases in this test, I suggest checking that the final `rd_analysis.sensor_filter` is as expected. At least by counting Trues and Falses. If not an explicit comparison.

rdtools/test/analysis_chains_test.py
```diff
@@ -131,6 +131,39 @@ def test_sensor_analysis_ad_hoc_filter(sensor_parameters):
...
+    with pytest.warns(UserWarning, match='ad_hoc_filter index does not match index'):
+        rd_analysis.sensor_analysis(analyses=['yoy_degradation'])
+
+    # warning about NaNs
+    ad_hoc_filter = pd.Series(True, index=sensor_parameters['pv'])
+    ad_hoc_filter.iloc[10] = np.nan
+    rd_analysis.filter_params['ad_hoc_filter'] = ad_hoc_filter
+    with pytest.warns(UserWarning, match='ad_hoc_filter contains NaN values; setting to False'):
+        rd_analysis.sensor_analysis(analyses=['yoy_degradation'])
+
```
mdeceglie, line 165: To be thorough, we should probably test some of this on the clearsky leg as well.

mdeceglie (commented):
This is looking close. I have a few suggestions to consider before we merge.

rdtools/test/analysis_chains_test.py
```diff
@@ -131,6 +131,39 @@ def test_sensor_analysis_ad_hoc_filter(sensor_parameters):
...
 
+def test_filter_components(sensor_parameters):
+    poa = sensor_parameters['poa_global']
+    poa_filter = (poa > 200) & (poa < 1200)
+    rd_analysis = TrendAnalysis(**sensor_parameters, power_dc_rated=1.0)
+    rd_analysis.sensor_analysis(analyses=['yoy_degradation'])
+    assert (poa_filter == rd_analysis.sensor_filter_components['poa_filter']).all()
+
+
+def test_filter_components_no_filters(sensor_parameters):
```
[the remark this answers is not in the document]
mdeceglie: Sorry! I missed the `assert_series_equal` part of this!

rdtools/test/analysis_chains_test.py
kanderso-nrel: I've added checks that the ad_hoc_filter is processed as described in the warning messages but didn't test the final `sensor_filter` value as I think other tests verify that the `ad_hoc_filter` is incorporated correctly into the final filter.  What do you think?

mdeceglie: That works, thanks!

mdeceglie:
Looks good. Thanks @kanderso-nrel!

PR #270, opened 2021-04-06
Add automated notebook check to CI

kanderso-nrel:
- [x] Updated changelog

Trying out using [nbval](https://nbval.readthedocs.io/en/latest/) with github actions to rerun notebooks and check for differences in the fresh output with the output stored in the notebook.  It is limited to checking text outputs, which means non-text outputs like plot images are NOT checked.  One goal is to make it easier to review PRs like #267 and #268.

Example passing run: https://github.com/NREL/rdtools/runs/2280180383#step:5:9
Example failing run: https://github.com/NREL/rdtools/runs/2279973037#step:5:9

Note that the `nbdime` section in the docs doesn't currently work with newer versions of pytest but will work if https://github.com/computationalmodelling/nbval/pull/166 is merged.

Other options:
- https://github.com/chrisjsewell/pytest-notebook
- https://github.com/nteract/testbook


kanderso-nrel:
The new plotly figures are a huge amount of text and dramatically increased the runtime of the checks (>1 hr).  I don't see any real benefit to checking that text anyway, so I've annotated those cells with `NBVAL_IGNORE_OUTPUT`.  Now the notebook checks run in a few minutes.

The failures are due to tiny numerical differences, for example `-0.5089691988109405` vs `-0.5089691988110401`.  I *think*, because this is looking at the stdout resulting from a `print` call, these are treated as strings (not floats) and there's not really a way to specify a numerical tolerance or anything.  @mdeceglie what do you think about printing those values with fewer digits?  For example changing `print(yoy_rd)` to `print("{:0.04f}".format(yoy_rd))`?

Aside from that I think this PR is ready to go.

kanderso-nrel:
After sprinkling a few calls to `np.round` in the notebooks, all green now (except for the spurious lingering 3.6 jobs).

On the most recent RdTools call it was suggested to save the newly-executed notebook as an artifact for manual inspection.  As far as I can tell this is unfortunately not currently possible with nbval.  There is an open feature request for this (https://github.com/computationalmodelling/nbval/issues/112).  As described in that issue, a workaround is to run `nbconvert` as a separate step.  This would execute the notebooks a second time and save those results as CI artifacts.  It's a shame to have to execute the notebooks twice, but that seems like the most reasonable way to do it at the moment.  I would not object to doing that, but I'm not convinced it's worthwhile.  Thoughts?

mdeceglie:
I think it might be worthwhile to try nbconvert. If we can't inspect the notebooks in full I think we still need to go through the whole manual integration test with a fresh environment etc (otherwise we never would have caught that recent problem with the the plots in #302). Can we give it a shot?

mdeceglie:
This looks good to go to me. Thanks @kanderso-nrel. Before merging please change the changelog entry to "pending" to avoid confusion since we don't have a real release in the works yet.

mdeceglie (approved):

PR #273, opened 2021-04-23
Test suite warning clean-up

kanderso-nrel:
- [ ] Updated changelog

Cleans up pytest output by (1) suppressing expected warnings like the experimental module warnings and (2) solving whatever the warning was about for unexpected warnings.

Note that our minimum requirements tests are always going to dump a bunch of deprecation warnings, but at least the others are cleaned up a bit.

### DIFF
--- a/rdtools/test/soiling_test.py
+++ b/rdtools/test/soiling_test.py
@@ -115,7 +115,7 @@ def test_soiling_srr_dayscale(soiling_normalized_daily, soiling_insolation):
     with pytest.raises(NoValidIntervalError):
         np.random.seed(1977)
         sr, sr_ci, soiling_info = soiling_srr(soiling_normalized_daily, soiling_insolation,
-                                              confidence_level=68.2, reps=10, day_scale=90)
+                                              confidence_level=68.2, reps=10, day_scale=91)
 
 
 def test_soiling_srr_clean_threshold(soiling_normalized_daily, soiling_insolation):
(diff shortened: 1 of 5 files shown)


rdtools/test/soiling_test.py
```diff
@@ -50,8 +50,8 @@ def test_soiling_srr(soiling_normalized_daily, soiling_insolation, soiling_times
     expected_means = expected_means[['soiling_rate', 'soiling_rate_low', 'soiling_rate_high',
                                      'inferred_start_loss', 'inferred_end_loss',
                                      'length', 'valid']]
-    pd.testing.assert_series_equal(expected_means, soiling_info['soiling_interval_summary'].mean(),
-                                   check_exact=False, check_less_precise=6)
+    actual_means = soiling_info['soiling_interval_summary'][expected_means.index].mean()
+    pd.testing.assert_series_equal(expected_means, actual_means, check_exact=False)
```
mdeceglie, line 54: @kanderso-nrel Can you say a little more about this change?

kanderso-nrel: `soiling_info['soiling_interval_summary']` has some datetime columns that aren't present in `expected_means`.  Older versions of pandas silently dropped those columns when taking the mean, so it ended up working out that there wasn't any mismatch when calling `assert_series_equal`.  However new versions of pandas warn that datetime columns won't get dropped in a future version, so this change just explicitly drops them prior to taking the mean.  This serves both to silence the warning and to be future-proof for whenever pandas makes that change.  Here's an example of that warning on another PR:  https://github.com/NREL/rdtools/runs/2549105826?check_suite_focus=true#step:5:342

mdeceglie: That makes sense, thanks!

mdeceglie (approved):
Looks good, thanks!

mdeceglie:
Was it intentional not to do the change log? I'm fine with that, just checking.

kanderso-nrel:
> Was it intentional not to do the change log? I'm fine with that, just checking.

Intentional; I figured it would just be clutter that nobody would be looking for.  Left it unchecked in case anyone else felt differently.

Issue #120, opened 2019-08-16
clearsky_temperature error when certain lat/lon values passed


cdeline:
There's an uncaught exception in clearsky_temperature.py when certain lat/lon values are passed.  For instance,  
>get_clearsky_tamb(df.index[0:1],35,31)

returns the correct values, but 
>get_clearsky_tamb(times[0:1],35,30)

returns "ValueError: Must have equal len keys and value when setting with an iterable"

The actual issue appears to be when someone inserts an incorrect lat/lon pair that points to ocean instead of land, there's no clearsky temperature to be returned, but the error is not at all descriptive.  This should be an exception that is caught and returned with a message that is somewhat helpful like 'incorrect lat/lon pair - no land found at this point'  or something like that...

PR #274, opened 2021-05-04
Issue #120 Revise NaN pixel handling in clearsky_temperature.py

cdeline:
Bug fix - previous code improperly handled a pixel with NaN average clearsky temperature.  Now just returns the NaN value with a userWarning that the clearsky temperature includes NaN's, and it's possible an invalid lat/lon pair has been passed.

~~- [ ] New functions added to `__init__.py`~~
~~- [ ] API.rst is up to date, along with other sphinx docs pages~~
- [x] Updated changelog


cdeline:
I'm not sure if we need to include additional tests for this case - NaN values are returned to the function with a warning.

rdtools/clearsky_temperature.py
```diff
@@ -85,11 +85,13 @@ def get_clearsky_tamb(times, latitude, longitude, window_size=40,
         day = _get_pixel_value(a, lon_index, lat_index, k, radius)
         night = _get_pixel_value(b, lon_index, lat_index, k, radius)
 
+        # Following code results in errors, likely incorrect
+        '''
```
kanderso-nrel, line 89: Should we just take this out, or is there value in preserving the history explicitly in code instead of hiding it away in the git history?

cdeline: I'm fine with removing it.

cdeline (commented):
I like the changes to the tests here, more consistent with our other testing methods.

kanderso-nrel (approved):
LGTM

PR #278, opened 2021-06-07
Adjust trend analysis interpolation behavior

mdeceglie:
- ~[ ] Updated changelog~


mdeceglie:
`TrendAnalysis` hasn't been part of a non-beta release, so I'm leaning toward skipping a separate change log entry for this, but I could be talked into including it.

rdtools/test/analysis_chains_test.py
```diff
@@ -78,6 +78,48 @@ def sensor_analysis_exp_power(sensor_parameters):
     return rd_analysis
 
 
+def test_interpoloation(basic_parameters, degradation_trend):
```
kanderso-nrel, line 81:
```suggestion
def test_interpolation(basic_parameters, degradation_trend):
```

kanderso-nrel (commented):
This seems like a good approach to me.  Right now the test only covers the interpolations in the constructor, so should we do the same index checks for the series involved in the `set_clearsky` method as well?

kanderso-nrel (approved):
LGTM

Issue #243, opened 2020-11-20
Improper clearsky interval labeling


mdeceglie:
**Describe the bug**
Clear sky irradiance and sun position are calculated instantaneously but treated as right-labeled averages in the example notebooks and `TrendAnalysis` of #117 

**Expected behavior**
The most robust behavior would be to calculate at high frequency and aggregate.

wholmgren:
Relevant thread on the pvlib python Google group: https://groups.google.com/d/msgid/pvlib-python/94230269-3421-4d50-a49c-79e46c3c815fn%40googlegroups.com

martin-springer:
Closed with #281

PR #281, opened 2021-06-17
Use interval averages for clear-sky irradiance modeling

kanderso-nrel:
- ~Updated changelog~ (similar to #278, not sure this needs an entry)

See #243.  Instead of calculating solar position and transposition for `self.poa_global.index`, which produces instantaneous clear-sky irradiance, this calculates solpos+transposition at 1-minute scale and then aggregates to `self.poa_global.index` to produces interval averages.  Check the difference in CSI stability for this example day in the PVDAQ4 notebook (modified to resample the data to hourly averages prior to analysis):

image

image


kanderso-nrel:
I reran the PVDAQ4 TrendAnalysis notebook and the results did not change (as expected -- the PVDAQ4 data is 1-minute already).  The DKASC notebook's clear-sky Rd changed from +0.01 to -0.02 (changes included in this PR).  I'm not wholly convinced that the DKASC data are actually right-labeled though, need to verify that somehow.

rdtools/analysis_chains.py
```diff
@@ -204,7 +204,9 @@ def _calc_clearsky_poa(self, times=None, rescale=True, **kwargs):
         Parameters
         ----------
         times : pandas.DateTimeIndex
-            times on for which to calculate clearsky poa
+            times on for which to calculate clearsky poa.  If not provided then
+            it will be simulated at 1-minute frequency and averaged to match the
+            index of self.poa_global
```
kanderso-nrel, line 209: Something to think about:  back in `2.0.0-alpha.0` this was a public method, but it's a private method now and the `times` and `rescale` parameters never get used anywhere, so consider removing them

rdtools/analysis_chains.py
```diff
@@ -242,7 +245,17 @@ def _calc_clearsky_poa(self, times=None, rescale=True, **kwargs):
             **kwargs)
         clearsky_poa = clearsky_poa['poa_global']
 
+        if aggregate:
+            interval_id = pd.Series(range(len(self.poa_global)), index=self.poa_global.index)
+            interval_id = interval_id.reindex(times).bfill()
```
mdeceglie, line 250: I think this would be more robust
```suggestion
            interval_id = interval_id.reindex(times.union(self.poa_global.index).bfill()
```

Here is an example that breaks the existing code:
```
original_index = pd.to_datetime([
    '2021/01/01 12:00',
    '2021/01/01 12:15:01',
    '2021/01/01 12:30:01',
    '2021/01/01 12:45:01'
])
times = pd.date_range(original_index[0], original_index[-1], freq="1min")

x = pd.Series(1, times)

interval_id = pd.Series(range(len(original_index)), index = original_index)
interval_id = interval_id.reindex(times).bfill()

out = x.groupby(interval_id).mean()
out.index = original_index
```

rdtools/analysis_chains.py
```diff
@@ -242,7 +245,17 @@ def _calc_clearsky_poa(self, times=None, rescale=True, **kwargs):
...
+            clearsky_poa = clearsky_poa.groupby(interval_id).mean()
+            clearsky_poa.index = self.poa_global.index
```
mdeceglie, line 252: I think we should set the first clearsky value to `NaN` to be strictly correct for right-labeled interval average

mdeceglie (commented):
Thanks @kanderso-nrel , here are a couple of points to consider

rdtools/analysis_chains.py
kanderso-nrel: Good counterexample.  What about `interval_id.reindex(times, method='backfill')` instead of `interval_id.reindex(times.union(self.poa_global.index).bfill()`?

kanderso-nrel (commented):
It's no surprise that additional solar position calcs are slow, but I wasn't expecting this much of a slowdown.  `pytest rdtools\test\analysis_chains_test.py --durations=0` reports these timing comparisons:

**development (numpy)**|**PR (numpy)**|**PR (numba)**|**PR (ephemeris)**|**test function**
:-----:|:-----:|:-----:|:-----:|:-----:
2.04s|58.16s|24.99s|7.41s|test\_clearsky\_analysis\_exp\_power
1.64s|55.92s|26.04s|7.00s|test\_clearsky\_analysis
1.58s|54.83s|25.03s|6.91s|test\_clearsky\_analysis\_optional
1.55s|53.25s|26.38s|6.89s|test\_errors
1.58s|52.63s|27.35s|7.02s|test\_plot\_cs
0.80s|31.57s|25.99s|3.45s|test\_no\_set\_clearsky

mdeceglie:
Is it possible to use the numba implementation here? Would there be a downside?

rdtools/analysis_chains.py
mdeceglie: I like it!

kanderso-nrel:
> Is it possible to use the numba implementation here? Would there be a downside?

Yes possible, I added timings to the table above.  No big downside from my perspective (I use numba for our internal RdTools runs).  Using numba on non-mainstream CPUs like single-board computers or new CPU architectures (e.g. Apple Silicon), but I don't think that's likely to be an issue for us. 

But even with numba, switching to 1min->average is still a significant slowdown.  Not sure there's any way around it unless we think the half-interval shift method is an acceptable substitute.

wholmgren:
This PR caught my attention since I've worried about similar problems in `solarforecastarbiter-core` and considered a few options in pvlib.

Is the bottleneck the solar position calculation or the aggregation? And if the aggregation, is it because you don't want to assume regular intervals to resample to and so you are using `groupby`?

kanderso-nrel:
I think the solar position calculation is the bottleneck, something like 90+% of the runtime for getting clear-sky poa_global.  The `groupby` aggregation is indeed to handle irregular intervals and takes something <1% of the runtime.

Outside of this PR I've been wondering if it would make sense to move away from SPA to a faster, less accurate algorithm -- I think the uncertainties in the clear-sky irradiance (turbidities etc) must dwarf the solar position error, so maybe there's little point in wasting CPU cycles on SPA for simple clear-sky simulations.  I want to do a runtime/accuracy comparison of various solpos algorithms but not sure when I'll be able to get to it.  Maybe for the next PVPMC workshop, maybe not.

mdeceglie:
I suggest we add an argument to `TrendAnalysis.set_clearsky()` that takes a method for `loc.get_solarposition()` Then people can try speed things up if they want. Thoughts?

kanderso-nrel:
> I suggest we add an argument to `TrendAnalysis.set_clearsky()` that takes a method for `loc.get_solarposition()` Then people can try speed things up if they want. Thoughts?

Not opposed, but note that the current set of alternatives to SPA isn't very appealing: the accuracy of `ephemeris` "is not guaranteed" and `pyephem`'s author doesn't recommend it anymore either.  I'll add a keyword with the hope that more options will be available in the future :)

cwhanse:
> Not opposed, but note that the current set of alternatives to SPA isn't very appealing: the accuracy of `ephemeris` "is not guaranteed" and `pyephem`'s author doesn't recommend it anymore either. I'll add a keyword with the hope that more options will be available in the future :)

For the purposes here, I doubt that any differences among the solar position calculators are meaningful. Someone in the past did a comparison which I haven't been able to locate; what I recall was that the differences were fractions of a degree. That warning in `ephemeris` is warranted because there never was a published validation, but it would be very surprising if it's inaccuracy was meaningful.

rdtools/analysis_chains.py
```diff
@@ -167,7 +168,8 @@ def set_clearsky(self, pvlib_location=None, pv_azimuth=None, pv_tilt=None,
         albedo : numeric
             Albedo to be used in irradiance transposition calculations. Can be right-labeled
             Pandas Time Series or single numeric value.
-
+        solar_position_method : str, optional
+            Optional method name to pass to :py:func:`pvlib.solarposition.get_solarposition`
```
mdeceglie, line 172: I suggest adding "switching methods may improve calculation times"

rdtools/analysis_chains.py
```diff
@@ -138,7 +138,8 @@ def __init__(self, pv, poa_global=None, temperature_cell=None, temperature_ambie
 
     def set_clearsky(self, pvlib_location=None, pv_azimuth=None, pv_tilt=None,
                      poa_global_clearsky=None, temperature_cell_clearsky=None,
-                     temperature_ambient_clearsky=None, albedo=0.25):
+                     temperature_ambient_clearsky=None, albedo=0.25,
+                     solar_position_method=None):
```
mdeceglie, line 142: Let's be explicit about the default behavior, otherwise we rely on the default behavior of pvlib, which is in version 0, so could potentially change.

mdeceglie (commented):
Looks good. Let's add some testing to make sure the method gets passed down

rdtools/analysis_chains.py
kanderso-nrel: Is the concern that the default algorithm might change and affect the analysis results?  My thought was that relying on the pvlib default is actually preferable; I think the default method name (e.g. `nrel_numpy`) is more likely to change than the default algorithm is, if I ever dust off https://github.com/pvlib/pvlib-python/pull/1098 anyway.

mdeceglie: Yep, the concern is silently changing the calculations under the hood.

mdeceglie (approved):
Looks good. Please go ahead and merge when you're ready. Thanks!

wholmgren:
Some comparisons here: https://github.com/pvlib/pvlib-python/blob/master/docs/tutorials/solarposition.ipynb

PR #282, opened 2021-06-30
Plotting function allows data gap visualization

cdeline:
New plotting function to visualize data gaps and contribution of time series points to the Rd distribution better.  Use `plotting.degradation_summary_plots(detailed=True)`

New returns to `degradation.degradation_year_on_year:  calc_info{'usage_of_points'}`.  (time-series with number of times each datapoint enters into the Rd distribution calculation.

- [X] Code changes are covered by tests
- ~New functions added to `__init__.py` (N/A)~
- [x] Updated changelog


cdeline:
Some example plots from the pvdaq4 example notebook with `detailed=True` option. Green points: only 1 vote in distribution. Red points: 0 votes in distribution. Default color: 2 votes in distribution.
image
image

rdtools/plotting.py
```diff
@@ -92,8 +98,14 @@ def degradation_summary_plots(yoy_rd, yoy_ci, yoy_info, normalized_yield,
     ax2.set_xlabel('Annual degradation (%)')
 
     renormalized_yield = normalized_yield / yoy_info['renormalizing_factor']
-    ax1.plot(renormalized_yield.index, renormalized_yield, 'o',
-             color=plot_color, alpha=scatter_alpha)
+    if detailed:
+        def f(x): return {0: 'red', 1: 'green', 2: plot_color}[x]
+        colors = yoy_info['usage_of_points'].apply(f)
```
kanderso-nrel, line 103: I think using a wrapper function with `apply` can be replaced with `yoy['usage_of_points'].map({0: 'red', 1: 'green', 2: plot_color})`.

kanderso-nrel (commented):
Neat!  It seems to behave the way I expect it to.  The clear-sky plot confused me at first because I thought the orange points were red and weren't contributing to the distribution but I get it now.  

Should we also exercise the new plotting parameter in the `test_degradation_summary_plots_kwargs` function in `plotting_test.py`?

rdtools/plotting.py
cdeline: Nice!  I figured there had to be a simpler way, but couldn't figure it out.

cdeline:
Hmm.  trying to make the html sphinx docs on this branch, I get the following error:
>Running Sphinx v3.2.1
>Extension error:
>Could not import extension sphinx_gallery.load_style (exception: No module named 'sphinx_gallery')
I wonder if sphinx_gallery now a dependence for making our doc pages - if it's just my machine then no worries - it appears that readthedocs pages are generating ok.

kanderso-nrel:
Yep, as of #240 `sphinx_gallery` is a `doc` requirement: https://github.com/NREL/rdtools/blob/master/setup.py#L50-L61

kanderso-nrel (approved):
LGTM

PR #284, opened 2021-07-13
Fix pandas 1.3.0 error and deprecations

kanderso-nrel:
- [x] Updated changelog

This PR addresses a failure in the clear-sky tamb function with pandas 1.3.0 (noticed in the eager-upgrade CI job on #281) and suppresses a bunch of deprecation warnings for casting datetime indexes to int dtype:
- `AttributeError: 'numpy.float32' object has no attribute 'is_integer'`
- `FutureWarning: casting timedelta64[ns] values to int64 with .astype(...) is deprecated and will raise in a future version. Use .view(...) instead.`

Note that the remaining `astype/view` warnings are from pvlib and have already been addressed on that end: https://github.com/pvlib/pvlib-python/pull/1256


cdeline:
I've tested this out and it seems to work, although I couldn't explain exactly how or why it works.  I've seen other projects taking this approach to move from `.astype` to `.view` because it is suggested by the deprecation warning.  I will say that using the `.view` function seems very kludgy and un-Pythonic, and I wonder if there isn't a different way to explicitly convert from datetime to epoch nanoseconds.  All of the other changes you have here are just fine though.

mdeceglie (approved):
Thought the pandas changes necessitating this are still somewhat mysterious to me, these changes all look good. Thanks @kanderso-nrel

PR #290, opened 2021-09-02
Support pvlib 0.9.x

kanderso-nrel:
- [x] Updated changelog

Extra to-do:
- [x] Add note to SAPM functions saying an older pvlib is needed

`pvlib==0.9.0` is up on pypi as of today.  Our current setup.py specifies `'pvlib >= 0.7.0, <0.9.0'`, but it'd be nice to support the `0.9.x` series if it's easy.  I think there are only two relevant changes (but we'll see what the CI has to say):
- The SAPM normalization functionality won't work on 0.9.0 because `LocalizedPVSystem` is removed
- For clear-sky modeling, the `tables` dependency is removed and `h5py` is added in its place.  Note that `h5py` is required for pvlib 0.9.0 whereas `tables` was optional for previous versions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,8 +55,6 @@ EXTRAS_REQUIRE = {
         'sphinx==3.2',
         'nbsphinx==0.8.5',
         'nbsphinx-link==1.3.0',
-        'pandas==0.23.0',
-        'pvlib==0.7.1',
         'sphinx_rtd_theme==0.5.2',
         'ipython',
         # sphinx-gallery used indirectly for nbsphinx thumbnail galleries; see:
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -56,8 +55,6 @@
         'sphinx==3.2',
         'nbsphinx==0.8.5',
         'nbsphinx-link==1.3.0',
-        'pandas==0.23.0',
-        'pvlib==0.7.1',
```
mdeceglie, line 60: What was the reasoning for removing these? Redundant with requirements.txt?

mdeceglie (commented):
Looks good to me

setup.py
kanderso-nrel: `[doc]` is the requirements to build the sphinx docs, which doesn't actually require any rdtools functionality (e.g. `make html` does not run the notebooks or tests or anything).  So no need to pin specific versions since the packages don't get used for building the docs anyway.  And if we're not pinning versions then they're redundant with `INSTALL_REQUIRES`.

mdeceglie (approved):

PR #297, opened 2021-09-28
Create github action to check consistency of requirements.txt and notebook_requirements.txt

kanderso-nrel:
- [x] Updated changelog

@mdeceglie suggested adding a job to verify the consistency of `requirements.txt` and `notebook_requirements.txt` to make it easier to review PRs from dependabot.  Note that consistency of `requirements.txt` is already checked by the pytest jobs, but currently no CI job uses `notebook_requirements.txt`.  Also note that dependabot does suggest updates to `notebook_requirements.txt`, not just `requirements.txt`, e.g. #288.  Maybe this job will become redundant if we ever get around to #270. 

I tested some inconsistent environments on my fork to see what failures look like; here are the corresponding jobs:
- inconsistent 1: https://github.com/kanderso-nrel/rdtools/runs/3737515003?check_suite_focus=true
- inconsistent 2: https://github.com/kanderso-nrel/rdtools/runs/3737560421?check_suite_focus=true
- consistent: https://github.com/kanderso-nrel/rdtools/runs/3737621395?check_suite_focus=true

Note that `inconsistent 1` requested a version of numpy without wheels for py 3.7, so it tried to compile it from source and it was that compilation that failed, not a consistency error, so it's not really relevant.  But `inconsistent 2` shows an actual consistency failure.

### DIFF
--- a/.github/workflows/requirements.yaml
+++ b/.github/workflows/requirements.yaml
@@ -1,0 +1,22 @@
+name: requirements
+
+on: [pull_request, push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
+    - name: Install ${{ matrix.env }}
+      run: |
+        python -m pip install --upgrade pip wheel
+        pip install -r requirements.txt -r docs/notebook_requirements.txt
+    - name: Import RdTools 
+      run: |
+        python -c "import rdtools"
--- a/.github/workflows/requirements.yaml
+++ b/.github/workflows/requirements.yaml
@@ -13,7 +13,7 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: 3.7
-    - name: Install ${{ matrix.env }}
+    - name: Install notebook environment
       run: |
         python -m pip install --upgrade pip wheel
         pip install -r requirements.txt -r docs/notebook_requirements.txt
--- a/.github/workflows/requirements.yaml
+++ b/.github/workflows/requirements.yaml
@@ -5,7 +5,11 @@ on: [pull_request, push]
 jobs:
   requirements:
 
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+
+    runs-on: ${{ matrix.os }}
 
     steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/requirements.yaml
```diff
@@ -0,0 +1,22 @@
+name: requirements
+
+on: [pull_request, push]
+
+jobs:
+  requirements:
+
+    runs-on: ubuntu-latest
```
kanderso-nrel, line 8: Should we test other OSs?  I guess it depends on whether we are checking that the environment can be installed easily, or checking whether the package versions are compatible (which is a weaker requirement than the package versions can be installed easily on all OSs).

.github/workflows/requirements.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
```
kanderso-nrel, line 15: Same question as above, but for different python versions.  I don't think we claim that the notebooks support anything other than 3.7

.github/workflows/requirements.yaml
mdeceglie: Although I'm a little worried we will open a can of worms, it does sound useful to check on windows and osx. Can we give it a try?

.github/workflows/requirements.yaml
mdeceglie: I'm fine with sticking with just 3.7

.github/workflows/requirements.yaml
mdeceglie: Or maybe just add the other OSs to the pytest workflow?

mdeceglie:
This looks good, but i'm imagining cutting 2.1.1, so we should probably update the changelog about the xgboost requirement

kanderso-nrel:
Given this is just a maintenance PR, do we need to cut a release?  I imagined the xgboost requirement change to be a short-term change that gets reverted before we're ready to cut another release.  If that's the case then I don't see a point in a changelog entry.

mdeceglie:
I think we should do a patch, an updated `setup.py` has limited utility sitting on github, we should get it out in pypi

kanderso-nrel:
Oh right, if it's broken for us, it's broken for everyone.  Forgot about that part.  Changelog added!

mdeceglie (approved):
Looks good to me, thanks @kanderso-nrel !

Issue #299, opened 2021-11-22
Python 3.6 will reach end of life soon


kanderso-nrel:
I don't think we have a formal policy dictating what python versions we support.  I think we always want to support new python versions as they are released (probably with some delay to allow the base pydata packages to release wheels), but what about older python versions?  One non-subjective method is to drop versions when they reach end of life, which python 3.6 will reach in a few weeks: https://devguide.python.org/#status-of-python-branches

I think semver says changing dependency requirements only requires a patch release: https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api

It might make sense to drop python 3.6 at the same time that we add python 3.10, as it is proving difficult to find dependency versions that are available for both versions.  See some preliminary tests here: https://github.com/kanderso-nrel/rdtools/actions/runs/1491923588

PR #302, opened 2021-11-30
Drop python 3.6; add python 3.10; update requirements-min

kanderso-nrel:
- ~Code changes have been evaluated for compatibility/integration with TrendAnalysis~
- [x] Updated changelog

See #299.  scikit-learn is the only dependency without python 3.10 wheels, but the CI passes anyway because it manages to build it from source.  Still, I suggest waiting for them to release 3.10 wheels on [PyPI](https://pypi.org/project/scikit-learn/#files) so that we can use the new version (likely 1.0.2) in `requirements.txt`.  According to https://github.com/scikit-learn/scikit-learn/issues/21882 this might happen as soon as mid December.


rdtools/test/filtering_test.py
```diff
@@ -156,10 +157,12 @@ def test_logic_clip_filter(generate_power_time_series_no_clipping,
...
     # Test that a warning is thrown when the time series is tz-naive
     warnings.simplefilter("always")
-    with warnings.catch_warnings(record=True) as w:
+    with warnings.catch_warnings(record=True) as record:
         logic_clip_filter(power_nc_tz_naive)
         # Warning thrown for it being an experimental filter + tz-naive
-        assert len(w) == 2
+        assert_warnings(['The logic-based filter is an experimental',
+                         'Function expects timestamps in local time'],
+                        record)
```
kanderso-nrel, line 165: These warning changes aren't related to py 3.10 or 3.6.  It's just because testing the number of emitted warnings is rather brittle and failed on the new 3.10 test because of some unrelated warning from an indirect dependency.  Normally I'd used `pytest.warns` to check the specific warning message text, but I don't think it has a way to check for multiple warnings like we need to do here.  Thus the new function `assert_warnings` in `conftest.py`.

kanderso-nrel (commented):
@mdeceglie I think this is ready for you to take a look whenever is convenient

rdtools/plotting.py
```diff
@@ -395,18 +395,19 @@ def availability_summary_plots(power_system, power_subsystem, loss_total,
     ax3 = fig.add_subplot(gs[2, 0], sharex=ax1)
     ax4 = fig.add_subplot(gs[:, 1], sharex=ax1)
 
-    # inverter power
-    power_system.plot(ax=ax1)
-    ax1.set_ylabel('Inverter Power [kW]')
     # meter power
```
mdeceglie, line 398: At the risk of nitpicking, should this comment be "inverter power"?

kanderso-nrel: no, but the one a few lines down should be.  d'oh

mdeceglie (approved):

mdeceglie:
For some reason after I changed the base to `master` 3.6 tests are showing up as pending, but we aren't supposed to be testing on 3.6 anymore. I'll merge it and hope for the best.

PR #304, opened 2021-12-01
Converted dat file to json file to remove XGBoost versioning issues.

kperrynrel:
- ~Updated changelog~


kanderso-nrel:
Cool!  We should also change `xgboost >= 1.3.3, <1.5.0` back to `xgboost >= 1.3.3` in setup.py and make sure the CI still passes with the latest version of xgboost.  And I guess make a new v2.1.2 changelog file with a note about removing the version restriction. 

This PR is currently going to master (not development); just pointing that out in case it doesn't match @mdeceglie's release plan.

kperrynrel:
image
So I've got the model loading but I keep getting a weird MacOS error for the requirements.txt that I've never seen before-- @kanderso-nrel , @mdeceglie have you seen this before? Is there a way to get libomp into the MacOS environment?

kanderso-nrel:
@mdeceglie ran into that a while ago, check the thread in the rdtools teams channel from Aug 27.  Seems like he fixed it by installing from conda; maybe we should do the same for the OS X tests.

Those tests were passing before though, so maybe for some reason it's caused by the new versions you're specifying in setup.py and requirements.txt?  Are those changes necessary for some other reason or can we revert them?

kanderso-nrel:
Oh, it's the `requirements` job that fails, not the pytest one, and specifically it's the `import rdtools` step:

https://github.com/NREL/rdtools/blob/600991d047c04ef0dbf1b978b62abeeb7d84b362/.github/workflows/requirements.yaml#L24-L26

We could just remove that step -- it seemed like a good idea to include, but that step isn't actually necessary for the job to fulfill its purpose, so I'd say let's just remove it if it's causing problems.  @kperrynrel, want to delete those workflow lines and see what happens here?

We might have to do something with conda if we pursue #303, but we can cross that bridge when we come to it.

kperrynrel:
@kanderso-nrel removed the rdtools import and it's all passing. Let me know what you guys think!

setup.py
```diff
@@ -45,7 +45,7 @@
     'h5py >= 2.7.1',
     'plotly>=4.0.0',
     'joblib >= 0.16.0',
-    'xgboost >= 1.3.3, <1.5.0',
+    'xgboost >= 1.5.1',
```
kanderso-nrel, line 48: What's the reason for updating the minimum?  I thought the motivation for switching to JSON was so that we wouldn't have to worry about keeping track of xgboost versions.  1.3.3 seems to work for me locally with the JSON model, so maybe we don't actually need to change the version to 1.5.1 here and in the two requirements files?  So long as it's not a headache for us it's nice to not be overly restrictive on dependency version ranges, so keeping 1.3.3 would be preferable if it doesn't break anything.

kperrynrel:
@kanderso-nrel I just updated it in a previous push when I was trying to debug the failing rdtools import. I can drop back down again, it'll work

mdeceglie:
As an integration test, I installed this in a fresh environment and ran `TrendAnalysis_example_pvdaq4.ipynb` and I am getting different results in the "Modifying and inspecting the filters" section. Can others reproduce this? This varies from the [existing result](https://rdtools.readthedocs.io/en/stable/examples/TrendAnalysis_example_pvdaq4.html#Modifying-and-inspecting-the-filters) 
Screen Shot 2021-12-09 at 11 23 37 AM

kperrynrel:
@kanderso-nrel successfully removed joblib in new commit

kperrynrel:
@mdeceglie I did redownload the package in the new environment and was able to reproduce your results:
image
I had to regenerate the model as a JSON, and that involved retraining it with the same values. I didn't set a seed so the model would vary somewhat even with the same parameters + training data. I can retrain it again and see how much variance we get

kperrynrel:
Hey all, I just wanted to keep things simple, so I resaved the old model in the new JSON format and replicated the original results in the jupyter notebook:
image

mdeceglie (approved):
Looks good, thanks @kperrynrel!

PR #317, opened 2022-03-02
Update github actions triggers

### LINKED ISSUE #312: The number of CI jobs is a bit excessive
Between linting, pytest, environment consistency, and now notebook checks, RdTools PRs have 25-30 CI jobs nowadays.  However, a big chunk of that is because the pytest jobs are doubled up -- they run on both `pull_request` and `push` ([ref](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request)).  I think running `push` on PRs is unnecessary: it's running the tests on the latest commit of the PR branch, but really we're more interested in what would happen if the PR branch were merged into the target branch (which is what `pull_request` does).  However, `push` might still be useful outside of the context of a PR, for example if we pushed straight to `master` or something.  

I did a brief survey of what other packages do and found that a good number of them restrict the `push` to only apply to a select few important branches like main/master.  That means that `push` would not apply to a regular feature branch used to open a PR, and so the CI jobs on the PR would not be doubled up.  Here are the packages I found in a few minutes of searching: [astropy](https://github.com/astropy/astropy/blob/main/.github/workflows/ci_workflows.yml), [metpy](https://github.com/Unidata/MetPy/blob/main/.github/workflows/tests-pypi.yml), [geopandas](https://github.com/geopandas/geopandas/blob/main/.github/workflows/tests.yaml), [numpy](https://github.com/numpy/numpy/blob/main/.github/workflows/build_test.yml), [pandas](https://github.com/pandas-dev/pandas/blob/main/.github/workflows/code-checks.yml), [scipy](https://github.com/scipy/scipy/blob/main/.github/workflows/linux.yml).

I suggest we take that approach here, i.e. change this line in the pytest job config:

https://github.com/NREL/rdtools/blob/074909079bcb6c9caa6ca63a2e2719edce432b6c/.github/workflows/pytest.yaml#L3

to:

```yaml
on:
  push:
    branches:
    - master
    - development
  pull_request:
```

I think I'd probably support doing that for the flake8 and requirements jobs too.

kanderso-nrel:
- ~[ ] Code changes are covered by tests~
- [x] Updated changelog

Closes #312


setup.py
```diff
@@ -56,7 +56,7 @@
 EXTRAS_REQUIRE = {
     'doc': [
         'sphinx==3.2',
-        'nbsphinx==0.8.5',
+        'nbsphinx==0.8.8',
```
kanderso-nrel, line 59: This resolves an unrelated incompatibility between `jinja2` and `markupsafe` that was causing the docs build to fail.

docs/sphinx/source/changelog/v2.1.4.rst
```diff
@@ -0,0 +1,9 @@
+************************
+v2.1.4 (TBD)
+************************
+
+Requirements
+------------
+* Bump ``nbsphinx`` version from 0.8.5 to 0.8.8 in the optional
+  ``[doc]`` requirements (:pull:`317`)
+
```
mdeceglie, line 9: Should this go in `pending.rst`?

kanderso-nrel: d'oh, yes

PR #332, opened 2022-07-01
Fix unpacking in cell 9 and re-run

mdeceglie:

This PR re-runs the CODS notebook in the latest environment and fixes a bug in the notebook with the unpacking of variables in cell 9.


mdeceglie:
Note that the soiling and degradation values change slightly I think this is because we need to both increase the number of monte carlo realizations and seed the numpy random number generator.

PR #335, opened 2022-07-15
334 degradation timeseries

cdeline:
New function in TrendAnalysis object and examples in notebooks for time series of degradation rate to be plotted.

- [X] Code changes have been evaluated for compatibility/integration with TrendAnalysis
- [X] New functions added to `__init__.py`
- [X] API.rst is up to date, along with other sphinx docs pages
- [X] Example notebooks are rerun and differences in results scrutinized
- [X] Updated changelog


kanderso-nrel:
@cdeline what would you think about using `.rolling(resample_days, center=True).median()` instead of `.resample(f'{round(resample_days)}d').median()`?  It would show a lot more detail, which might be good or bad depending on point of view.  Looks like this for PVDAQ4 w/ soiling:

image


Something else to consider would be to move the plotting code to `rdtools.plotting` and have the TrendAnalysis method call it like how `TrendAnalysis.plot_degradation_summary` does.

kanderso-nrel:
My above suggestion is flawed in that it takes 365 consecutive slopes, which is not the same as slopes from 365 days if some days are missing.  Better to use a pandas offset like @cdeline did rather than integer number of days like I did.  `center=True` is also likely not desirable in this context.

Anyway following some discussion with @mdeceglie, here's another possible alternative.  It is still a moving window but does bootstrapping in each window to determine a moving CI around the moving median.  Here orange is the moving median, the blue region shows the moving 95% bootstrapped CI, and black shows the overall median that we would normally calculate:

image

cdeline:
Cool!  Did Mike have some example code on achieving the above figure?  That seems like a nice way to do it.

kanderso-nrel:
Here's the code I used, but there's certainly room for improvement:

```python
def _bootstrap(x, percentile, reps):
    # stolen from degradation_year_on_year
    n1 = len(x)
    xb1 = np.random.choice(x, (n1, reps), replace=True)
    mb1 = np.nanmedian(xb1, axis=0)
    return np.percentile(mb1, percentile)

resample_days = 365
roller = yoy_info['YoY_values'].resample('d').mean().rolling(resample_days, min_periods=resample_days//2)
# unfortunately it seems that you can't return multiple values in the rolling.apply() kernel.
# TODO: figure out some workaround to return both percentiles in a single pass
ci_lower = roller.apply(_bootstrap, kwargs={'percentile': 2.5, 'reps': 100}, raw=True)
ci_upper = roller.apply(_bootstrap, kwargs={'percentile': 97.5, 'reps': 100}, raw=True)

fig, ax = plt.subplots()
ax.fill_between(ci_lower.index, ci_lower, ci_upper)
ax.plot(roller.median(), c='tab:orange')
ax.axhline(yoy_info['YoY_values'].median(), c='k', ls='--')
plt.ylabel('Degradation trend (%/yr)')
```

rdtools/plotting.py
```diff
@@ -438,3 +438,67 @@ def availability_summary_plots(power_system, power_subsystem, loss_total,
...
+        mb1 = np.nanmedian(xb1, axis=0)
+        return np.percentile(mb1, percentile)
+
+    try:
+        results_values = yoy_info['YoY_values']
+
+    except KeyError:
+        raise KeyError("yoy_info input dictionary does not contain key `YoY_values`.")
+
+    roller = results_values.resample('d').mean().rolling(rolling_days, min_periods=rolling_days//2)
```
mdeceglie, line 489: I’m not sure about `.mean()` here what’s the reasoning?

kanderso-nrel: I used `.resample('d').mean()` as a hacky way of converting an index with gaps/jumps to a continuous daily index, filling NaN for the missing values.  The point was to be able to define the window width with an integer instead of an offset.  An explicit `reindex()` is probably clearer, or maybe just use an offset in the first place.

cdeline: I found the following code to result in equivalent results if you don't like the resample:

`roller = results_values.rolling(f'{rolling_days}d', min_periods=rolling_days//2)`

data is 1:1 comparable with fewer NaN's in the beginning:
image

cdeline: I'm not sure what's causing this nbval failure, it's occurring in a soiling calculation that I haven't changed at all...

kanderso-nrel:
> I'm not sure what's causing this nbval failure, it's occurring in a soiling calculation that I haven't changed at all...

@cdeline I think the difference is that the bootstrapping in `degradation_timeseries_plot()` uses numpy's RNG and leaves it in a different state afterwards when the soiling calculations are done.  Disabling that bootstrapping by setting `include_ci=False` resolves the difference in output for me locally.  Note that the TrendAnalysis example does the soiling calculations _before_ the plotting, so this isn't a problem in that notebook.

PR #337, opened 2022-07-29
Dependencies: increase numpy, notebook & send2trash versions

kanderso-nrel:
- ~[ ] Code changes are covered by tests~
- [X] Example notebooks are rerun and differences in results scrutinized (trusting the CI notebook checks on this one)
- ~[ ] Updated changelog~ leaving this for one big update before the release (https://github.com/NREL/rdtools/pull/326#discussion_r876962230)

Supersedes #328.

~Not addressing #329 because (1) it doesn't seem like there's any cause for concern about these numpy versions (https://github.com/pvlib/pvanalytics/issues/137) and (2) the suggested version of 1.22.0 isn't available for python 3.7, so we wouldn't be able to test on it without dropping python 3.7. That doesn't seem like a good decision at this point.~ Decided to bump numpy to 1.22.0 as suggested by dependabot, and just not use `requirements.txt` for python 3.7 anymore.


kanderso-nrel:
> I think we should/can update numpy in requirements.txt because we (should) only care about that with python 3.10.

Well we still use `requirements.txt` with other python versions, including python 3.7 where `numpy==1.22.0` isn't available.  Would you be okay with adding an [exclude](https://github.com/NREL/rdtools/blob/fc9f61ece9de75e79d98f65254762be2652d14e2/.github/workflows/pytest.yaml#L23) item for the 3.7+requirements.txt variant?  Or maybe we should go a step farther and only use `requirements.txt` with the python version we use for the notebooks?

mdeceglie:
I would be OK excluding 3.7+requirements. While it's tempting to exclude every version of python except that of the notebook environment, it would be good to have testing with every version of python we nominally support. I think the proposed situations is manageable, but I can see it sliding toward a situation where we have a separate requirements files for every version of python we support, which sounds less desirable, so that gives me pause.

kanderso-nrel:
We may have to wait a bit to address #338.  The latest non-beta release of `nbconvert` requires `mistune<2`.  We could use a beta release of `nbconvert` but I'm not wild about that.

mdeceglie (approved):
I think this is good to go. @kanderso-nrel would you give it a quick look since I made so many changes myself?

kanderso-nrel:
I made a small [script](https://github.com/NREL/rdtools/blob/dependency-check/.github/workflows/validate_requirements_versus_installed.py) to compare the list of installed packages with the requirements files to see if any other indirect dependencies snuck in.  Running it w/ the new requirements files in this PR, the only things "missing" are `'setuptools', 'pip', 'wheel'` for [linux](https://github.com/NREL/rdtools/runs/8158155703?check_suite_focus=true#step:5:14) and [macos](https://github.com/NREL/rdtools/runs/8158155887?check_suite_focus=true#step:5:13) and `'pip', 'pywinpty', 'pywin32', 'setuptools', 'wheel'` for [windows](https://github.com/NREL/rdtools/runs/8158156064?check_suite_focus=true#step:5:13), which both seem OK to me.  So LGTM!

PR #326, opened 2022-04-27
Update notebook environment to python 3.10

kanderso-nrel:
- ~[ ] Code changes are covered by tests~
- ~[ ] Code changes have been evaluated for compatibility/integration with TrendAnalysis~
- ~[ ] New functions added to `__init__.py`~
- ~[ ] API.rst is up to date, along with other sphinx docs pages~
- [x] Example notebooks are rerun and differences in results scrutinized
- [x] Updated changelog

See #319

PR #344, opened 2022-09-13
2.2.0-beta release

mdeceglie:
- [x] Updated changelog


kanderso-nrel:
Is it time to clean up/summarize the dependency updates as mentioned in https://github.com/NREL/rdtools/pull/326#discussion_r876962230?

#339 is also ready for review and could go in this release if we want.

mdeceglie:
Good point on the dependency updates. I'm toying with the idea of just capturing the details of changes to setup.py, for others perhaps we could just point to the relevant PRs and note that there are changes but not worry about every specific package. Another heuristic would be to only report the details of changes to direct dependencies. What do you think @kanderso-nrel?

kanderso-nrel:
Copying over all the sundry `requirements.txt` changes to the changelog does seem unnecessary; fine with me to say "we changed some things" and link the PRs.  I do like the idea of continuing to list the changes to `setup.py` though.

mdeceglie:
Regarding #326, that might go to `master` as a patch, so I suggest not holding up this PR

mdeceglie:
i _think_ I got the change log clean up correct...

mdeceglie:
Should `degradation_timeseries_plot()` be added to `__init__.py`?

mdeceglie:
I agree. The things we left out of __init__.py would produce a warning on rdtools import, and we didn't feel that this new function needed a warning, so I think it's ok to include, especially since this is for a beta release.

PR #346, opened 2022-09-26
336 hampel clearsky filter

cdeline:
- [x] Updated changelog
- [ ] Custom demonstration notebook documenting features


kanderso-nrel:
> At this point I am a little bit on the fence on how the new optional filters ta.filter_params['csi_filter'] and ta.filter_params['hampel_filter'] should be enabled.

Technically all existing filters are optional and can be enabled/disabled by the user by including/excluding their keys in the `filter_params`.  They're all enabled by default by associating their keys with empty dictionaries so that they don't error on the unpacking, but since they're empty they don't override the default parameter values in the underlying filter functions.  Removing a filter's key/value pair from `filter_params` makes the checks like `if 'poa_filter' in self.filter_params` evaluate to False and so the filter gets skipped.  Any reason not to use that same pattern here?

Something else to consider is whether we want to enable it by default in TrendAnalysis now (i.e. before 3.0) -- I think so far we've been conservative and making users opt in to results-changing features.

rdtools/analysis_chains.py
```diff
@@ -717,6 +744,8 @@ def _clearsky_preprocess(self):
         except AttributeError:
             raise AttributeError("No poa_global_clearsky. 'set_clearsky' must be run " +
                                  "prior to 'clearsky_analysis'")
+        if 'csi_filter' not in self.filter_params:
+            self.filter_params['csi_filter'] = {}
```
kanderso-nrel, line 748: This is another results-changing difference I'm not sure we should enable without a major version increment.

kanderso-nrel: Another question: isn't it valid to do a "clearsky-based" analysis without necessarily using the clear-sky filter?

rdtools/filtering.py
```diff
@@ -743,3 +743,35 @@ def xgboost_clip_filter(power_ac,
...
+        Threshold value, defaults to 3 sigma Pearson's rule.
+
+    Returns
+    -------
+    pandas.Series
+        Boolean Series of whether the given measurement is within 3 sigma of the
+        median.  False points indicate outliers to be removed.
+    '''
+    # Make copy so original not edited
+    vals_copy = vals.copy()
```
kanderso-nrel, line 770: I don't think a copy is necessary here

rdtools/test/filtering_test.py
```diff
@@ -333,3 +334,9 @@ def test_normalized_filter_default():
     pd.testing.assert_series_equal(normalized_filter(
                         pd.Series([0.01 - eps, 0.01 + eps, 1e308])),
                         pd.Series([False, True, True]))
+
+
+def test_hampel_filter():
+    a = pd.Series([5, 4, 7, 3, 4, 5, 4, 100, 3, 4, 5, 3, 5, 3, 5, 4, 12])
```
kanderso-nrel, line 340: Should we have tests with some NaN values as well?

rdtools/filtering.py
```diff
@@ -743,3 +743,35 @@ def xgboost_clip_filter(power_ac,
...
+    # Hampel Filter
+    L = 1.4826
+    rolling_median = vals_copy.rolling(k, center=True, min_periods=1).median()
```
kanderso-nrel, line 773: On our last RdTools call I said I thought `rolling(k, center=True)` didn't work if `k` was a pandas offset instead of an integer, but it works for me with pandas 1.4.2 so I retract that point.  (No change needed here, just wanted to correct my misstatement)

kanderso-nrel (commented):
Did we ever reach a conclusion on whether it makes sense to document new filters in existing notebooks versus creating new notebooks?  I'm not sure what I think.  Duplicating a bunch of content isn't really ideal, but neither is trying to stuff everything and the kitchen sink into a single notebook.

rdtools/filtering.py
cdeline: Note that I tried setting k = '14d' as a default value in filtering.hampel_filter, but it resulted in tests failing.  This may have more to do with how the tests were written vs anything fundamentally incompatible.

rdtools/analysis_chains.py
```diff
@@ -142,6 +143,7 @@ def __init__(self, pv, poa_global=None, temperature_cell=None, temperature_ambie
             'ad_hoc_filter': None  # use this to include an explict filter
         }
         self.filter_params_aggregated = {
+            'hampel_filter': {},
```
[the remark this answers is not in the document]
mdeceglie: It's possible that this will be part of the 3.0 release. The main goal here is to have a branch we can test and validate. Once we've done the validation exercise with this and other filtering options that have been proposed, we'll have a better sense for what the ultimate merge/version strategy will be. So at this point, I think it could go either way. The main thing is what makes it easiest for @kanderso-nrel to get this in the Fleets pipeline for validation and testing.

rdtools/filtering.py
mdeceglie: I suggest we get to the bottom of why the tests failed before giving up on the time-based default window.

rdtools/analysis_chains.py
```diff
@@ -418,6 +420,17 @@ def _filter(self, energy_normalized, case):
...
+        def _callCSI(filter_string):
+            if self.poa_global is None or self.poa_global_clearsky is None:
+                raise ValueError('Both poa_global and poa_global_clearsky must be available to '
+                                 f'do clearsky filtering with {filter_string}')
+            f = filtering.csi_filter(
+                self.poa_global, self.poa_global_clearsky, **self.filter_params[filter_string])
+            return f
+
         filter_components = {'default': pd.Series(
             True, index=energy_normalized.index)}
```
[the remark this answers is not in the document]
cdeline: done

rdtools/filtering.py
cdeline: done

rdtools/test/filtering_test.py
cdeline: done

rdtools/analysis_chains.py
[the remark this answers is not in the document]
cdeline: I would argue for separate example notebooks here.  My assumption is that with a major RdTools release (3.0?) we'll be making some of these improvements part of the new 'default' analysis.  In this case, a comparison with the old and new filter methods will mainly be a legacy question that someone could check out in a side notebook if they're really interested.  The Hampel filter approach is very lightweight and would be easy to enable by default.  The clearsky filtering requires additional processing and availability of csi irradiance, so might be better as an option.  I can re-work the example notebook to only focus on the hampel / clearsky filter improvements.

rdtools/analysis_chains.py
cdeline: This is removed - it's no longer logically necessary since we use a different key sensor_csi_filter.  so csi_filter can be enabled by default, but removed by the user in clearsky analysis if they really wanted to for some reason.

rdtools/analysis_chains.py
```diff
@@ -677,7 +692,19 @@ def _sensor_preprocess(self):
...
+            try:
+                if self.poa_global_clearsky is None:
+                    self._calc_clearsky_poa(model='isotropic')
+            except AttributeError:
+                # Review needed here: if user tries and fails to apply csi filter
+                # on sensor-based analysis then this provides a warning and disables
+                # csi filter.
+                warnings.warn("No poa_global_clearsky. 'set_clearsky' must be run " +
+                              "to allow filter_params['sensor_csi_filter']. " +
+                              " Disabling sensor_csi_filter.")
```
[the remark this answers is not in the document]
cdeline: Done

rdtools/filtering.py
cdeline: OK, here's some analysis of the different options.  Note that there is a very different result between 14 and '14d'.  In general, using '14d' leads to the incorrect dates being selected as outliers, often the date selected by '14d' is the day prior to when 14 would indicate a NaN, but not always.  This is the output of the test case test_sensor_analysis_hampel_filter() plotted in excel.  The points indicate where each outlier option indicates a NaN value.
image

cdeline: Here's the full difference spreadsheet
[hampel_comparison.xlsx](https://github.com/NREL/rdtools/files/10586562/hampel_comparison.xlsx)

cdeline: OK, when I run this on the example notebook, the difference between the two approaches is very similar.  I guess when you have 'actual' outlier data, they are both doing well.  Maybe '14d' is slightly better at removing outlier points.

k='14d' plot.  N=1991 remaining points
image

k=14 plot.  N=1989 remaining points
image

cdeline:
OK, the tests are failing at requirements-min.txt using Python 3.7.  There's currently a conflict between two things - the default value in the hampel filter is to use '14d' as the rolling window instead of '14'.  But for the pandas version in requirements-min.txt (pandas 0.23.2) using center=True for time-like data will throw an error.  This is documented here: https://github.com/pandas-dev/pandas/issues/20012 and was fixed in Pandas 1.3.0.

cdeline:
We've got two options - revert back from '14d' to '14' in the hampel filter, or bump our minimum requirements to pandas 1.3 and compatible packages.  The full list is as follows: pandas from 0.23.2 to 1.3.0, numpy from 1.16 to 1.17.3,  scipy from 1.1.0 to 1.2.0  and pvlib from 0.7.0 to 0.9.   Then it should work.

cdeline:
Per Mike, we're ok with bumping some of the minimum requirements - Pandas 1.3 has been out for long enough.

mdeceglie:
@cdeline We think that everything from this PR has been incorporated into #390 and this is stale. I will close this, but let me know if we've missed something. Thanks!

PR #348, opened 2022-10-14
added new private method for filtering on daily aggregated data

kperrynrel:
- [x] Updated changelog


kperrynrel:
Only private methods updated, no API.rst updates needed

rdtools/analysis_chains.py
```diff
@@ -475,6 +475,45 @@ def _filter(self, energy_normalized, case):
             self.clearsky_filter = bool_filter
             self.clearsky_filter_components = filter_components
 
+    def _daily_filter(self, aggregated, case):
+        """
+        Mirrors the _filter private function, but with daily filters applied.
+        These daily filters based on those in rdtools.filtering. Uses
+        self.filter_params, which is a dict, the keys of which are names of
```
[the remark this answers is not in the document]
kperrynrel: In some ways, after writing this, I'm thinking that maybe filter_params_daily is better as it's more explicit and we're less likely to use the wrong filter in the wrong section. I can add it in in my next push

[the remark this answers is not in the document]
kperrynrel: I added the new variable self.filter_params_daily by https://github.com/NREL/rdtools/pull/348/commits/668db1e6a05e659691a884e519504b8f799e26ed

rdtools/analysis_chains.py
```diff
@@ -475,6 +475,45 @@ def _filter(self, energy_normalized, case):
...
+            self.clearsky_filter_daily)
+
+        Returns
+        -------
+        None
+        """
+        filter_components_daily = {'default':
+                                   pd.Series(True, index=aggregated.index)}
+        # Add daily filters as they come online--this logic is omitted until
+        # we add the hampel clearsky filter and other daily filters.
```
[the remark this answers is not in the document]
kperrynrel: I added an ad_hoc_filter_daily option by https://github.com/NREL/rdtools/pull/348/commits/668db1e6a05e659691a884e519504b8f799e26ed. I copied the same logic we did for the other filter process @kanderso-nrel

rdtools/analysis_chains.py
```diff
@@ -252,7 +255,8 @@ def _calc_clearsky_poa(self, times=None, rescale=True, **kwargs):
         clearsky_poa = clearsky_poa['poa_global']
 
         if aggregate:
-            interval_id = pd.Series(range(len(self.poa_global)), index=self.poa_global.index)
+            interval_id = pd.Series(
+                range(len(self.poa_global)), index=self.poa_global.index)
```
kanderso-nrel, line 259: FYI we've adopted a max line length of 99 for rdtools (https://github.com/NREL/rdtools/blob/master/.flake8#L5), so no need to wrap at 80 in this repo

rdtools/analysis_chains.py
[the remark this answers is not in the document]
kanderso-nrel: Ok sounds good.  This docstring needs an update to reflect that

rdtools/analysis_chains.py
```diff
@@ -475,6 +482,63 @@ def _filter(self, energy_normalized, case):
...
+        # Convert the dictionary into a dataframe (after running filters)
+        filter_components_daily = pd.DataFrame(
+            filter_components_daily).fillna(False)
+        # Run the ad-hoc filter from filter_params_daily, if available
+        if self.filter_params_daily.get('ad_hoc_filter', None) is not None:
+            ad_hoc_filter_daily = self.filter_params_daily['ad_hoc_filter']
+
+            if ad_hoc_filter_daily.isnull().any():
+                warnings.warn(
+                    'ad_hoc_filter contains NaN values; setting to False (excluding)')
```
kanderso-nrel, line 519:
```suggestion
                    'daily ad_hoc_filter contains NaN values; setting to False (excluding)')
```

kanderso-nrel (commented):
Code looks good to me, so I guess next up is tests.  Probably makes sense to use the pre-aggregation filter tests in `analysis_chains_test.py` as a template here -- there are several relevant tests starting around `test_sensor_analysis_ad_hoc_filter`

kperrynrel:
Hey @kanderso-nrel and @mdeceglie I just added a unit test for the daily ad hoc filter and I wanted to get your input. The daily filter MUST have the same timestamp index as the aggregated data currently, which means there's a time zone attached:

image

Are we good with this? It means we need to make sure all daily filters return a series that is tz-aware.

kanderso-nrel:
In general (outside the context of TrendAnalysis) I think it makes sense to require our filter functions be implemented such that their outputs are tz-consistent with their inputs, so I think it's fine for TrendAnalysis to assume that if it passes a tz-aware aggregated Series to a filter function then it'll get a tz-aware mask back.

rdtools/analysis_chains.py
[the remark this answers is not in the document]
kanderso-nrel: This docstring still needs a small edit for "daily"

rdtools/test/analysis_chains_test.py
```diff
@@ -179,12 +179,25 @@ def test_sensor_analysis_ad_hoc_filter(sensor_parameters):
         rd_analysis.sensor_analysis(analyses=['yoy_degradation'])
 
 
+def test_sensor_analysis_daily_ad_hoc_filter(sensor_parameters):
```
kanderso-nrel, line 182: This looks good to me!  I think if we want complete test coverage we should also have daily analogs for:
- `test_filter_components`
- `test_filter_components_no_filters`
- `test_filter_ad_hoc_warnings`

kanderso-nrel (commented):
Getting close, a bit more testing and an entry in `docs\sphinx\source\changelog\pending.rst` and I think it'll be ready.  @mdeceglie should take a look too at some point.

Oh, this PR is also currently going to `master` but probably we'll want to switch it to `development` before merging.

rdtools/test/analysis_chains_test.py
kperrynrel: Added associated unit tests for the daily filter by https://github.com/NREL/rdtools/pull/348/commits/fde7cae818b3c5bbe015e5659a35b44d7df67de9

rdtools/test/analysis_chains_test.py
```diff
@@ -196,12 +223,21 @@ def test_filter_components_no_filters(sensor_parameters):
...
 
 
+def test_daily_filter_components_no_filters(sensor_parameters):
+    rd_analysis = TrendAnalysis(**sensor_parameters, power_dc_rated=1.0)
+    rd_analysis.filter_params = {}  # disable all index-based filters
+    rd_analysis.daily_filter_params = {}  # disable all daily filters
+    rd_analysis.sensor_analysis(analyses=['yoy_degradation'])
+    expected = pd.Series(True, index=rd_analysis.pv_energy.index)
+    pd.testing.assert_series_equal(rd_analysis.sensor_filter, expected)
+    assert rd_analysis.sensor_filter_components.empty
```
mdeceglie, line 233: I think these lines didn't get updated and are still checking the high frequency filter

mdeceglie (commented):
I think this is looking great, thanks for putting this together @kperrynrel. A couple items in addition to the specific comment I made in the tests:

1. Let's change the labeling of everything from "daily" to "aggregated" since we technically allow arbitrary aggregation frequency 
2. Let's add something to the docstring on the original filter to indicate it is for the aggregation filtering
3. We should add an explanation of `filter_params_daily` to the `__init__` docstring
4. This breaks the soiling analysis which wants to verify daily frequency. Perhaps after applying the daily filter the preprocess method can resample back to `self.aggregation_freq` filling with nans?

rdtools/test/analysis_chains_test.py
kperrynrel: Updated to check aggregated filter in https://github.com/NREL/rdtools/pull/348/commits/bd6084faad5b67a8c0302f055450a4d47468dc77

rdtools/test/analysis_chains_test.py
```diff
@@ -196,12 +223,23 @@ def test_filter_components_no_filters(sensor_parameters):
     assert rd_analysis.sensor_filter_components.empty
 
 
+def test_aggregated_filter_components_no_filters(sensor_parameters):
+    rd_analysis = TrendAnalysis(**sensor_parameters, power_dc_rated=1.0)
+    rd_analysis.filter_params = {}  # disable all index-based filters
+    rd_analysis.aggregated_filter_params = {}  # disable all daily filters
```
cdeline, line 229: this should be `filter_params_aggregated`

rdtools/analysis_chains.py
```diff
@@ -75,6 +75,14 @@ class TrendAnalysis():
...
         in which case those functions use default parameter values,  `ad_hoc_filter`
         defaults to None. See examples for more information.
+    filter_params_aggregated: dict
+        parameters to be passed to rdtools.filtering functions that specifically handle
+        aggregated data (dily filters, etc). Keys are the names of the rdtools.filtering functions.
+        Values are dicts of parameters to be passed to those functions. Also has a special key
+        `ad_hoc_filter`; this filter is a boolean mask joined with the rest of the filters.
+        filter_params_aggregated defaults to empty dicts for each function in rdtools.filtering,
+        in which case those functions use default parameter values,  `ad_hoc_filter`
+        defaults to None. See examples for more information.
```
mdeceglie, line 85: We'll need a way to distinguish which filters are aggregated and which are high-frequency. Options that come to mind are to put aggregated filters in their own module or to adopt a naming convention. (e.g. `aggregated_outlier_filter()`) Anyone have a preference?

rdtools/test/analysis_chains_test.py
kperrynrel: Good catch @cdeline, addressed in https://github.com/NREL/rdtools/pull/348/commits/85886d11d5007bd686575f27e14981ea996a1b5f

docs/sphinx/source/changelog/pending.rst
```diff
@@ -0,0 +1,23 @@
...
+------------
+* Upgrade the notebook environment from python 3.7 to python 3.10.
+  Several dependency versions in ``docs/notebook_requirements.txt`` are
+  updated as well. (:issue:`319`, :pull:`326`)
+* Bump ``ipython==7.16.3``, ``jupyter-console==6.4.0``,
+  and ``prompt-toolkit==3.0.27`` in ``docs/notebook_requirements.txt``
+  and bump ``Pillow==9.0.0`` in ``requirements.txt`` (:pull:`314`)
+* Bump ``sphinx`` version from 3.2 to 4.5 and ``nbsphinx`` version
+  from 0.8.5 to 0.8.8 in the optional ``[doc]`` requirements (:pull:`317`, :pull:`325`)
+* A number of other requirements updates (:pull:`337`)
```
kanderso-nrel, line 19: I think these old `pending` entries are already covered in `v2.2.0-beta.0.rst`; let's clean out everything but the entry for this PR

rdtools/analysis_chains.py
mdeceglie: Capturing an update from a meeting: It's possible that some filters would technically work for both high-frequency and aggregated data. The most general option would be to keep everything in the single filtering module, but specify in the TrendAnalysis docstring which keys are appropriate for which attribute. (We don't need to finalize this for this PR)

docs/sphinx/source/changelog/pending.rst
kperrynrel: Updated to just this PR in pending.rst in commit https://github.com/NREL/rdtools/pull/348/commits/c3838d65fc6a0b71a961455ff6a36641c7054cc7

mdeceglie:
@cdeline Did you mention that you still have a bug with this? Could you provide an example that occurs with the most up to date version of this branch? Thanks!

rdtools/analysis_chains.py
```diff
@@ -651,8 +732,12 @@ def _clearsky_preprocess(self):
...
         cs_aggregated, cs_aggregated_insolation = self._aggregate(
             cs_normalized[self.clearsky_filter], cs_insolation[self.clearsky_filter])
-        self.clearsky_aggregated_performance = cs_aggregated
-        self.clearsky_aggregated_insolation = cs_aggregated_insolation
+        # Run daily filters on aggregated data
+        self._aggregated_filter(cs_aggregated, 'clearsky')
+        # Apply daily filter to aggregated data and store
+        self.clearsky_aggregated_performance = cs_aggregated[self.clearsky_filter_aggregated]
+        self.clearsky_aggregated_insolation = \
+            cs_aggregated_insolation[self.clearsky_filter_aggregated]
```
cdeline, line 741: @kperrynrel  - I'm still getting errors when I exercise the aggregated filter on clearsky conditions.  On lines 703-706 of `_sensor_preprocess` you re-index the data, but don't do the same thing here in `_clearsky_preprocess`.  Copying those four lines into here to re-establish the originally indexed time series will fix the errors.  Thanks!

kanderso-nrel:
linter failures are from #351 and can be ignored

PR #354, opened 2022-11-30
Version 2.1.4

mdeceglie:
- [x] Updated changelog


docs/sphinx/source/changelog/v2.1.4.rst
```diff
@@ -18,3 +26,10 @@ Requirements
 * Bump ``sphinx`` version from 3.2 to 4.5 and ``nbsphinx`` version
   from 0.8.5 to 0.8.8 in the optional ``[doc]`` requirements (:pull:`317`, :pull:`325`)
 * A number of other requirements updates (:pull:`337`)
```
kanderso-nrel, line 28: Some of these entries are already listed in `v2.2.0-beta.0`; should we delete them here?  They do apply to 2.1.4, but I don't think it makes sense to list them under multiple releases.

mdeceglie: Good point. The puzzle is that `v2.2.0-beta.0` is not a parent of this, so if we deleted them, they would be nowhere in the change log on `master`. When we merge `master` back into `development` after this release, they would appear twice on the development change log; we could delete them at that point. But that would create a confusing situation when we do a real 2.2 release because they would appear to have been added in a version later than they really were. Maybe the best thing is to try and remember to delete them from the 2.2.0-beta.0 change log when we release 2.2? Or just leave them as redundant, because the downside isn't that big. After all, they were introduced in different releases in our parallel universes.

kanderso-nrel: Agreed that it's not the end of the world to list changes under multiple versions, and maybe that's the best option here.  Whatever you want to do is fine with me.  If we are going to list these changes twice, we should consider copying the updated wording in the `v2.2.0-beta.0` changelog.

docs/sphinx/source/changelog/v2.1.4.rst
```diff
@@ -1,6 +1,14 @@
-************************
-Pending
-************************
+**************************
+v2.1.4 (November 30, 2022)
+**************************
```
kanderso-nrel, line 3:
```suggestion
*************************
v2.1.4 (December 1, 2022)
*************************
```

kanderso-nrel (approved):
Aside from incrementing the release date, LGTM

PR #362, opened 2023-03-03
Fixes for pandas 2.0.0 compatibility

kandersolar:
- [x] Code changes are covered by tests
- [x] Code changes have been evaluated for compatibility/integration with TrendAnalysis
~- [ ] New functions added to `__init__.py`~
~- [ ] API.rst is up to date, along with other sphinx docs pages~
~- [ ] Example notebooks are rerun and differences in results scrutinized~
- [x] Updated changelog

Closes #361.

PR #366, opened 2023-04-26
remove warnings from logic clip filter

### LINKED ISSUE #365: 3.0.0a1 logic clipping filter warnings
using the logic filter with `ta.filter_params['clip_filter'] = {'model':"logic"}` leads to warnings about the experimental nature of the logic clipping filter.  This was, in fact, the subject of a 2021 PVSC publication and has been validated, and I doubt that the core details of this implementation will change in the future.  I propose removing the experimental label for our future 3.0 release.  This will also clean up logfiles when we're running RdTools 1000's of times on the fleet pipeline.

cdeline:
~~Code changes are covered by tests~~
~~Code changes have been evaluated for compatibility/integration with TrendAnalysis~~
~~New functions added to `__init__.py`~~
~~API.rst is up to date, along with other sphinx docs pages~~
~~Example notebooks are rerun and differences in results scrutinized~~
- [ ] Updated changelog

Fixes #365.  Logic clipping filter is published and validated, not likely to change in the future.  for the 3.0 release (and 3.0.0a1 runs) it would be nice to get rid of the logfile clutter.  We can still decide on whether to make it the default method or not, but the method itself is stable, or at least will be by the time of any 3.0.0 public release..

### DIFF
--- a/rdtools/filtering.py
+++ b/rdtools/filtering.py
@@ -471,11 +471,11 @@ def logic_clip_filter(power_ac,
        detection techniques in AC power time series", 2021 IEEE 48th Photovoltaic
        Specialists Conference (PVSC). DOI: 10.1109/PVSC43889.2021.9518733.
     '''
-    # Throw a warning that this is still an experimental filter
-    warnings.warn("The logic-based filter is an experimental clipping filter "
-                  "that is still under development. The API, results, and "
-                  "default behaviors may change in future releases (including "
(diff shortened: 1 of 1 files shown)


mdeceglie:
I'm OK with this change. I think we can ignore all the failing tests at this stage. Some will be fixed by #362 , others have been introduced in the `aggregated_filters_for_trials` branch and we will clean them up when we figure out what 3.0 will really look like. Do you agree @cdeline ?

Issue #372, opened 2023-05-18
NonExistentTimeError in rdtools.get_clearsky_tamb


BernatNicolau:
I have a dataset which has a Daylight Saving Time at 00:00h:

2018-08-11 22:00:00-04:00
2018-08-11 23:00:00-04:00
2018-08-12 01:00:00-03:00
2018-08-12 02:00:00-03:00

When running rdtools.get_clearsky_tamb, this error occurs 
NonExistentTimeError: 2018-08-12 00:00:00

Triggered in line 59 of rdtools/clearsky_temperature.py:

dt_daily = pd.date_range(times.date[0] - buffer, times.date[-1] + buffer, freq='D', tz=times.tz)

mdeceglie:
Thanks for submitting this issue @BernatNicolau. Can you please provide the complete code for a minimal example that reproduces the error?

BernatNicolau:
Hello,

Sure, here it is:


```
import datetime
import pandas as pd
import rdtools

tz = 'America/Santiago'
start_date = datetime.datetime(2018, 1, 1, 0, 0, 0)
end_date = datetime.datetime(2023, 5, 18, 23, 0, 0)
freq ='H'
lat = -24
lon = -70

times = pd.date_range(start=start_date, end=end_date, freq=freq)
df = pd.DataFrame(index=times)

df.index = df.index.tz_localize(tz = tz, ambiguous=True, nonexistent='shift_forward')

tamb = rdtools.get_clearsky_tamb(df.index, lat, lon)
```

I think the problem might be that in rdtools.get_clearsky_tamb, there's not the option to add a nonexistent order. If there is, I'm sorry for the time spent...

mdeceglie:
Thanks, that helps me reproduce the error. It seems the underlying issue is this: https://github.com/pandas-dev/pandas/issues/30378 I think I have a workaround in #373

BernatNicolau:
Thank you very much @mdeceglie !

mdeceglie:
Closed with https://github.com/NREL/rdtools/pull/373 and version 2.1.6

PR #373, opened 2023-05-19
avoid NonExistentTimeError in get_clearsky_tamb()

mdeceglie:
Fix for #372 
- [x] Updated changelog


mdeceglie:
@BernatNicolau thanks for your help in identifying and isolating this bug. If you'd like your name to appear next to your github handle under "contributors" for this update, please let me know what your full name is. Thanks!

BernatNicolau:
> @BernatNicolau thanks for your help in identifying and isolating this bug. If you'd like your name to appear next to your github handle under "contributors" for this update, please let me know what your full name is. Thanks!

Thank you for fixing it. My name is Bernat Nicolau and it would be great to appear in contributors!

kandersolar:
What a fun timezone bug :P

This fix seems reasonable to me.  But I think we should consider not using the local time zone for the daily time series at all.  In addition to causing the weird issue @BernatNicolau noticed, I suspect it's to blame for the following undesirable behavior, where you get different temperatures for equivalent timestamps with different time zones: 

```python
In [1]: import rdtools
   ...: import pandas as pd
   ...:
   ...: times_utc = pd.date_range('2019-01-01', freq='h', periods=3, tz='UTC')
   ...: times_est = times_utc.tz_convert('Etc/GMT+5')
   ...:
   ...: latitude = 40
   ...: longitude = -80
   ...:
   ...: df = pd.DataFrame({
   ...:     'tamb_utc': rdtools.get_clearsky_tamb(times_utc, latitude, longitude),
   ...:     'tamb_est': rdtools.get_clearsky_tamb(times_est, latitude, longitude).tz_convert('UTC'),
   ...: })
   ...:
   ...: df
Out[1]:
                           tamb_utc  tamb_est
2019-01-01 00:00:00+00:00 -1.516891 -1.499855
2019-01-01 01:00:00+00:00 -1.863861 -1.846259
2019-01-01 02:00:00+00:00 -2.280853 -2.262572
```

Maybe it would make more sense to do all the calculations in UTC and just convert results to the input timezone at the end.

mdeceglie:
Thanks for taking a look at this @kandersolar. I think it will be a bit of a project to trace the logic through and make sure that doing all the calculations in UTC doesn't affect the original intent of the model. The model uses the hour of the day to construct the the temperature signal, and it makes sense for that hour of the day to be local. 

Unless you have a solution ready to go, how would you feel about proceeding with this fix for now and opening your suggestion as an issue?

mdeceglie:
I also note that the docstring asks for the times in local time, so passing UTC times is a bit of an off-label use 😃

kandersolar:
> Unless you have a solution ready to go, how would you feel about proceeding with this fix for now and opening your suggestion as an issue?

Sure!

> I also note that the docstring asks for the times in local time, so passing UTC times is a bit of an off-label use 😃

Good point!  Perhaps much ado about nothing, then.  I leave it to you whether an issue is warranted.

PR #375, opened 2023-05-24
Hour angle filter

mdeceglie:
- [ ] Updated changelog


rdtools/analysis_chains.py
```diff
@@ -464,6 +464,18 @@ def _call_clearsky_filter(filter_string):
             f = filtering.clip_filter(
                 self.pv_power, **self.filter_params['clip_filter'])
             filter_components['clip_filter'] = f
+        if 'hour_angle_filter' in self.filter_params:
+            if self.pvlib_location is None:
```
cdeline, line 468: if undefined, self.pvlib_location won't be None, it'll actually return an AttributeError.  I've got a fix for the handful of places this crops up.

cdeline (approved):
It's working for me, the run time is a bit slow though.  Maybe something to optimize later.

mdeceglie:
Our tests are failing because the pvlib clearsky function doesn't seem to like our artificial test case. Will revisit when it comes time for a more stable release.

PR #379, opened 2023-07-28
Release/2.1.6

mdeceglie:
- [x] Updated changelog


docs/sphinx/source/changelog/v2.1.6.rst
```diff
@@ -0,0 +1,37 @@
...
+
+Testing
+-------
+* Extended pip timeout (:pull:`360`)
+* Updated example notebooks with new figure sizes (:pull:`360`)
+
+Contributors
+------------
+* Michael Deceglie (:ghuser:`mdeceglie`)
+* Bernat Nicolau (:ghuser:'BernatNicolau')
```
kandersolar, line 36:
```suggestion
* Bernat Nicolau (:ghuser:`BernatNicolau`)
```

kandersolar:
I notice that PyPI rejected the automatic wheel upload, seemingly due to the package's version number not being determined correctly (`0+untagged.1`).  Since version numbers are derived from git tags, I guess the issue is that the `2.1.6` tag was not present in the git repo clone used in the distribution build job.  It seems that the job checked out the tagged commit, but was not aware of the tag itself.  See this [line](https://github.com/NREL/rdtools/actions/runs/5716155816/job/15487032428#step:2:371) in the 2.1.6 deploy job, compared with the [same](https://github.com/NREL/rdtools/actions/runs/4997166824/job/13533332502#step:2:371) in the 2.1.5 job.

So I think that's the problem, but I'm not sure how to fix it for future deployments.  `git pull --tags` or similar might be helpful, but I don't know why this would be needed now when it wasn't needed for 2.1.5.  The release seems properly associated with the `2.1.6` tag to me.  In any case, looks like the 2.1.6 distribution files need to be built and uploaded to PyPI manually.

PR #389, opened 2023-09-12
change filt variable from global to local

martin-springer:
- [ ] Updated changelog

### DIFF
--- a/rdtools/normalization.py
+++ b/rdtools/normalization.py
@@ -428,7 +428,7 @@ def irradiance_rescale(irrad, irrad_sim, max_iterations=100,
 
         guess = np.percentile(irrad.dropna(), 90) / \
             np.percentile(irrad_sim.dropna(), 90)
-        min_result = minimize(_rmse, guess, method='Nelder-Mead')
+        min_result = minimize(_rmse, guess, (filt), method='Nelder-Mead')
         factor = min_result['x'][0]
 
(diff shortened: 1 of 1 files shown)


rdtools/normalization.py
```diff
@@ -392,10 +392,9 @@ def _rmse(fact):
 
         def _single_rescale(irrad, irrad_sim, guess):
             "Optimizes rescale factor once"
-            global filt
             csi = irrad / (guess * irrad_sim)  # clear sky index
             filt = (csi >= 0.8) & (csi <= 1.2) & (irrad > 200)
-            min_result = minimize(_rmse, guess, method='Nelder-Mead')
+            min_result = minimize(_rmse, guess, (filt), method='Nelder-Mead')
```
mdeceglie, line 397: I think we may need to use a lambda function like `lambda x : _rmse(x, filt)` here as the first argument to `minimize`

martin-springer: @mdeceglie - I'm not seeing why we might need a lambda function. I created a jupyter notebook that re-creates the analysis and I obtain the same result between my suggested changes and the initial implementation. I opened a new PR #399 from the master branch to have running tests and to check whether my changes cause any troubles. So far, all pytests seem to be passing.

mdeceglie:
Replaced with #399

PR #400, opened 2023-11-17
return results instead of none when CODS soiling signal is small

martin-springer:
- ~~[ ] New functions added to `__init__.py`~~
- ~~[ ] API.rst is up to date, along with other sphinx docs pages~~
- [x] Updated changelog

### DIFF
--- a/rdtools/test/soiling_cods_test.py
+++ b/rdtools/test/soiling_cods_test.py
@@ -48,7 +48,7 @@ def test_iterative_signal_decomposition(cods_normalized_daily):
         ['soiling_ratio', 'soiling_rates', 'cleaning_events',
          'seasonal_component', 'degradation_trend', 'total_model', 'residuals']]
     pd.testing.assert_series_equal(expected_means, df_out.mean(),
-                                   check_exact=False, check_less_precise=True)
+                                   check_exact=False, rtol=1e-3)
 
 
 def test_iterative_signal_decomposition_with_nan_interval(cods_normalized_daily):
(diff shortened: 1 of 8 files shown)


mdeceglie:
I re-ran the offending notebook and scrutinized the output to be sure no substantial changes occurred in the outputs. That plot does appear to change size, but I don't think it is a concern.

rdtools/test/soiling_cods_test.py
```diff
@@ -111,13 +111,26 @@ def test_soiling_cods(cods_normalized_daily):
...
-        assert x in expected_summary_columns,\
+        assert x in expected_summary_columns, \
             "'{}' not an expected column in result_df]".format(x)
     for x in expected_summary_columns:
-        assert x in actual_summary_columns,\
+        assert x in actual_summary_columns, \
             "'{}' was expected as a column, but not in result_df".format(x)
 
 
+def test_soiling_cods_small_signal(cods_normalized_daily_small_soiling):
```
martin-springer, line 121: I added the CODS small signal test

martin-springer:
I did update all packages in my conda environment and saved the notebook. Somehow, there's still a discrepancy in one cell. I could try re-creating a new environment the exact way nbval does here. On the other hand, we get quite a lot of warning and deprecation messages. Might be time to start resolving them.

mdeceglie:
I think we can look past the notebook failure this time I opened #401 today because of the message we're seeing in the notebook.

mdeceglie:
Let's get the change log updated, then I think we can move forward with this. The version will be `2.2.0-beta.2`

martin-springer:
Sounds good, I updated the changelog. Could you also merge these changes into `aggregated_filters_for_trial` and create a tag for the next pvfleets run?

mdeceglie (approved):
Looks good. Thanks @martin-springer!

PR #413, opened 2024-04-29
Add missing tests

martin-springer:
This is a draft. Mainly to get Codecov updated.

- [ ] Updated changelog

### DIFF
--- a/rdtools/test/irradiance_rescale_test.py
+++ b/rdtools/test/irradiance_rescale_test.py
@@ -16,13 +16,8 @@ def simple_irradiance():
 def test_rescale(method, simple_irradiance):
     # test basic functionality
     if method == "error":
-        pytest.raises(
-            ValueError,
-            irradiance_rescale,
-            simple_irradiance,
-            simple_irradiance * 1.05,
-            method=method,
-        )
+        with pytest.raises(ValueError):
+            irradiance_rescale(simple_irradiance, simple_irradiance * 1.05, method=method)
     else:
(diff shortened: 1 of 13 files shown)


rdtools/test/irradiance_rescale_test.py
```diff
@@ -7,19 +7,28 @@
...
-    expected = measured
-    assert_series_equal(rescaled, expected, check_exact=False)
+    if method == "error":
+        pytest.raises(
+            ValueError,
+            irradiance_rescale,
+            simple_irradiance,
+            simple_irradiance * 1.05,
+            method=method,
+        )
```
mdeceglie, line 25: Same question as above regarding use of `pytest.raises` without `with`

mdeceglie (commented):
@martin-springer  Thanks for putting this together, a few questions but otherwise looks great.

mdeceglie:
@martin-springer  is it possible to re-trigger the codecov report?

martin-springer:
> @martin-springer is it possible to re-trigger the codecov report?

I do believe so, but haven't done that much myself yet. Looking through the quick start guide (https://docs.codecov.com/docs/quick-start), you can execute the upload of coverage reports from the command line. This page outlines the process (https://docs.codecov.com/docs/codecov-uploader). 

We might want to tweak the github action, to not only run when a new PR is created but also when commits are made to a PR. This should then take care of it. I can take a look at how we would need to change the config for that.

PR #423, opened 2024-06-26
Advance pvlib upper bound from 0.11 to 0.12

kandersolar:
- [ ] Updated changelog

RdTools currently requires pvlib<0.11.0. With pvlib now in the 0.11.x series, I don't think there turned out to be any relevant breaking changes (see [what's new](https://pvlib-python.readthedocs.io/en/stable/whatsnew.html)), so we can increase that upper limit to allow 0.11.*.


mdeceglie:
Thanks for the PR @kandersolar! For ease of managing the flow here, I'm going to sweep this into our changes for 3.0 expected this summer.

mdeceglie:
It's not immediately clear which of the breaking changes listed in the pvlib docs are causing the tests to fail. any thoughts @kandersolar?

kandersolar:
@mdeceglie it looks to be fixed now!

mdeceglie (approved):

PR #428, opened 2024-08-21
Add python 3.11 and 3.12 versions to testing matrix

martin-springer:
- [x] Updated changelog


rdtools/test/conftest.py
```diff
@@ -9,30 +9,32 @@
...
-rdtools_base_version = \
-    parse_version(parse_version(rdtools.__version__).base_version)
+# rdtools_base_version = \
+#     parse_version(parse_version(rdtools.__version__).base_version)
 
 
 # decorator takes one argument: the base version for which it should fail
 # for example @fail_on_rdtools_version('3.0.0') will cause a test to fail
 # on rdtools versions 3.0.0, 3.0.0-alpha, 3.1.0, etc
-def fail_on_rdtools_version(version):
```
martin-springer, line 19: @mdeceglie - pkg_resources is depricated for python 3.12. I didn't find a single instance in the tests where this decorator is used. Should we deprecate the decorator or refactor it to work with the importlib.metadata package?

martin-springer:
@mdeceglie - Regarding numpy > 2. We would need pandas 2.2.2 (https://pandas.pydata.org/docs/whatsnew/v2.2.2.html). However, the bug you found in pandas is still persistent in 2.2.2 (https://github.com/pandas-dev/pandas/issues/55794). 

Not sure, how to proceed...

mdeceglie:
Good catch @martin-springer the only place I can find where that bug comes into play is in `get_clearsky_tamb()` where the frequency of a series gets checked twice, one on the whole thing and again on the first 10 elements if the first try fails. Updating to pandas 2.2.2 sure enough caused an error associated with the pandas bug in our testing of that function. I updated `get_clearsky_tamb()` with @kandersolar's work around and that seems to work. (Also fixing another testing error that cropped up)

mdeceglie:
Some of these minimum requirements are the latest versions, what's driving the need for such recent version in our minimum requirements?

rdtools/availability.py
```diff
@@ -531,6 +536,11 @@ def _combine_losses(self, rollup_period='M'):
             The period on which to roll up losses and calculate availability.
         """
 
+        # Allow pandas < 2.0 to use 'M' as an alias for MonthEnd
+        # https://pandas.pydata.org/docs/whatsnew/v2.2.0.html#deprecate-aliases-m-q-y-etc-in-favour-of-me-qe-ye-etc-for-offsets
+        if rollup_period == "ME":
```
mdeceglie, line 541: Let's update the docstrings for _combine_losses and run to explain the situation. Maybe something like "Default value "ME" triggers monthly rollup period. For other aliases, be sure they are compatible with your version of Pandas"

rdtools/test/filtering_test.py
```diff
@@ -521,4 +519,4 @@ def test_hour_angle_filter():
 
     # Check that the result is the correct boolean Series
     expected_result = pd.Series([False, False, True, True, True], index=index)
-    pd.testing.assert_series_equal(result, expected_result)
+    pd.testing.assert_series_equal(result, expected_result, check_names=False)
```
mdeceglie, line 522: What caused the need for this change?

setup.py
```diff
@@ -32,60 +32,59 @@
...
-    'filterpy >= 1.4.2'
+    "matplotlib >= 3.5.3",
+    "numpy >= 1.22.4",
+    "pandas >= 1.4.4",
+    "statsmodels >= 0.13.5",
+    "scipy >= 1.8.1",
+    "h5py >= 3.7.0",
+    "plotly>=4.0.0",
+    "xgboost >= 1.3.3",
+    "pvlib >= 0.9.0, <0.12.0",
```
mdeceglie, line 54: requirments_min specifies 0.11. We should change the minimum here too right?

mdeceglie (changes requested):
Looks good, a few minor comments to address, then it should be ready to go. Thanks!

rdtools/test/filtering_test.py
martin-springer: In the newest pvlib version 0.11.1 there was a bug fix for the hour_angle() function (https://pvlib-python.readthedocs.io/en/stable/whatsnew.html). 

The series that's returned does have a new argument `Name: equation_of_time`. 
image

In order to make the test run for older pvlib versions in which the name attribute is not present, I decided to not add the name to the expected result series but rather don't check for series names.

setup.py
martin-springer: Absolutely right, I overlooked that. I've bumped it to 0.11 now.

rdtools/availability.py
martin-springer: Good call. I've updated the docstrings.

mdeceglie (approved):

PR #426, opened 2024-08-02
SRR and CODS updates

martin-springer:
- [ ] Code changes are covered by tests
- [ ] Code changes have been evaluated for compatibility/integration with TrendAnalysis
- [ ] New functions added to `__init__.py`
- [ ] API.rst is up to date, along with other sphinx docs pages
- [ ] Example notebooks are rerun and differences in results scrutinized
- [ ] Updated changelog

PR #429, opened 2024-09-17
update examples and docs for RdTools 3

mdeceglie:
- [ ] Updated changelog


mdeceglie:
To do: Get the example file on the DuraMAT datahub updated to include the NSRDB data

mdeceglie:
@martin-springer @kandersolar  Any ideas on how I managed to break the nbval tests here?

martin-springer:
> @martin-springer @kandersolar Any ideas on how I managed to break the nbval tests here?

A missing comma in the workflow file and upload-artifact had to be updated to v4. 

There's an issue tough with the newer version of nbval ignoring the semicolon: https://github.com/computationalmodelling/nbval/issues/194

We need to decide on how we want to address that.

mdeceglie:
@kandersolar Is it time to remove the experimental warning from the availability module? Seems pretty locked in, we've used it in a number of different applications successfully, and I"m not aware of plans to modify it.

kandersolar:
I have no plans to change the method either.  No objection from me on removing the warning.

docs/degradation_and_soiling_example.ipynb
martin-springer, line 1: cell 1:
I believe the python version should be updated to 3.12?

cell 5: semicolon after:
ax.set_ylabel('soiling signal')**;**

cell 8:  whitespace in between = 800
example_subset.loc[example_subset['ac_power']>800,'ac_power'] =800

cell 14:
Note the default confidence_level of 68.2 (should this be percent?)

Section 5:
There are still some warnings popping up that the soiling module is currently experimental. Should those be removed?

Cell 19:
Same question as above should the confidence interval be in %?

Cell 30: semicolon after:
ax3.set_ylabel('Residuals')**;**

martin-springer (approved):
Looking great! I only found a few minor things.

docs/degradation_and_soiling_example.ipynb
mdeceglie: Good catches!

For cell 14 I think we should leave off the percent because this is an explicit parameter value in that context.

For the soiling warnings, we should leave them in until we resolve #426

PR #434, opened 2024-10-30
add citation.cff file

martin-springer:
- [x] Updated changelog


mdeceglie (commented):
In order to keep maintenance to a minimum, thoughts on removing version number and release date? I am ambivalent about it. If we leave it the same we should consider adding this to the PR checklist so that it doesn't get forgotten with new releases.

martin-springer:
> In order to keep maintenance to a minimum, thoughts on removing version number and release date? I am ambivalent about it. If we leave it the same we should consider adding this to the PR checklist so that it doesn't get forgotten with new releases.

Agreed. I like to keep it simple and removed the version and release date. I've also added Quyen and Noah and updated the reference in the readme.

martin-springer:
Looks like scikit-learn just had a new release (https://scikit-learn.org/stable/whats_new/v1.6.html#version-1-6-0) and introduced a new tag `__sklearn_tags__` that seems to be buggy and makes one of our tests fail.

mdeceglie:
Let's use a "pending" section in the change log since 3.0.0b0 is already out.

mdeceglie (approved):

Issue #377, opened 2023-06-06
TrendAnalysis should raise an error for unknown filters


mdeceglie:
**Is your feature request related to a problem? Please describe.**
TrendAnalysis checks its attribute `filter_params` (and `filter_params_aggregated`) to control filtering. These attributes are dicts, where the keys are the names of the filter functions to use. If an unknown filter function name is supplied, TrendAnalysis quietly ignores it. 

**Describe the solution you'd like**
Raise an error

martin-springer:
addressed with PR #436

PR #436, opened 2024-11-06
Add checks to filter arguments for TrendAnalysis (Issue #377)

martin-springer:
- [ ] Updated changelog

### DIFF
--- a/docs/sphinx/source/changelog/v3.0.0-beta.0.rst
+++ b/docs/sphinx/source/changelog/v3.0.0-beta.0.rst
@@ -22,7 +22,8 @@ Enhancements
 * Added codecov.yml configuration file (:pull:`420`)
 * Availability module no longer considered experimental (:pull:`429`)
 * Add capability to seed the CircularBlockBootstrap (:pull:`429`)
-* Allow sub-daily aggregation in :py:func:`~rdtools.degradation.degradation_year_on_year` (:pull:`390`) 
+* Allow sub-daily aggregation in :py:func:`~rdtools.degradation.degradation_year_on_year` (:pull:`390`)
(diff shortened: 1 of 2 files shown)


docs/sphinx/source/changelog/v3.0.0-beta.0.rst
```diff
@@ -22,7 +22,8 @@ Enhancements
 * Added codecov.yml configuration file (:pull:`420`)
 * Availability module no longer considered experimental (:pull:`429`)
 * Add capability to seed the CircularBlockBootstrap (:pull:`429`)
-* Allow sub-daily aggregation in :py:func:`~rdtools.degradation.degradation_year_on_year` (:pull:`390`) 
+* Allow sub-daily aggregation in :py:func:`~rdtools.degradation.degradation_year_on_year` (:pull:`390`)
+* Added checks to TrendAnalysis for `filter_params` and `filter_params_aggregated`. Raises an error if unkown filter is supplied. (:pull:`436`)
```
mdeceglie, line 26: Let's add this to a pending section of the change log

mdeceglie (changes requested):
The solution works great! Let's add some simple tests for the intended behavior

martin-springer:
> The solution works great! Let's add some simple tests for the intended behavior

Done.

mdeceglie:
Thank you for adding the unit tests. I suggest we also add some simple tests to the trend analysis as well: checking that every key we think should work does work (including ad_hoc) and checking that some bogus key in either `filter_params` or `filter_params_aggregated` raises an error

martin-springer:
> Thank you for adding the unit tests. I suggest we also add some simple tests to the trend analysis as well: checking that every key we think should work does work (including ad_hoc) and checking that some bogus key in either `filter_params` or `filter_params_aggregated` raises an error

added

Issue #370, opened 2023-05-17
`energy_from_power` returns incorrect index for shifted hourly data


kandersolar:
TrendAnalysis does not like hourly data where the timestamps are not at the top of the hour.  Here is a toy example where it errors when given an input time series timestamped at the half-hour mark instead of the usual `00`:

```python
import pandas as pd
import numpy as np
import rdtools

# toy hourly dataset, note that minute=30
times = pd.date_range('2019-01-01 00:30:00', periods=8760*3, freq='H')
df = pd.DataFrame({'pv': 1 - 0.005*np.arange(len(times))/8760,
                   'poa_global': 1000,
                   'temperature_cell': 25},
                  index=times)

ta = rdtools.TrendAnalysis(**df)
ta.filter_params.pop('clip_filter')  # clipping filter doesn't like this toy dataset
ta.sensor_analysis()  # ValueError: Less than two years of data left after filtering
```

Inspection of the object's attributes reveals that `pv_energy`'s index is misaligned with the original data's index (the minutes have been truncated to `00`):
```python
In [181]: ta.pv_energy[:3]
Out[181]: 
2019-01-01 01:00:00    0.500000
2019-01-01 02:00:00    0.999999
2019-01-01 03:00:00    0.999999
Freq: H, Name: energy_Wh, dtype: float64
```
Index misalignment between energy and the irradiance data messes up the filtering, resulting in the "less than two years" error.  Of course, the index difference traces back to `energy_from_power`:
```python
In [189]: rdtools.normalization.energy_from_power(df['pv'])[:3]
Out[189]: 
2019-01-01 01:00:00    0.500000
2019-01-01 02:00:00    0.999999
2019-01-01 03:00:00    0.999999
Freq: H, Name: energy_Wh, dtype: float64
```

And from there, I think, to the call to `_aggregate` and a `resample` it performs:

https://github.com/NREL/rdtools/blob/d19fa83dd2b17a32129383976eb47d94be5c9ae8/rdtools/normalization.py#L615-L618

martin-springer:
addressed with PR #437

PR #437, opened 2024-11-06
Fix issue #370 (energy_from_power returns incorrect index for shifted hourly data)

martin-springer:
- [ ] Updated changelog


docs/sphinx/source/changelog/v3.0.0-beta.0.rst
```diff
@@ -31,6 +31,7 @@ Bug fixes
 * Deploy workflow was replaced with trusted publisher workflow for pypi (:pull:`427`)
 * Fix pandas 2.0.0 deprications and update syntax changes (:pull:`428`)
 * Fix numpy 2.0.0 deprications and update syntax changes (:pull:`428`)
+* Fix `energy_from_power`` returns incorrect index for shifted hourly data (:issue:`370`, :pull:`437`)
```
mdeceglie, line 34: Let's move this to a pending changelog section

mdeceglie (changes requested):
The solution looks good. I suggest we add a test, can just use the example from the issue.

docs/sphinx/source/changelog/v3.0.0-beta.0.rst
martin-springer: Done.

martin-springer:
> The solution looks good. I suggest we add a test, can just use the example from the issue.

Done.

rdtools/test/analysis_chains_test.py
```diff
@@ -743,6 +743,23 @@ def test_plot_degradation_timeseries(sensor_analysis, clearsky_analysis):
     )
 
 
+def test_energy_from_power_hourly_data():
+
+    times = pd.date_range("2019-01-01 00:00:00", periods=3, freq="H")
```
mdeceglie, line 748: `H` is deprecated, let's change to `h`
```suggestion
    times = pd.date_range("2019-01-01 00:00:00", periods=3, freq="h")
```

rdtools/test/analysis_chains_test.py
```diff
@@ -743,6 +743,23 @@ def test_plot_degradation_timeseries(sensor_analysis, clearsky_analysis):
...
+    pv = pd.Series([1.2, 2.8, 2.0], index=times)
+
+    energy = normalization.energy_from_power(pv)
+    pd.testing.assert_series_equal(energy, pv[1:], check_names=False)
+
+
+def test_energy_from_power_shifted_hourly_data():
+
+    times = pd.date_range("2019-01-01 00:30:00", periods=3, freq="H")
```
mdeceglie, line 757:
```suggestion
    times = pd.date_range("2019-01-01 00:30:00", periods=3, freq="h")
```

mdeceglie (commented):
Minor fix suggested in the tests, then this should be good to go.

rdtools/test/analysis_chains_test.py
martin-springer: @mdeceglie  - Thanks for pointing those out. I've changed them to lowercase.

rdtools/test/analysis_chains_test.py
martin-springer: Done.

mdeceglie (approved):

Issue #271, opened 2021-04-12
TrendAnalysis uses `power_expected` for both sensor- and clearsky-based workflows


kandersolar:
Not necessarily a bug, but a perhaps unintuitive behavior:  when passing in an external `power_expected` time series, it is used for both sensor-based and clearsky-based workflows.  To my mind, part of the spirit behind the clear-sky workflow is that the normalization is tied to the modeled clear-sky irradiance, which is not the case when using a custom `power_expected`.  I think if passing in an external `power_expected` time series, the sensor and clear-sky workflows become identical except for the `csi_filter` (not 100% sure about this though).

Some ideas:

- Separate `power_expected_sensor` and `power_expected_clearsky` inputs
- Keep the code the way it is now, but emit a warning if `power_expected` is used in the clear-sky workflow
- Do nothing, because the code is just doing what you told it to do

martin-springer:
Let's implement the warning.

martin-springer:
addressed with PR #439

PR #439, opened 2024-12-02
Fix issue #271

martin-springer:
- [x] Updated changelog

### DIFF
--- a/rdtools/analysis_chains.py
+++ b/rdtools/analysis_chains.py
@@ -957,9 +957,9 @@ class TrendAnalysis:
                 pv_input="energy",
             )
             warnings.warn(
-                """Clear-sky analysis is peformed but `power_expected` was passed in by user.
+                """Clear-sky analysis is performed but `power_expected` was passed in by user.
                    In this case, the power normalization is not tied to the modeled clear-sky
-                   irradiance and the clear-sky workflow will provide the same results as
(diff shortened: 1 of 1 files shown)


rdtools/analysis_chains.py
```diff
@@ -956,6 +956,12 @@ def _clearsky_preprocess(self):
                 self.poa_global_clearsky,
                 pv_input="energy",
             )
+            warnings.warn(
+                """Clear-sky analysis is peformed but `power_expected` was passed in by user.
+                   In this case, the power normalization is not tied to the modeled clear-sky
+                   irradiance and the clear-sky workflow will provide the same results as
+                   the sensor workflow."""
+            )
```
mdeceglie, line 964: There should be some subtle differences in the filtering, so I suggest we modify the warning a bit to indicate results will be similar.
```suggestion
                """Clear-sky analysis is performed but `power_expected` was passed in by user.
                   In this case, the power normalization is not tied to the modeled clear-sky
                   irradiance and the clear-sky workflow may provide similar results to
                   the sensor workflow."""
            )
```

mdeceglie (changes requested):

rdtools/analysis_chains.py
martin-springer: @mdeceglie - yep, that sounds good to me.

mdeceglie (approved):

Issue #313, opened 2022-02-03
Difference in TrendAnalysis filtering behavior between power=0 and power=nan 


kandersolar:
Changing `TrendAnalysis(pv=power, ...)` to `TrendAnalysis(pv=power.replace(0, np.nan), ...)` can change the degradation output.  There are at least two differences in behavior between `power=0` and `power=np.nan`, both of which occur in the pre-aggregation filtering steps:

1) In `rdtools.clipping.quantile_clip_filter` the clipping threshold is calculated as the 98th percentile of `power_ac`.  This ignores `power_ac=np.nan` but not `power_ac=0`.  With `power_ac=0` the threshold tends to be lower and the filter excludes more points.
2) In `TrendAnalysis._pvwatts_norm`, `energy_normalized` is optionally renormalized by dividing by its 95th percentile.  Times with `pv_energy=0` and nonzero irradiance affect this 95th percentile, but times where `pv_energy=np.nan` do not.  By changing this renormalization constant, the entire `normalized_energy` series is scaled up or down and points get moved into or out of the allowed range specified in `rdtools.filtering.normalized_filter`.

Note that the clipping effect applies to any `power=0` while the normalization effect only applies to `power=0` during daylight hours (i.e., when expected production is nonzero). 

Here's a modified version of the PVDAQ4 TrendAnalysis notebook showing how the clipping threshold varies (856 to 819 in this case): https://gist.github.com/kanderso-nrel/ea025bf4ea94397b16865e3fcacd9d88

I don't yet have a good example for the second effect.

Marking this as a bug since it seems to me that both of these are undesirable effects.  Thanks @williamhobbs for noticing the difference in output!

mdeceglie:
We should be careful about changing answers for users relying on the default behavior. Let's add a kwarg, whose default value triggers a deprecation warning and a suggestion. The preferred option for that kwarg would use some minimum as a threshold for inclusion in the quantile calculations. We'll use the preferred option in updated examples.

mdeceglie:
New plan: address this bug in the examples for now, then implement a more robust but result-changing fix in RdTools 3

PR #442, opened 2024-12-11
Fix issue #313 (different results with Nan's and Zeros in power series)

martin-springer:
- [x] Updated changelog


rdtools/utilities.py
```diff
@@ -0,0 +1,79 @@
+"""Utility functions for rdtools."""
+
+
+def robust_quantile(x, q):
+    """
+    Compute the q-th quantile of a time series (x), ignoring small values and NaN's.
+    NaN's and small values [x < Q(x,q)/1000] are removed before calculating the quantile.
+    This function ensures that time series with NaN's and distributions without
+    NaN's return the same results.
```
mdeceglie, line 9:
```suggestion
    NaN's return the same results. Should only be used if x is expected to be ≥0.
```

rdtools/utilities.py
```diff
@@ -0,0 +1,79 @@
...
+
+    return q
+
+
+def robust_median(x, q=0.99):
+    """
+    Compute the median of a time series (x), ignoring small values and NaN's.
+    NaN's and small values [Q(x,q)/1000] are removed before calculating the mean.
+    This function ensures that time series with NaN's and distributions without
+    NaN's return the same results.
```
mdeceglie, line 35:
```suggestion
    NaN's return the same results.  Should only be used if x is expected to be ≥0.
```

rdtools/utilities.py
```diff
@@ -0,0 +1,79 @@
...
+
+    return mdn
+
+
+def robust_mean(x, q=0.99):
+    """
+    Compute the mean of a time series (x), ignoring small values and NaN's.
+    NaN's and small values [x < Q(x,q)/1000] are removed before calculating the mean.
+    This function ensures that time series with NaN's and distributions without
+    NaN's return the same results.
```
mdeceglie, line 61:
```suggestion
    NaN's return the same results.  Should only be used if x is expected to be ≥0.
```

mdeceglie (commented):
Looks good. I suggest we make a minor change to the docstrings to remind us and users about the x≥0 assumption

martin-springer:
> Looks good. I suggest we make a minor change to the docstrings to remind us and users about the x≥0 assumption

@mdeceglie - Yes, that sounds good to me. I committed your additions.

mdeceglie (approved):

PR #464, opened 2025-08-15
394 multi yoy - allows multiple slope calculations across multi-year periods

### LINKED ISSUE #394: Multi-YoY analysis based on Hugo Quest et al 2023
At EUPVSEC this year, Dirk came across the following great publication about improvements to the YoY method by not just taking the Y1-Y2 slopes, but also taking all of the Y1-Y3 and Y1-Y4 and Y2-Y4 slopes etc.

[Multi-yoy rd analysis_Quest_EPFL_EUPVSEC_2023_poster.pdf](https://github.com/NREL/rdtools/files/12809044/Multi-yoy.rd.analysis_Quest_EPFL_EUPVSEC_2023_poster.pdf)

This could be accomplished relatively easily in degradation.py

cdeline:
This PR builds on #456 and #460 to allow 'multi-YoY' degradation calculations similar to Hugo Quest et al 2023.  Fixes #394.
It is enabled by calling `degradation.degradation_year_on_year('multi_yoy'=True)`, or in TrendAnalysis with `ta.sensor_analysis(analyses=['yoy_degradation'], yoy_kwargs={'multi_yoy':True, 'label':'center'}) `.  At the moment there are no example notebooks that demonstrate the process, but these could be developed.

- [x] Updated changelog
- [x] Add pytests for multi_yoy = True
- [ ] Add pytests for plotting(detailed=True)


cdeline:
Comments from 8/19 PV Fleet review:
- [ ] The two dataframes returned by calc_info: `YoY_values` and `YoY_times`  have non-unique index values, therefore to compare them you have to implicitly rely on an integer index rather than the actual non-unique timeseries index. 
- [x]  Detailed=True plotting: instead of even vs odd coloring, look at the number of times the number appears as a right label minus the number of times it appears as a left label.  Is this number zero or non-zero?

Chris: this first point is non-ideal, but these return values are largely for internal use within Rdtools, so maybe it would be ok as long as it's internally consistent and all of the RdTools functions still work.

cdeline:
One consequence of the non-monotonic `YoY_values` index is that the `degradation_timeseries_plot` will not work with multi-YoY values.   A possible solution here is that if multi-YoY values are detected, we use `pd.resample.mean()` to flatten the multiple slope values and get back to a monotonic index, just for the purpose of the timeseries_plot.
- [x] based on PV Fleet feedback, just resample multiple daily slope values for the purposes of the timeseries plotting and display a warning message.

cdeline:
Confirmed that TrendAnalysis is compatible with the multi-YoY changes by passing kwargs:
`ta.sensor_analysis(analyses=['yoy_degradation'], yoy_kwargs={'multi_yoy':True, 'label':'center'}) `

martin-springer:
Closing - replaced with #470

PR #465, opened 2025-08-21
3.0.1 patch

mdeceglie:
Dependency updates

- [ ] ~Code changes have been evaluated for compatibility/integration with TrendAnalysis~
- [x] Updated changelog


setup.py
```diff
@@ -47,8 +47,7 @@
     "numpy >= 1.22.4",
     "pandas >= 1.4.4",
     "statsmodels >= 0.13.5",
-    # statsmodels 0.14.4 is not able to handle the latest scipy
-    "scipy >= 1.8.1, <1.16.0",
+    "scipy >= 1.8.1",
```
martin-springer, line 50: Good catch to remove the SciPy limitation while bumping the other dependencies!

martin-springer (approved):

setup.py
mdeceglie: Thanks to @cdeline for catching that fix!

Issue #459, opened 2025-06-22
degradation_year_on_year should allow center-labeled YoY_values


cdeline:
**Is your feature request related to a problem? Please describe.**
At the moment, the `YoY_values` returned in degradation_year_on_year.calc_info is only right-labeled.  This is a problem because to enable #394 (multi-year-on-year calculations) all of the values should probably be center-labeled to be able to line up together correctly. And #455 (time dependent degradation labeling) would benefit from starting with center-labeled `YoY_values`.

**Describe the solution you'd like**
Have a kwarg in degradation_year_on_year with label='center'.  Current default behavior is label='right'.  Consider changing default behavior to label='center'.

Issue #455, opened 2025-05-28
degradation_timeseries_plot should have different label= options


cdeline:
At the moment, the degradation_timeseries_plot shows the timeseries shifted pretty far to the right, relative to the raw data timeseries.  This is because the `yoy_results` output from the TA object is right-labeled (annual YoY slopes are 1 year in length and labeled on the right side), and the pd.rolling function to get the rolling median for the timeseries is also right labeled.  So if you have the default 365 day rolling median, your timeseries starts 2 years delayed from the raw data.

Being able to select options for 'right', 'center' and 'left' labeling would allow an output closer to what I would expect, which would have the rates line up more closely with the raw data timeseries.

PR #487, opened 2026-03-05
Release 3.2.0

martin-springer:
### Summary

This PR collects the changes for the v3.2.0 release, including new multi-year-on-year degradation analysis features, plotting improvements, bug fixes, and expanded test coverage.

### Enhancements

- **`degradation_year_on_year` — `label` parameter** (#459): New `label=` parameter (`'right'`, `'center'`, or `'left'`) controls which edge of the YoY slope is used for labeling `calc_info['YoY_values']`. Default is `'right'` (backward-compatible).
- **`degradation_year_on_year` — `multi_yoy` parameter** (#394): New `multi_yoy=` parameter (default `False`) enables multi-year-on-year degradation calculations similar to Hugo Quest et al 2023, computing slopes over N × 365 day separations.
- **`degradation_year_on_year` — `YoY_times` output** (#459): `calc_info['YoY_times']` now returns a DataFrame with `dt_right`, `dt_center`, and `dt_left` columns for each slope.
- **`degradation_timeseries_plot` improvements** (#455): Rolling median is now centered (`center=True`) and `min_periods` reduced from `rolling_days//2` to `rolling_days//4`.
- **`degradation_timeseries_plot` multi-YoY support** (#394): Handles `multi_yoy=True` data by resampling overlapping YoY values to their daily mean (with a warning).
- **`degradation_summary_plots` parity coloring** (#394): `detailed=True` mode now properly colors points by odd/even usage count, not just 0/1/2.
- **`sensor_analysis` / `clearsky_analysis`**: Now explicitly default `yoy_kwargs={"label": "right"}`.
- **New example notebook**: `docs/Multi-year_on_year_example.ipynb` demonstrates `label='center'` and `multi_yoy=True` (#394).

### Bug Fixes

- Fixed `usage_of_points` calculation in `degradation_year_on_year` to properly handle `multi_yoy=True` mode with overlapping slopes (#394).

### Maintenance

- Added `_avg_timestamp_old_Pandas` helper for pandas <2.0 compatibility when calculating center labels.
- Fixed nbval workflow command syntax (`--sanitize-with` → `--nbval-sanitize-with`).
- Improved pandas 3.0 compatibility with datetime resolution handling.
- Updated `docs/notebook_requirements.txt` to require `numexpr>=2.10.2` and `tabulate>=0.9.0`.
- Removed trailing semicolons in example notebooks.
- Added `.coverage.*` pattern to `.gitignore`.

### Testing

- Added error-handling tests for `analysis_chains`: `filter_params` validation, `clearsky_rescale_index_mismatch`, `poa_filter_without_poa`, `tcell_filter_without_temperature`, `hour_angle_filter_without_location`, `clearsky_filter_without_poa`, `degradation_timeseries_plot_invalid_case`.
- Added error-handling tests for `degradation`: `classical_decomposition` missing/irregular data, `year_on_year` circular block validation, no valid pairs, `_mk_test` edge cases.
- Added tests for `multi_yoy=True` in `degradation_year_on_year`.
- Added tests for `degradation_timeseries_plot` covering `label='center'`, `label='left'`, multi-YoY duplicate index, and `KeyError` path.
- Set matplotlib backend to `Agg` in `conftest.py` to avoid tkinter issues.

### Checklist

- [ ] Updated changelog


cdeline:
@martin-springer : I'm getting setup.py deprecation messages when I install.  Would this be the time to switch to a pyproject.toml?

rdtools/analysis_chains.py
```diff
@@ -1001,7 +1001,7 @@ def _clearsky_preprocess(self):
         )
 
     def sensor_analysis(
-        self, analyses=["yoy_degradation"], yoy_kwargs={}, srr_kwargs={}
+        self, analyses=["yoy_degradation"], yoy_kwargs={"label": "right"}, srr_kwargs={}
     ):
```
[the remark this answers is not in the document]
cdeline: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/NatLabRockies/rdtools/pull/487#discussion_r2891078915)

rdtools/plotting.py
```diff
@@ -484,7 +487,22 @@ def _bootstrap(x, percentile, reps):
...
+            "Input `yoy_info['YoY_values']` appears to have multiple annual "
+            "slopes per day, which is the case if "
+            "degradation_year_on_year(multi_yoy=True). "
+            "Proceeding to plot with a daily mean which will average out the "
+            "time-series trend. Recommend re-running with "
+            "degradation_year_on_year(multi_yoy=False)."
+        )
+        roller = results_values.resample('D').mean().rolling(f'{rolling_days}d',
+                                                             min_periods=rolling_days//4,
+                                                             center=True)
```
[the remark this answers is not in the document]
cdeline: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/NatLabRockies/rdtools/pull/487#discussion_r2891078937)

martin-springer:
> @martin-springer : I'm getting setup.py deprecation messages when I install. Would this be the time to switch to a pyproject.toml?

@cdeline - yes, we should do that. I'll open a PR and make the changes.

rdtools/analysis_chains.py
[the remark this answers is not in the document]
cdeline: As I described in the PR, I dont like these proposed changes. The 'issue' that is solved is not really an issue because our mutable types that gets passed in like the 'yoy_kwargs' dict is only ever read, never written or updated. And the downside to CoPilot's suggestion is that the proposed change moves important default behavior deeper into the function and makes it harder to trace the logic and document how to interact with the TA object for the user.  @martin-springer do you agree?

rdtools/plotting.py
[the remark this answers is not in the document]
cdeline: Merged this PR as #490.

rdtools/analysis_chains.py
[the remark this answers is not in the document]
martin-springer: @cdeline - yes, we can disregard this suggestion

mdeceglie:
With `calc_info['YoY_values'['dt_right', 'dt_center', 'dt_left']` why do we also need to introduce the `label` parameter, this seems redundant?

rdtools/plotting.py
```diff
@@ -484,7 +487,23 @@ def _bootstrap(x, percentile, reps):
...
```
mdeceglie, line 503: I'm unclear on why this is needed. At the end of the day we're plotting a median of all the relevant slopes within `rolling_days` why not just throw the multi-yoy slopes in with the rest of the sample instead of taking an intermediate mean?

cdeline:
> With `calc_info['YoY_values'['dt_right', 'dt_center', 'dt_left']` why do we also need to introduce the `label` parameter, this seems redundant?

The main thing that `label=` does is to change the timestamp in the `calc_info['YoY_values']` timeindex to either match right, center or left.  That is the timeindex that gets passed into the plotting functions, so I suppose it mainly affects the degradation_timeseries_plot.

mdeceglie:
Which label is the best one to use, especially when multi-yoy is in play? Center?

rdtools/plotting.py
cdeline: This may be a limitation of the pd.rolling function which I believe requires a unique timeindex to operate on.  I can investigate options here.

cdeline:
> Which label is the best one to use, especially when multi-yoy is in play? Center?

Yes I would definitely recommend center-labeling of the 'YoY_values' timeseries, particularly with multi-yoy analysis.  If we get rid of the `label=` kwarg option in `degradation_year_on_year()` I would recommend changing from the default of right-labeled to center-labeled.

rdtools/plotting.py
cdeline: I'm opening a side branch 'multi_YoY_no_label' to refactor around removing the `label=` kwarg in  `degradation_year_on_year`  and re-write `degradation_timeseries_plot()` to plot the median of all slopes that have either their starting or ending point within the `rolling_days` window.

mdeceglie:
Well we can't change the default behavior of `calc_info['YoY_values']` until version 4. But I think we can achieve all the functionality here but just adding the `calc_info['YoY_times']`. Since `degradation_timeseries_plot` gets `calc_info` as it's argument, it can make the right decision about how to plot.

I suggest that `degradation_timeseries_plot` get a key word argument that tells it which of the columns in `calc_info['YoY_times']` to use. Since the current default behavior is confusing/not ideal, let's consider whether we do a work around wherein if the new kwarg is not passed, the old behavior prevails but with a warning that the new best practice is e.g. `slope_centering=True`, and use the new kwarg with the preferred value in the examples.

My concern here is twofold:

1. The current changes to `degradation_year_on_year` are redundant
2. We are trying to control the behavior of `degradation_timeseries_plot` by proxy through `degradation_year_on_year` instead of in the plotting function itself

docs/sphinx/source/changelog/v3.2.0.rst
```diff
@@ -0,0 +1,78 @@
+*************************
+v3.2.0 (X, X, 2026)
+*************************
+
+Enhancements
+------------
+* :py:func:`~rdtools.degradation.degradation_year_on_year` has new parameter ``multi_yoy``
+  (default False) to trigger multiple YoY degradation calculations similar to Hugo Quest et
```
mdeceglie, line 8: Let's include a citation to the Quest paper

docs/sphinx/source/changelog/v3.2.0.rst
```diff
@@ -0,0 +1,78 @@
+*************************
+v3.2.0 (X, X, 2026)
```
mdeceglie: We need to add an explanation that the index of YoY_values has changed from a timestamp to integer and that the original behavior can be reconstructed with the help of `calc_info['YoY_times']`. Strictly speaking, I don't think this is backward compatible, but for this case I think it may be ok to keep it in a minor version update.

docs/sphinx/source/changelog/v3.2.0.rst
```diff
@@ -0,0 +1,78 @@
...
+*************************
```
mdeceglie, line 3: Another even more in the weeds changes is that we have replaced NaNs in `calc_info['usage_of_points']` with zeros. I don't know if this ever happens in practice but probably worth it to call out the edge case change.

rdtools/degradation.py
```diff
@@ -218,7 +224,7 @@ def degradation_year_on_year(energy_normalized, recenter=True,
         degradation rate estimate
     calc_info : dict
```
mdeceglie, line 225: should add explanation of `calc_info['YoY_times']`

mdeceglie (commented):
This is looking pretty close! In addition to my contextual comments, a few comments on the example notebooks:

1. Change the links to the example data in the notebooks' markdown sections
2. Let's add some more explanation to the TrendAnalysis example notebook `plot_degradation_timeseries` part. Something like:  "In this case the fluctuations in apparent degradation rate are probably due to the soiling component of the signal. This example illustrates the pitfalls of attempting to look at how degradation rate evolves in time in the presence of a large uncorrected soiling signal. Nevertheless it illustrates the plotting functionality that can be useful in applications not affected by soiling or other reversible changes."

martin-springer:
> This is looking pretty close! In addition to my contextual comments, a few comments on the example notebooks:
> 
> 1. Change the links to the example data in the notebooks' markdown sections
> 2. Let's add some more explanation to the TrendAnalysis example notebook `plot_degradation_timeseries` part. Something like:  "In this case the fluctuations in apparent degradation rate are probably due to the soiling component of the signal. This example illustrates the pitfalls of attempting to look at how degradation rate evolves in time in the presence of a large uncorrected soiling signal. Nevertheless it illustrates the plotting functionality that can be useful in applications not affected by soiling or other reversible changes."

@mdeceglie - done

mdeceglie (approved):
Looks great, thanks everyone!

PR #518, opened 2026-07-15
Add signal-decomposition degradation analysis

bmeyers:
Ports the CVXPY seasonal-trend decomposition prototype into a new `rdtools.signal_decomposition` module and wires it end-to-end through `TrendAnalysis`. The implementation now provides a unified framework for conventional degradation analysis, expanded lifecycle trend analysis, and SD++ soiling analysis.

New module (`rdtools/signal_decomposition.py`):

- Public `degradation()` entry point returning `(Rd_pct, Rd_CI, sd_trend_results)`
- Three trend types: linear, year-1 piecewise-linear breakpoint, and nonlinear monotone
- Four loss functions: L2, L1, Huber, and conditional quantile
- Optional natural-log transform; log-Huber is the recommended and default configuration
- Missing data handled natively through a masked equality constraint
- Moving-block bootstrap confidence intervals for overall and lifecycle-dependent degradation rates
- Stability and conditional-quantile sweeps for diagnosing model sensitivity and degradation patterns
- Plotting/reporting: `plot_decomposition`, `plot_trend`, `plot_stability`, `animate_degradation`, and `format_degradation_report`

SD++ soiling:

- Adds a nonpositive soiling component for dry, accumulation-and-recovery behavior
- Solves a deterministic regularization path and applies an explicit structural detection rule
- Returns a true no-soiling model when no coherent component is detected
- Freezes the detection decision before applying two interval-weighted IRL1 refinement steps to reduce shrinkage
- Reports time-averaged and quarterly loss, insolation-weighted loss when available, local soiling-rate intervals, cleaning-event strength, bootstrap uncertainty, and selector diagnostics
- Adds soiling decomposition plots and formatted reports

TrendAnalysis wiring (`rdtools/analysis_chains.py`):

- `_signal_decomposition_degradation` private integration method
- `sensor_analysis` and `clearsky_analysis` accept `signal_decomposition` in the `analyses` list and pass configuration through `sd_kwargs`
- `plot_signal_decomposition_summary` method
- Existing preprocessing can be reused explicitly across repeated analyses

Dependencies and minimum environment:

- Python 3.11 is now the minimum supported version; Python 3.10 was removed from CI and Pixi environments
- CVXPY >= 1.5.3 and SPCQE >= 0.0.3 are hard runtime dependencies
- SciPy minimum raised from 1.8.1 to 1.13.0
- pvlib minimum lowered to the tested 0.11.0 floor
- CVXPY and SPCQE are included in the exact minimum-dependency environment
- `check-deps` verifies that all direct dependency pins match their declared lower bounds
- `test-min` builds the Python 3.11 lower-bound environment and runs the complete test suite
- Pixi is pinned in CI for reproducible lockfile behavior

Tests and documentation:

- Expanded `signal_decomposition_test.py` coverage for trend models, losses, bootstrap uncertainty, structural soiling selection, IRL1 refinement, null behavior, metrics, plots, and reports
- Integration coverage in `analysis_chains_test.py`
- Complete minimum-environment suite passes with 491 tests
- Added a full signal-decomposition guide covering basic degradation, expanded trend analysis, diagnostic sweeps, and SD++ soiling
- Added public PVDAQ examples with synthetic year-1 breakpoint and dry-soiling signals
- Added generated decomposition and soiling figures
- Updated `api.rst` and the v3.3.0 changelog

Todo list:

- [x] Code changes have been evaluated for compatibility/integration with `TrendAnalysis`
- ~~[ ] New functions added to `__init__.py`~~
  This is intentionally not being done. The function should be imported as `rdtools.signal_decomposition.degradation` or `from rdtools.signal_decomposition import degradation` to avoid clashing with the existing `rdtools.degradation` module.
- [x] `api.rst` and the other Sphinx documentation pages are up to date
- [x] Example notebooks have been rerun and result differences scrutinized
- [x] Changelog updated


bmeyers:
Just finishing up examples and package dependency testing. Ready for general review. Open item: minimum scipy version support.

bmeyers:
We updated the minimum supported environment to Python 3.11 and defined `dev-min` as a coherent, tested lower-bound stack. Each direct runtime dependency is pinned to its declared minimum, `check-deps` verifies those pins remain synchronized, and `test-min` builds the environment and runs the full test suite. Pixi is also pinned in CI for reproducibility.

The newest package in this minimum environment is CVXPY 1.7.5, released December 5, 2025. As of August 25, 2026, the environment is therefore about 8 months and 20 days old. Most other minimum dependencies date from 2022–2024, with a few newer requirements introduced by the signal-decomposition stack. This gives us a realistic, reproducible minimum environment rather than a collection of independently old version floors that may not build or work together.

bmeyers:
I am currently working on raising the "age" of the minimum environment, which is currently about 8 months. The new target is 2 years. Cvxpy has been downgraded successfully, and I'm now testing additional binding packages.

bmeyers:
**Follow-up:** The minimum-version review is complete. We lowered the tested floors to CVXPY 1.5.3, pvlib 0.11.0, and SPCQE 0.0.3, and regenerated the Pixi lockfile. The complete `dev-min` suite passes with 491 tests.

The newest signal-decomposition dependency floor is now SPCQE 0.0.3, released October 28, 2024—1 year, 9 months, and 28 days old as of August 25, 2026. This is close to the two-year target while retaining the first stable SPCQE release; the older pre-cutoff release is an alpha with an incomplete public API.

The complete environment still contains one newer direct requirement: `bayesian-filters 1.4.5`, released October 20, 2025. This is the maintained drop-in replacement for the unmaintained FilterPy package and has no earlier PyPI release, so reverting it solely to improve the environment-age statistic would undo an existing maintenance decision. Subject to that exception, the newly introduced signal-decomposition stack now has a practical minimum age of approximately two years.

bmeyers:
Last remaining open item is to update the changelog, per the PR template

bmeyers:
I have no more planned commits for this PR. Ready for final review!

cdeline:
This is more of an issue for getting this into Main, not neceesary for getting it into Dev branch.  I like the Marimo notebook `signal_decomposition_example.py` but worry about how people will understand how to interact with it without instructions.  I'd recommend: A) re-writing the `signal_decomposition_soiling_example.py' as a traditional ipynb notebook.  B) Include instructions at the end of that notebook for running the signal_decomposition_example marimo in edit mode so they can see the guts. C) rename the marimo notebooks to include _marimo  somewhere in the filename so people understand that its' not just a standard python to be run.

docs/generate_signal_decomposition_figures.py
```diff
@@ -0,0 +1,287 @@
...
+def central_limits(values, lower=1, upper=99, margin_fraction=0.1):
+    """Return robust plotting limits without changing the fitted data."""
+    low, high = np.nanpercentile(values, [lower, upper])
+    margin = max((high - low) * margin_fraction, 0.005)
+    return low - margin, high + margin
+
+
+def load_pvdaq_system_4():
+    """Load the public PVDAQ record used by the documentation notebooks."""
+    data = pd.read_csv(
```
cdeline, line 38: on my computer this is saved as a pickle file, so I needed to swap this for pd.read_pickle

docs/generate_signal_decomposition_figures.py
```diff
@@ -0,0 +1,287 @@
+"""Generate static figures for the signal-decomposition documentation.
+
+Run from the repository root with::
+
+    pixi run -e dev python docs/generate_signal_decomposition_figures.py
```
cdeline, line 5: pixi wasn't working for me, so I ran it in my rdtools environment

docs/generate_signal_decomposition_figures.py
```diff
@@ -0,0 +1,287 @@
...
+
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+
+import rdtools
+
+
+DOCS_DIR = Path(__file__).resolve().parent
+IMAGE_DIR = DOCS_DIR / "sphinx" / "source" / "_images"
```
cdeline: my python interpreter barfed at these.  replace with os.path.join() ?

rdtools/analysis_chains.py
```diff
@@ -201,6 +202,9 @@ def __init__(
         if power_expected is not None and temperature_cell is None:
             del self.filter_params["tcell_filter"]
 
+        self._sensor_preprocessed = False
```
cdeline, line 205: Nice. it's probably good that we're explicitly adding a flag now that these preprocess steps have been done, rather than implicitly looking for effects of preprocess runs.

rdtools/analysis_chains.py
```diff
@@ -1010,18 +1090,33 @@ def sensor_analysis(
...
+            (e.g. a reactive notebook). Note: if ``filter_params`` have been
+            modified since the last preprocess, set this to False to pick up
+            the changes.
 
         Returns
         -------
         None
         """
-        self._sensor_preprocess()
+        if not (skip_preprocess and self._sensor_preprocessed):
```
cdeline, line 1118: This is changing the original workflow slightly, in that **if** self.sensor_analysis is called multiple times, the _sensor_preprocess() is only called once, rather than once for each call.  We should just confirm that there are no conditions in which re-running self.sensor_analysis would result in weird behavior.

rdtools/signal_decomposition.py
```diff
@@ -0,0 +1,2121 @@
+"""
+Signal decomposition degradation analysis for PV systems.
+
+Seasonal-trend decomposition via convex optimization:
+
```
cdeline, line 5: I would like to see some additional documentation and citations in here including the Boyd / Meyers papers.  This docstring is what shows up in ReadtheDocs when people try to look up what is going on in this modules, and at the moment it is kind of barebones.  There should be some more breadcrumbs to follow for more information.

rdtools/signal_decomposition.py
```diff
@@ -0,0 +1,2121 @@
...
+        Unique, sorted window lengths each ending on a non-NaN sample.
+    """
+    y_full = np.asarray(y_full, dtype=float)
+    N = len(y_full)
+    valid_ix = np.where(~np.isnan(y_full))[0]
+
+    if len(valid_ix) == 0:
+        raise ValueError("y_full contains no valid (non-NaN) samples.")
+
+    nominal = np.arange(int(T), N + 1, step)
```
cdeline, line 1052: Claude suggestion:
get_valid_endpoints crashes on short series
In signal_decomposition.py, get_valid_endpoints does:

nominal = np.arange(int(T), N + 1, step)
then unconditionally uses nominal[-1]
This fails when the series is shorter than one year, because np.arange(int(T), N + 1, step) is empty when N < T. In that case the function raises an IndexError instead of handling the input gracefully.

That is a real correctness issue because the new decomposition APIs generally accept arbitrary-length series and the code tries to support shorter histories elsewhere, but this helper does not guard for them. At minimum it should:

return a clear validation error, or
handle short records by returning the valid length(s) without crashing

cdeline (approved):
I think these changes are fine to include in Development.  I'm including a few discussion points and one or two requests for additional content, but otherwise it's looking good.

